Issue #1, opened 2014-12-25
[RFC] Pluggable API documentation generator


sloria:
Now that smore has many of the lower-level functions for converting marshmallow `Schema` and webargs `Args` to swagger definitions, next step is to implement a system for generating full API docs.

Ideas for initial iteration:
- Based on Swagger 2.0 spec. This will allow us to leverage the latest Swagger-UI
- Pluggable. The documentation generator will work with any web framework, with or without webargs, etc. etc. Plugins provide helpers for generating metadata.
- Easy way to serve swagger docs. Possibly part of the Flask plugin.

Ideas for the future:
- Generate swagger-based from docstrings.
- Sphinx extension?
## Proof of concept

I wrote up a simple proof-of-concept in this gist: https://gist.github.com/sloria/dc1b2d2e43fbcea866ae
## Prior art
- [django-rest-swagger](https://github.com/marcgibbons/django-rest-swagger)
- [flask-restful-swagger](https://github.com/rantav/flask-restful-swagger)
- [flask-restplus](https://github.com/noirbizarre/flask-restplus)
- [cornice](http://cornice.readthedocs.org/en/latest/sphinx.html) (not swagger-based, but may provide ideas for sphinx extension)

hello-josh:
I like the direction this is going in, integrating webargs / marshmallow to handle the minutiae around input and output. One thing that irks me with the current flask-based approaches is that you spend a bunch of time documenting your api but then you still have to manually extract your args from the request (e.g. `args = parser.parse_args()`) and most of them rely on magic dictionaries that match up 1-to-1 with the swagger spec. IMO you end up writing so much code that you wonder what the library is actually buying you in the first place?

Sometimes I think it would be less word to craft the swagger json schema by hand and then pass it into your framework's router and have it enforce, marshal, and pass parsed input directly into your view handler instead of adding decorators on top of decorators.

I guess what I really am getting at is that the current implementations I have seen are not written at a high enough level of abstraction to make them any more useful than implementing it by hand.

sloria:
Thanks for your input @Trii. I agree with your point re: solutions that require you to write your entire Swagger spec manually (in code).

The idea I'm proposing here is that plugins do the work of "writing" your Swagger spec. So if you use no plugins, you write your whole spec. Use Flask? The Flask plugin generates Swagger Paths so you don't have to. Use marshmallow? The marshmallow plugin generates Swagger Schema Objects for you. And so on. You mix and match plugins according to your stack.

[...]

PR #2, opened 2015-03-26
Make the tests pass by cleaning up some stuff.

hello-josh:
I pulled this today to see how I can use it for my current project based off of the notes from #1 as well as the [webapp2 parser](https://github.com/sloria/webargs/pull/36).

Here is my changelog in getting tests to pass:
- Fix up broken unit tests due to `webargs.Arg.target` being renamed to `location`
- Add some PEP8 newlines before classes, minor indentation
- Make the `responses` field required per spec on the [Operation Object](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#operation-object)
- Fix bug in `smore.apispec.core` that was causing some test failures with
- KeyError due to path operations not always matching up with the registered request helpers
- Made a pass at the TODO: nesting in the above fix
- Drank a lot of coffee

Take a look and I'll clean up anything you'd like me to clean. I'm only 75% happy with the nesting fix. It is still a bit too confusing to read.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -7,8 +7,6 @@ webargs>=0.9.0
 WTForms>=2.0.1
 colander>=1.0
 Flask
-marshmallow
-pyyaml
 
 # testing
 pytest
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -7,6 +7,8 @@ webargs>=0.9.0
 WTForms>=2.0.1
 colander>=1.0
 Flask
+marshmallow
+pyyaml
 
 # testing
 pytest
--- a/smore/apispec/core.py
+++ b/smore/apispec/core.py
@@ -29,11 +29,15 @@ class Path(object):
         invalid = set(iterkeys(operations)) - set(VALID_METHODS)
         # Reject invalid http methods
         if invalid:
-            raise APISpecError('One or more HTTP methods are invalid: {0}'.format(", ".join(invalid)))
+            raise APISpecError(
+                'One or more HTTP methods are invalid: {0}'.format(", ".join(invalid))
+            )
         # Reject invalid operations definitions
         for method, operation in iteritems(operations):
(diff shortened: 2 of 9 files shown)


sloria:
Wow, thanks for the hard work on this!

Will review this more in depth over the weekend, but this looks good after a quick initial pass.

hello-josh:
After this, I am working on webapp2 integration as well via https://github.com/Trii/smore/pull/1

dev-requirements.txt
```diff
@@ -7,6 +7,8 @@ webargs>=0.9.0
 WTForms>=2.0.1
 colander>=1.0
 Flask
+marshmallow
```
sloria: These are already dependencies of `smore`, so they will be installed after doing 

``` python
python setup.py develop
```

so I don't think it's necessary to add these here.

smore/apispec/core.py
```diff
@@ -78,18 +91,17 @@ def add_path(self, path=None, operations=None, **kwargs):
...
-                            for func in funcs:
-                                path.operations[method]['responses'][status_code].update(
-                                    func(self, **kwargs)
-                                )
+        # Process response helpers for any path operations defined.
+        # Rule is that method + http status exist in both operations and helpers
+        methods = set(iterkeys(path.operations)) & set(iterkeys(self._response_helpers))
+        for method in methods:
+            responses = path.operations[method]['responses']
+            statii = set(iterkeys(responses)) & set(iterkeys(self._response_helpers[method]))
```
sloria: Clever, but why not `statuses`? =)

sloria:
Nice! Once the comments above addressed, this is good to merge.

smore/apispec/core.py
hello-josh: lol, its my version of the old `42` answer

hello-josh:
All set @sloria

sloria:
Thanks @Trii !

PR #3, opened 2015-03-30
Initial webapp2 plugin

hello-josh:
This is what I am working on right now. I removed the check for non empty `responses` from  `smore/apispec/core.py` so that it could be aggregated from `APISpec.add_path` as well as docstring content. I was going under the assumption that the final call to generate the swagger document would take care of checking for missing items that are marked required per the spec.

### DIFF
--- a/smore/apispec/core.py
+++ b/smore/apispec/core.py
@@ -32,12 +32,6 @@ class Path(object):
             raise APISpecError(
                 'One or more HTTP methods are invalid: {0}'.format(", ".join(invalid))
             )
-        # Reject invalid operations definitions
-        for method, operation in iteritems(operations):
-            if len(operation.get('responses', {})) == 0:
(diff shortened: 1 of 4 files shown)


sloria:
Thanks again @Trii . I will look this over within the next few days.

sloria:
You will need to add `webapp2` to the dev-requirements to get the tests to run on Py2.

hello-josh:
Yeah, I realized that yesterday when I was playing with it more. I'm going to do some cleanup on this.

Were you thinking of letting plugins be their own installable package instead of included with this package? If so I will pull out the webapp2 plugin into its own package and just push back the core updates. I am at pycon this weekend so I will probably work on it s'more (bad pun)

hello-josh:
I'm gonna close this and just do some work against the paths merging instead.

PR #6, opened 2015-04-14
Add items description to array properties.

jmcarp:
"items" object is required for properties of type "array". See
https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#itemsObject.


sloria:
Thanks!

PR #7, opened 2015-04-14
Note required model fields in separate list.

jmcarp:
The official docs are vague on this point, but it looks like required
model fields are meant to be included in a `required` list on the model
definition, rather than in a `required` field on each property.
Including a `required` field on model properties also raises errors from
swagger-tools and swagger-ui.


sloria:
Good catch.

PR #13, opened 2015-05-29
Allow complete specification of top-level schema.

jmcarp:
- Remove unused parameters from `APISpec` constructor.
- Accept arbitrary `**options` to add to top-level schema.
- Update tests.

@sloria: Not sure if accepting `**options` is ideal here, since it doesn't validate that passed keyword arguments are valid Swagger fields, but I also don't want a huge constructor that lists all possible options. We could define a (JSON) schema for validating options to prevent users from passing bad values. But this works for now.

### DIFF
--- a/smore/apispec/core.py
+++ b/smore/apispec/core.py
@@ -58,14 +58,19 @@ class APISpec(object):
     :param str title: API title
     :param str version: API version
     :param tuple plugins: Paths to plugin modules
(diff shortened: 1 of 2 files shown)


sloria:
Good change. Smore doesn't try to do much validation, so it's ok that the top-level properties don't get validated. Thanks

PR #14, opened 2015-05-29
Check base path

jmcarp:
Strip `basePath` from `path` if `basePath` set.

Otherwise, setting `basePath` can break URLs in swagger-ui.

@sloria: I think this is a reasonable feature to add to smore, but you could also argue that it's the user's responsibility to strip the base path from all paths if they're going to set `basePath`. I don't have a strong intuition about this--mostly wanted to log for discussion.

Depends on #13.

### DIFF
--- a/smore/apispec/core.py
+++ b/smore/apispec/core.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 
+import re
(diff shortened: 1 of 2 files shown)


sloria:
+1

PR #16, opened 2015-06-17
Pass array collection format.

jmcarp:
Swagger allows different formats for arrays that are not passed in the
body, including comma-separated, tab-separated, and repeated parameters.
Since webargs supports repeated parameters (`foo=bar&foo=baz`), we
should set `collectionFormat` to `multi` by default. This also gets
swagger-ui to format request parameters correctly for array-typed
values.


sloria:
Thanks!

PR #20, opened 2015-07-11
[WIP] Initial prototype of pagination.

jmcarp:
TBD:
- Should this live in smore or elsewhere?
- Should we use ABCs or stick to a single ORM / database?
- Should we integrate with web frameworks for hypermedia purposes? For example, DRF pagination includes links to next / previous pages, which is neat.

### DIFF
--- a/smore/pagination/models.py
+++ b/smore/pagination/models.py
@@ -1,0 +1,124 @@
+import abc
+import math
+import collections
+
+import six
+
+
+class BasePage(six.with_metaclass(abc.ABCMeta, collections.Sequence)):
+    """A page of results.
+    """
+    def __init__(self, paginator, results):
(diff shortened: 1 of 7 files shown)


smore/pagination/models.py
```diff
@@ -0,0 +1,124 @@
+import abc
+import math
+import collections
+
+import six
+
+
+class BasePage(six.with_metaclass(abc.ABCMeta, collections.Sequence)):
```
sloria, line 8: `marshmallow.compat` already provides a `with_metaclass` function, so `six` shouldn't be necessary here.

jmcarp:
Moved to https://github.com/jmcarp/marshmallow-pagination for now.

PR #26, opened 2015-09-20
Convert marshmallow Schemas to Swagger parameters.

jmcarp:
Proposal:

Given plans to add request parsing based on marshmallow `Schema` classes, we need a mechanism to convert `Schema`s to Swagger parameter objects. This patch adds the `schema2parameters` helper, which returns a list of a single parameter object if the location is the JSON body of the request (since Swagger only permits a single parameter with "in" equal to "body"), and a list of one parameter per `Field` otherwise.

We may also want to support the use case of converting a mapping of parameter names to a list of parameters (e.g. `{'name': fields.Str()}`). This would be straightforward if we decide to add it.

If this interface makes sense, I'll flesh out tests and documentation. See the integration test in `tests/swagger/test_swagger.py` for usage examples.

:guitar:

### DIFF
--- a/smore/ext/marshmallow.py
+++ b/smore/ext/marshmallow.py
@@ -40,7 +40,7 @@ def schema_path_helper(spec, view, **kwargs):
 def resolve_schema_dict(spec, schema):
     if isinstance(schema, dict):
         return schema
-    plug = spec.plugins[NAME]
+    plug = spec.plugins[NAME] if spec else {}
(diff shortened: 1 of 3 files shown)


sloria:
Approach looks good :+1:

jmcarp:
Patch updated @sloria. Added support for ingesting dicts of `Field` objects in `fields2parameters2`.

sloria:
Thanks!

PR #29, opened 2015-09-29
Support `enum` field of JSON Schema.

### LINKED ISSUE #28: Support `enum` property of JSON Schema
Now that webargs is using marshmallow and has access to the `OneOf` and `ContainsOnly` validators, smore can populate the `enum` property of JSON Schema fields. If this seems reasonable, I'm down to send a patch @sloria.

jmcarp:
With obligatory use of `functools`. Longer term, we might want to expose a more flexible interface for pulling JSON-Schema properties off Schemas and arbitrary validators (like the helper plugins in `APISpec`), but, you know, YAGNI.

[Resolves #28]

### DIFF
--- a/smore/swagger/__init__.py
+++ b/smore/swagger/__init__.py
@@ -9,6 +9,7 @@ and marshmallow :class:`Schemas <marshmallow.Schema>`.
 Swagger 2.0 spec: https://github.com/wordnik/swagger-spec/blob/master/versions/2.0.md
 """
 from __future__ import absolute_import, unicode_literals
+import operator
 import warnings
 import functools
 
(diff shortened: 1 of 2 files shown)


smore/swagger/__init__.py
```diff
@@ -45,6 +46,24 @@ def _get_json_type_for_field(field):
...
+
+    :param Field field: A marshmallow field.
+    :rtype: set
+    """
+    validators = [
+        validator.choices for validator in field.validators
+        if hasattr(validator, 'choices')
+    ]
+    return (
+        functools.reduce(lambda x, y: set(x).intersection(y), validators)
```
sloria: Can you use `operator.and_` here?

sloria:
Nice. Just one suggestion and this is good to merge.

jmcarp:
Updated.

PR #36, opened 2015-11-07
Add global parameters to spec.

jta:
Swagger supports referencing common parameters by inserting a `$ref` to
`#/parameters/<name>`. This commit adds a `add_parameter` method to
APISpec, which appends parameters to the base swagger config.

When parsing the YAML doc, string objects in the parameters field get
converted into references. This makes the docstrings much shorter, e.g:

```
    get:
        description: Retrieve object
        parameters:
         - arg1
         - name: arg2
           in: path
           ...
```

Currently no check is made to verify `arg1` is a valid reference, but
this could easily be done by validating the resulting yaml from
`spec.to_dict()`.

I'm not hung up on implementation since I've only played around with the
code for a bit, but I do need the feature in order to be able to write
larger APIs without repeating myself constantly. Feel free to suggest
improvements.

### DIFF
--- a/apispec/core.py
+++ b/apispec/core.py
@@ -19,16 +19,22 @@ SWAGGER_VERSION = '2.0'
 
 def clean_operations(operations):
     """Ensure that all parameters with "in" equal to "path" are also required
-    as required by the Swagger specification.
+    as required by the Swagger specification, as well as normalizing any
+    references to global parameters.
 
     See https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#parameterObject.
 
     :param dict operations: Dict mapping status codes to operations
     """
+    get_ref = lambda x: x if isinstance(x, dict) else {'$ref': '#/parameters/' + x}
+
     for operation in (operations or {}).values():
-        for parameter in operation.get('parameters', []):
-            if parameter['in'] == 'path':
-                parameter['required'] = True
+        if 'parameters' in operation:
+            parameters = operation.get('parameters')
+            for parameter in parameters:
+                if isinstance(parameter, dict) and parameter['in'] == 'path':
+                    parameter['required'] = True
+            operation['parameters'] = [get_ref(p) for p in parameters]
 
 
 class Path(dict):
@@ -90,6 +96,7 @@ class APISpec(object):
         self.options = options
         # Metadata
         self._definitions = {}
+        self._parameters = {}
         self._paths = {}
         # Plugin and helpers
         self.plugins = {}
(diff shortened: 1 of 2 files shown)


hairychris:
The Travis error is just a tiny flake8 thing:

./apispec/core.py:29:79: E226 missing whitespace around arithmetic operator

I really like the idea of this change although I've not had time to dig into how it works yet.

apispec/core.py
```diff
@@ -106,11 +113,22 @@ def to_dict(self):
             'swagger': SWAGGER_VERSION,
             'info': self.info,
             'definitions': self._definitions,
+            'parameters': self._parameters,
             'paths': self._paths,
         }
         ret.update(self.options)
         return ret
 
+    def add_parameter(self, _name, _in, **kwargs):
```
sloria: Not sure about the argument names here. Maybe `name` and `display_name`? is there a need to pull off `kwargs` rather than including the arguments explicitly in the function declaration? 

Also, use a trailing underscore to avoid clashing with a keyword, i.e. `in_`, as per PEP8.

apispec/core.py
```diff
@@ -19,16 +19,22 @@
...
 
     :param dict operations: Dict mapping status codes to operations
     """
+    get_ref = lambda x: x if isinstance(x, dict) else {'$ref': '#/parameters/' + x}
+
     for operation in (operations or {}).values():
-        for parameter in operation.get('parameters', []):
-            if parameter['in'] == 'path':
-                parameter['required'] = True
+        if 'parameters' in operation:
```
sloria, line 32: I personally liked the brevity of 

``` python
 for parameter in operation.get('parameters', []):
```

Is there a specific reason this was changed?

sloria:
Thanks for the PR. This is a nice feature. Could you please add tests for the new method?

apispec/core.py
jta, line 32: I changed it in order to pass the current test suite. The previous code inserted an empty list  for `operation['parameters']` if it did not exist.

apispec/core.py
jta: Have tried to address this in latest commit. Have replaced `in` for `location`, which is consistent with the swagger spec.

The reason I pulled from kwargs was because the spec for parameters is relatively [complex](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#parameter-object). I suspect it wouldn't make sense to validate whether the caller is constructing a valid parameter at this point, since it would be possible for a parameter to be defined through alternat code paths.

jta:
Will add some tests once I get more time, thanks!

jta:
I've added a test and rebased + squashed the commits.

sloria:
Thanks!

PR #38, opened 2015-11-23
Dump-only fields must be skipped if 'dump=False' is passed

frol:


apispec/utils.py
```diff
@@ -98,4 +98,4 @@ def validate_swagger(spec):
                 stderr=subprocess.STDOUT,
             )
         except subprocess.CalledProcessError as error:
-            raise exceptions.SwaggerError(error.output.decode('utf-8'))
+            raise exceptions.PluginError(error.output.decode('utf-8'))
```
sloria: Ah, good catch. I think the proper fix is to add `SwaggerError` to `exceptions`. A `PluginError` is raised for plugin-related code.

sloria:
Thanks! This looks good. Once the SwaggerError is added, this should be good to merge.

apispec/utils.py
frol: Please, go ahead and fix this the way you like it. I have only fixed it so the tests would pass (I don't know how they did pass before).

frol: I'm just not really sure if swagger mentions should be somewhere but in `ext/marshmallow/swagger.py`.

frol:
@sloria Added SwaggerError as you pointed out.

sloria:
Thanks!

PR #39, opened 2015-11-26
Fixed #37: Introspect fields of an instance (if passed) instead of falling back to a class

### LINKED ISSUE #37: Apispec doesn't play nicely with the schemas where fields are generated/modified on the fly
The following use-case doesn't work for me:

``` Python
from apispec.ext.marshmallow.swagger import schema2jsonschema

class A(Schema):
    q = fields.Integer()
    def __init__(self, default_q=None):
        super(A, self).__init__()
        self.fields['q'].default = default_q

class B(Schema):
    a = fields.Nested(A(default_q=1))

schema2jsonschema(B)
```

The output jsonschema is:

``` Python
{
    'properties': {
        'a': {
             'properties': {
                 'q': {
                     'format': 'int32',
                     'type': 'integer'
                 }
            }
        }
    }
}
```

I expect to see:

``` Python
{
    'properties': {
        'a': {
             'properties': {
                 'q': {
                     'format': 'int32',
                     'type': 'integer',
                     'default': 1
                 }
            }
        }
    }
}
```

Here are the relevant source code lines:

Only a Nested schema **class** (not instance) will be analyzed for the Nested fields:
https://github.com/marshmallow-code/apispec/blob/dev/apispec/ext/marshmallow/swagger.py#L105

Only fields that are declared on class definition will be analyzed when running [`schema2parameters`](https://github.com/marshmallow-code/apispec/blob/dev/apispec/ext/marshmallow/swagger.py#L124) and [`schema2jsonschema`](https://github.com/marshmallow-code/apispec/blob/dev/apispec/ext/marshmallow/swagger.py#L221)

frol:
More details are in #37.

### DIFF
--- a/apispec/ext/marshmallow/swagger.py
+++ b/apispec/ext/marshmallow/swagger.py
@@ -102,7 +102,7 @@ def field2property(field, spec=None, use_refs=True, dump=True):
         elif spec:
             schema = resolve_schema_dict(spec, field.schema)
         else:
-            schema = schema2jsonschema(field.schema.__class__)
+            schema = schema2jsonschema(field.schema)
         if field.many:
             ret['type'] = 'array'
             ret['items'] = schema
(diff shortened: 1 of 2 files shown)


sloria:
Thanks!

PR #40, opened 2015-12-03
Introduce handling of plugins with different and conflicting signatures

AndreiPashkin:
Subj. If there are multiple plugins with different signatures - Apispec fails. This patch fixes it.


sloria:
Thanks!

PR #43, opened 2015-12-21
Adjust field name in spec according to `load_from` and `dump_to`.

jta:
In marshmallow you can work around clashes with reserved names (e.g.
`global`) by using the `load_from` and `dump_to` attributes in a field
to effectively override the actual field name.

Doing so implicitly changes the schema however, and this commit attempts
to correct for that by replacing the field name with the values of
`load_from` and `dump_to` if they are the same.

An open question is what to do if `dump_to` and `load_from` differ. In
this case in practice the input schema differs from the output schema,
but there is no simple way of generating two schemas for the same
definition, and it's probably too much of an edge case to bother with.

### DIFF
--- a/apispec/ext/marshmallow/swagger.py
+++ b/apispec/ext/marshmallow/swagger.py
@@ -42,6 +42,20 @@ FIELD_MAPPING = {
 }
 
 
+def _observed_name(field, name):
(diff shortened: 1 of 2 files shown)


sloria:
Looks good. Thanks!

PR #45, opened 2016-01-25
change README example code to generate valid Swagger

max-orhai:
The key "description" should not appear at the root of valid Swagger 2.0 JSON. Also, a comma was missing from the commented output, so it wasn't even valid Python.

See https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields

The Swagger validation service is available at 
`http://online.swagger.io/validator/debug?url=<your swagger.json URL goes here>`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -32,7 +32,6 @@ Example Application
     spec = APISpec(
         title='Swagger Petstore',
         version='1.0.0',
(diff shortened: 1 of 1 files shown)


sloria:
Thanks! I'll go ahead and make the same change in the docs.

PR #47, opened 2016-01-27
Bug fix for observed field names.

benbeadle:
When the field is required, the observed field name should be used instead of the real field name.


sloria:
Thanks for the PR. Would you mind adding a regression test for this change?

sloria:
Replaced by #48

PR #48, opened 2016-01-29
Bug fix for observed field names.

benbeadle:
When the field is required, the observed field name should be used instead of the real field name.

@sloria - I couldn't figure out how to edit the other pull request, so I created a new request with the added regression test you asked for in #47.


sloria:
Thanks!

PR #50, opened 2016-02-10
APISpec.add_path now accepts Path objects

### LINKED ISSUE #49: How should I add an existing Path instance to an APISpec instance
I tried to do something like:

``` python
spec = APISpec(**kwargs)
spec.add_path(Path(**kwargs))
```

And I received the following error (as I should have)

```
File "/Users/Josh/Developer/Kaplan/AWS-Lambda-APIGateway-POC/env/lib/python2.7/site-packages/apispec/core.py", line 169, in add_path
    self._paths.setdefault(path.path, path).update(path)
TypeError: unhashable type: 'Path'
```

Is there an easy way to add an existing `Path` object or do I need to duplicate the logic of `self._paths.setdefault(path.path, path).update(path)`?

If this functionality seems worthwhile, I can submit a PR to update `APISpec.add_path` to accept `Path` objects.

hello-josh:
Had to do a little `basePath` dance, not as succinct as I'd like

Fixes #49

### DIFF
--- a/apispec/core.py
+++ b/apispec/core.py
@@ -136,11 +136,21 @@ class APISpec(object):
         """Add a new path object to the spec.
 
         https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#paths-object-
+
+        :param str|Path|None path: URL Path component or Path instance
+        :param dict|None operations: describes the http methods and options for `path`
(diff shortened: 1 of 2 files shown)


sloria:
Thanks!

PR #56, opened 2016-03-07
Tags

martinlatrille:
Added a "Tags" category to store information about tags, as seen in the Swagger petstore example ( http://petstore.swagger.io/v2/swagger.json ).

Also added elements in .gitignore, which are totally discardable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,3 @@
-.cache
-.idea
-env
-
 *.py[cod]
 
 # C extensions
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -1,3 +1,7 @@
+.cache
```
sloria: These belong in a [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore).

martinlatrille: Thanks for the tip, these are removed.

sloria:
Thanks!

PR #57, opened 2016-03-08
Definition Doc

martinlatrille:
A very tiny pull request, where I just added info about how to pass fields' metadatas in the definition specs by passing them in the kwargs of the field.

(I found out it was already implemented by looking in the code; didn't find any reference to this feature in the doc actually)

### DIFF
--- a/apispec/core.py
+++ b/apispec/core.py
@@ -182,6 +182,11 @@ class APISpec(object):
     def definition(self, name, properties=None, enum=None, **kwargs):
         """Add a new definition to the spec.
 
+        If you are using the Marshmallow extension, you can pass fields' metadatas in the fields' kwargs.
+        For example, if you want to add an enum to your field Status:
+
+        status = fields.String(required=True, enum=['open', 'closed']
+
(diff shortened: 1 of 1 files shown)


apispec/core.py
```diff
@@ -182,6 +182,13 @@ def add_path(self, path=None, operations=None, **kwargs):
     def definition(self, name, properties=None, enum=None, **kwargs):
         """Add a new definition to the spec.
 
+        If you are using the Marshmallow extension, you can pass fields'metadatas in the
```
sloria, line 185: Not sure if this should go here; this should probably go in the documentation for the marshmallow extension.

martinlatrille, line 185: In a way yes; in another way, from the example in the README, spec.definition is used to reference a Marshmallow Schema; chances are people will stop there and probably not go deeper in the doc than spec.definition (just like I did).

If it were to be moved in ext.marshmallow's doc anyway, where would you put it ?

sloria, line 185: Fair enough; I think it's ok to have this here for now.

PR #61, opened 2016-03-16
Bugfix: Nested Metadatas

martinlatrille:
This PR definitely fix Metadatas for Nested Fields.

On the current version, metadatas of a nested field with no refs would be overridden.
With this update, those are never removed.

(+ added backspaces in `marshmallow/__init__.py` for PEP8 compliance)


sloria:
Thanks!

PR #63, opened 2016-04-07
Fix parameter without 'in' on clean_operations

lucasrcosta:
The following is a valid YAML and breaks in the current code, since the param will be a `dict` but won't have `in`:

``` yaml
parameters:
  - in: path
    name: some_param
    required: true
    type: string
  - $ref: "#/parameters/ParamRef"
```


sloria:
Build failure is unrelated to these changes. Looks good. Thanks again!

PR #65, opened 2016-04-11
Support vendor extensions

lucasrcosta:
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#vendorExtensions
Useful e.g. for specifying `x-swagger-router-controller` when mocking requests.


sloria:
Thanks!

PR #70, opened 2016-04-28
[Fix #69] Support the flask APPLICATION_ROOT config

### LINKED ISSUE #69: Flask extension does not support application root
The flask extension only outputs the route path, but `flask.url_for()` takes into consideration that the path should be prefixed with the configured `APPLICATION_ROOT` path. [apispec/ext/flask.py#L58](https://github.com/marshmallow-code/apispec/blob/1076180/apispec/ext/flask.py#L58)
- http://flask.pocoo.org/docs/0.10/config/#builtin-configuration-values
- https://github.com/pallets/flask/blob/9f1be8e/flask/helpers.py#L309
- https://github.com/pallets/flask/blob/2bf477c/flask/app.py#L1761

deckar01:
Prefix the formatted swagger path with the `APPLICATION_ROOT` from the flask app's configuration settings.

I mentioned in [a comment on the issue](https://github.com/marshmallow-code/apispec/issues/69#issuecomment-215444641) that I preferred to lean on the `url_adapter` interface, but it does not expose a string transformation of the endpoint that does not require providing the values for the parameters.

> Note: I opted for the try/except import to support python 2 & 3, because [`future`](https://pypi.python.org/pypi/future) is not currently a dependency of the project.
> https://docs.python.org/3/howto/pyporting.html

### DIFF
--- a/apispec/ext/flask.py
+++ b/apispec/ext/flask.py
@@ -27,6 +27,11 @@ function to `add_path`. Inspects URL rules and view docstrings.
 from __future__ import absolute_import
 import re
 
+try:
+    from urllib.parse import urljoin
+except ImportError:
+    from urlparse import urljoin
+
 from flask import current_app
 
 from apispec.compat import iteritems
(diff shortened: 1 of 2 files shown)


sloria:
Looks great! Thanks @deckar01

PR #72, opened 2016-05-01
Marshmallow Schema Array

### LINKED ISSUE #71: Marshmallow extensions does not support plural schema instances
I expected

``` yaml
get:
  description: Get the list of pets.
  responses:
    200:
      description: A list of pet objects.
      schema:
        type: array
        items: PetSchema
```

to produce

``` json
"get": {
  "description": "Get the list of pets.",
  "responses": {
    "200": {
    "description": "A list of pet objects.",
    "schema": {
      "type": "array",
      "items": {
        "$ref": "#/definitions/Pet"
      }
    }
  }
}
```

but got

``` json
"get": {
  "description": "Get the list of pets.",
  "responses": {
    "200": {
    "description": "A list of pet objects.",
    "schema": {
      "type": "array",
      "items": "PetSchema"
    }
  }
}
```

The issue is that [apispec/ext/marshmallow/__init__.py#L77](https://github.com/marshmallow-code/apispec/blob/157682f/apispec/ext/marshmallow/__init__.py#L77) only checks if `many` is set to true on the class, but the more common marshmallow pattern is to set `many=True` when constructing schema instances. This also fails for the reverse case when the class is set to `many=True`, but the instance used for the view is constructed with `many=False` (although this seems like an anti-pattern).

Since the instance parameter overrides the class attribute, it should probably be considered a default, which means that it should be ignored and the format I used above should be used instead.

deckar01:
The marshmallow extension should support schema array definitions since it allows schema instances to override `many` during construction.
- [x] Tests #71 
- [x] Fixes #71
- [x] Update docs

### DIFF
--- a/apispec/ext/marshmallow/__init__.py
+++ b/apispec/ext/marshmallow/__init__.py
@@ -52,6 +52,63 @@ def schema_definition_helper(spec, name, schema, **kwargs):
 def schema_path_helper(spec, view, **kwargs):
     """Path helper that allows passing a Schema as a response. Responses can be
     defined in a view's docstring.
+    ::
+
+        from pprint import pprint
+
+        from my_app import Users, UserSchema
+
+        class UserHandler:
+            def get(self, user_id):
+                '''Get a user endpoint.
+                ---
+                description: Get a user
+                responses:
+                    200:
+                        description: A user
+                        schema: UserSchema
+                '''
+                user = Users.get(id=user_id)
+                schema = UserSchema()
+                return schema.dumps(user)
+
+        urlspec = (r'/users/{user_id}', UserHandler)
+        spec.add_path(urlspec=urlspec)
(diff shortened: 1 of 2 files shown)


sloria:
Looking good so far! Thanks @deckar01 .

deckar01:
@sloria The docs for `schema_path_helper` very briefly mention docstrings. I was thinking about examples for both schema patterns, but it might be more concise to link to the OpenAPI spec for doc strings and simply mention that it resolves schema names to OpenAPI references. Thoughts?

sloria:
I generally prefer examples over descriptions, but I think either approach is fine.

deckar01:
Here is a preview of the rendered docs:

schema_path_helper

sloria:
Looks good. Thanks again!

Issue #66, opened 2016-04-13
[marshmallow plugin] Inspect validators


sloria:
Many of marshmallow's validators can be inspected to provide additional data when constructing OpenAPI entities.

``` python
from marshmallow import fields, validate
from apispec.ext.marshmallow.swagger import field2property

field = fields.Int(validate=validate.Range(min=3, max=42))

prop = field2property(field)
assert prop['minimum'] == 3
assert prop['maximum'] == 42
```

deckar01:
There are 3 objects in the OpenAPI spec that support validation fields:
- [Parameter](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameter-object)
- [Items](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#items-object)
- [Header](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#header-object)

The header object only supports simple types, so there is no marshmallow schema to validate.

Matching validators:

| Marshmallow Validator | OpenAPI Validator |
| --- | --- |
| Email | pattern (specific) |
| Equal | enum (single) |
| Length (list?) | minItems, maxItems |
| Length (*) | minLength, maxLength |
| OneOf | enum |
| Range | minimum, maximum |
| Regexp | pattern |
| URL | pattern (specific) |

Unsupported marshmallow validators:

| Marshmallow Validator | OpenAPI Validator |
| --- | --- |
| ContainsOnly | (maybe enum?) |
| NoneOf |  |
| Predicate |  |
| Validator | (user defined?) |

It should be possible to design this mapping in such a way that user defined marshmallow validators could register their own OpenAPI validator mapping with apispec.

Unsupported OpenAPI validators:

| Marshmallow Validator | OpenAPI Validator |
| --- | --- |
|  | exclusiveMinimum |
|  | exclusiveMaximum |
|  | uniqueItems |
|  | multipleOf |

Maybe there should be an issue for marshmallow to support these validators, so that there can be a subset of marshmallow validators that completely cover the OpenAPI spec.

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameter-object
https://marshmallow.readthedocs.io/en/latest/api_reference.html#module-marshmallow.validate

lafrech:
`allow_none` is somehow related, although it is not a `Validator`.

[...]

PR #73, opened 2016-05-03
Marshmallow Extension Validation Fields

deckar01:
#66 Translate the following marshmallow validators to OpenAPI validation fields:
- [x] ~~Length (list)~~ (redundant)
  - ~~minItems~~
  - ~~maxItems~~
- [x] Length
  - minLength
  - maxLength
- [x] ~~OneOf~~ (already implemented)
  - ~~enum~~
- [x] Range
  - minimum
  - maximum

Punted
-  ~~Email~~
  - ~~pattern~~
- Equal
  - enum (single)
- Regexp
  - pattern
- ~~URL~~
  - ~~pattern~~

### DIFF
--- a/apispec/ext/marshmallow/swagger.py
+++ b/apispec/ext/marshmallow/swagger.py
@@ -110,6 +110,47 @@ def field2range(field):
                 attributes['maximum'] = validator.max
     return attributes
 
+def field2length(field):
+    """Return the dictionary of swagger field attributes for a set of
+    :class:`Length <marshmallow.validators.Length>` validators.
+
+    :param Field field: A marshmallow field.
+    :rtype: dict
+    """
+    validators = [
+        validator for validator in field.validators
+        if (
+            hasattr(validator, 'min') and
+            hasattr(validator, 'max') and
+            hasattr(validator, 'equal')
+        )
+    ]
+
+    attributes = {}
+    for validator in validators:
+        if not validator.min is None:
+            if hasattr(attributes, 'minLength'):
+                attributes['minLength'] = max(
+                    attributes['minLength'],
+                    validator.min
+                )
+            else:
+                attributes['minLength'] = validator.min
+        if not validator.max is None:
+            if hasattr(attributes, 'maxLength'):
+                attributes['maxLength'] = min(
+                    attributes['maxLength'],
+                    validator.max
+                )
+            else:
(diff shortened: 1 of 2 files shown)


sloria:
Thanks for looking into this, @deckar01 !

deckar01:
@sloria I have the range validator working! 🎉 

The email and URL validators may prove to be less plausible than I expected. They aren't subclasses of `Regexp` or single regex that can be serialized. Even if they were serialized into a "pattern" they would be pretty unreadable.

deckar01:
@sloria After looking at the implementation for `field2choices()`, I realized my implementation does not yet combine multiple validation fields. I am considering using the design pattern of `field2choices()` and creating utility methods for each validator type to encapsulate how they are reduced to a single validation field.

One thing I don't like about `field2choices()` is that it checks for the presence of the `choices` attribute instead of just checking if it is a sub/class instance of `validate.OneOf`.

Thoughts?

sloria:
@deckar01 Encapsulating the code into separate functions makes sense.

> One thing I don't like about field2choices() is that it checks for the presence of the choices attribute instead of just checking if it is a sub/class instance of validate.OneOf

This is following the Pythonic tendency towards duck typing over `isinstance`. This way, user validators that implement a `choices` attribute will work with `field2choices`.

deckar01:
That makes sense. Although after looking at [`Range`](https://github.com/marshmallow-code/marshmallow/blob/dev/marshmallow/validate.py#L159) and its subclass [`Length`](https://github.com/marshmallow-code/marshmallow/blob/dev/marshmallow/validate.py#L201), it is going to be a little weird differentiating the two with duck typing. Technically a `Length` looks like a  `Range`, but due to additional properties it must be excluded.

sloria:
Fair enough; duck typing may not be practical for Range/Length.

deckar01:
The issue stems from a symantically invalid inheritance model for those two validators. I opened an issue to correct it. https://github.com/marshmallow-code/marshmallow/issues/455

deckar01:
@sloria I think this PR is ready for review. I would like to land just `Length` and `Range` in this PR to keep thing rolling and punt the rest of the validators to their own PRs.

sloria:
Looks great! Thanks a lot for the PR.

PR #80, opened 2016-06-24
Dict Field Support

ericb:
Adds `marshmallow.fields.Dict` to the known field mappings.


sloria:
Build is failing due to invoke incompatibility. This has since been fixed on `dev`, so I've cherry-picked your commit into `dev`. Thanks for the contribution!

PR #83, opened 2016-07-01
Ensures nested properties with dump_only are excluded

### LINKED ISSUE #82: dump_only is ignored in nested schemas for parameters
Given the following:

``` python
class ParentSchema(Schema):
    id = fields.Int(dump_only=True)
    name = fields.String()
    child = fields.Nested('ChildSchema')

class ChildSchema(Schema):
    id = fields.Int(dump_only=True)
    name = fields.String()
```

The parameter spec is generated:  

``` json
"parameters": [
  {
    "in": "body", 
    "name": "body", 
    "required": false, 
    "schema": {
      "properties": {
        "child": {
          "properties": {
            "id": {
              "format": "int32", 
              "type": "integer"
            }, 
            "name": {
              "type": "string"
            }
          }, 
          "type": "object"
        }, 
        "name": {
          "type": "string"
        }
      }, 
      "type": "object"
    }
  }
]
```

The Parent "id" is omitted but the Child "id" remains. 

Will submit a PR on this.

incognick:
Resolves #82

### DIFF
--- a/apispec/ext/marshmallow/swagger.py
+++ b/apispec/ext/marshmallow/swagger.py
@@ -250,7 +250,7 @@ def field2property(field, spec=None, use_refs=True, dump=True, name=None):
         elif spec:
             ret.update(resolve_schema_dict(spec, field.schema))
         else:
-            ret.update(schema2jsonschema(field.schema))
+            ret.update(schema2jsonschema(field.schema, dump=dump))
     elif isinstance(field, marshmallow.fields.List):
         ret['items'] = field2property(field.container, spec=spec, use_refs=use_refs, dump=dump)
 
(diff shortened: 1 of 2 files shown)


apispec/ext/marshmallow/swagger.py
```diff
@@ -250,7 +250,7 @@ def field2property(field, spec=None, use_refs=True, dump=True, name=None):
         elif spec:
             ret.update(resolve_schema_dict(spec, field.schema))
```
incognick, line 251: line 251:  Does this line need `dump=dump` as well?  My test passes without so I didn't add this.

sloria:
Thanks!

apispec/ext/marshmallow/swagger.py
sloria, line 251: Yes, turns out it should be passed here as well; done in https://github.com/marshmallow-code/apispec/commit/cbe3f33b1f89dab2124d41d663e7633c8d6c8617

incognick:
Awesome! Thank you

PR #86, opened 2016-07-30
Maintain order of fields in ordered schemas

ranjanashish:
Respect `ordered` Meta setting in schemas so that generated swagger JSON has appropriate order of fields


sloria:
Thanks!

PR #87, opened 2016-08-09
Maintain order in which paths are added to spec

ranjanashish:
Providing API documentation in a particular order, in general, makes the API design more obvious and API usage more easier for the users. 

For example, in petstore documentation, what if the API `/pet/{petId}/uploadImage` was mentioned before the API for adding a new pet to the store. In a quick glance, the end user would not be so happy to see the jumbled APIs.

swagger-petstore

Currently, the generated spec (via `APISpec.to_dict`) does not maintain the order in which paths were added (via `APISpec.add_path`) to `APISpec`.

### DIFF
--- a/apispec/core.py
+++ b/apispec/core.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 """Core apispec classes and functions."""
 import re
+from collections import OrderedDict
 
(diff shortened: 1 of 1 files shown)


sloria:
Thanks again!

PR #89, opened 2016-09-14
Add 'query' to __location_map__'

lafrech:
Adding query to **location_map** allows users to use 'query' as location.

The goal here is not just to save 6 characters writing 'query' instead of 'querystring'. The idea is to be coherent with webargs, which allows 'query' as an equivalent for 'querystring', thus facilitate integration of apispec with webargs.


sloria:
Thanks!

PR #92, opened 2016-09-18
Normalize paths relative to basePath in all cases

### LINKED ISSUE #78: Core path does not support swagger basePath
Related to #69.

Flask allows configuring the `APPLICATION_ROOT`, which provides the base path for the routes. #70 fixed this issue so that the flask extension reports the fully qualified path.

When the specs are dumped as swagger specs using `to_dict()`, the full path is shown on every route, even though the swagger specification supports the [`basePath`](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields) field on the root object.

I think it is important to keep the meaningful portion of the route URL separate from the base path to maintain readable documentation. #70 was the right fix for the flask extension since the `apispec` core assumes paths are absolute, but the fact that both `flask` and `swagger` support a configurable base path indicates that the `apispec` core should probably internalize the concept of base paths and expose them to extensions.

Examples of base path in similar libraries:
- [Bottle: ResourceManager (default) path](http://bottlepy.org/docs/dev/api.html#bottle.ResourceManager)
- [RAML: Base URI](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md/#base-uri-and-base-uri-parameters)

> Edit: Updated to refect that generating swagger specs is a core functionality of `apispec`, not an extension.

asteinlein:
Paths that include the basePath as set in options on the APISpec
instance were normalized to not have the basePath as part of each path.
However, this didn't work with the Flask plugin as normalization occured
before the plugin generated it's URLs.

This fixes the issue by also normalizing paths after plugin processing in
`add_path`.

Fixes #78.

### DIFF
--- a/apispec/core.py
+++ b/apispec/core.py
@@ -154,12 +154,19 @@ class APISpec(object):
         :param dict|None operations: describes the http methods and options for `path`
         :param dict kwargs: parameters used by any path helpers see :meth:`register_path_helper`
         """
+        def normalize_path(path):
+            if path and 'basePath' in self.options:
+                pattern = '^{0}'.format(re.escape(self.options['basePath']))
+                path = re.sub(pattern, '', path)
+
+            return path
+
         p = path
         if isinstance(path, Path):
             p = path.path
-        if p and 'basePath' in self.options:
-            pattern = '^{0}'.format(re.escape(self.options['basePath']))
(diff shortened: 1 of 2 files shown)


sloria:
Apologies for the delay in getting to this. The behavior looks correct. Thanks!

PR #93, opened 2016-10-12
Added handling for description in model definitions

martinlatrille:
Allow to set a description when declaring a definition of a Schema (as described [here](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schemaObject)).

```
spec.definition(
    'Pet',
    schema=PetSchema,
    description="An animal which lives with humans."
)
```

```
"Pet": {
    "description": "An animal which lives with humans.",
    "properties": {...},
    ...
}
```

### DIFF
--- a/apispec/core.py
+++ b/apispec/core.py
@@ -200,7 +200,7 @@ class APISpec(object):
 
         self._paths.setdefault(path.path, path).update(path)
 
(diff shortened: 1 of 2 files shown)


sloria:
Looks good. Thanks!

PR #94, opened 2016-10-24
Allow Patterned Objects

martinlatrille:
Allow to add attributes to fields which are not in `_VALID_PROPERTIES` if their key starts with `x-` (as written everywhere in https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md)


sloria:
+1 for adding support for x-\* properties.

It would be nice if they could be passed in without having to unpack (e.g. `x_foo=42` vs. `**{'x-foo': 42}`. That doesn't have to block merging this though. This looks good to merge as is. Thanks!

PR #96, opened 2016-11-02
Update the condition to show warning about Meta.fields

### LINKED ISSUE #95: fields2jsonschema warning about Shema Object
Hi! Can you help me to understand the warning I'm getting.

```
... swagger.py:450: UserWarning: Only explicitly-declared fields will be included in the Schema Object. Fields defined in Meta.fields or Meta.additional are ignored.
  "Only explicitly-declared fields will be included in the Schema Object. "
```

My Schema

```python
class BaseUserSchema(ModelSchema):
    """
    Base user schema exposes only the most general fields.
    """

    class Meta:
        # pylint: disable=missing-docstring
        model = User
        fields = (
            User.id.key,
            User.username.key,
            User.first_name.key,
            User.middle_name.key,
            User.last_name.key,
        )
        dump_only = (
            User.id.key,
        )
```

And here is the sample of code where I'm getting the warning.

```python
from apispec.ext.marshmallow.swagger import schema2jsonschema
pprint(schema2jsonschema(BaseUserSchema))

... swagger.py:450: UserWarning: Only explicitly-declared fields will be included in the Schema Object. Fields defined in Meta.fields or Meta.additional are ignored.
  "Only explicitly-declared fields will be included in the Schema Object. "
{'properties': {'first_name': {'maxLength': 30, 'type': 'string'},
                'id': {'format': 'int32', 'type': 'integer'},
                'last_name': {'maxLength': 30, 'type': 'string'},
                'middle_name': {'maxLength': 30, 'type': 'string'},
                'username': {'maxLength': 80, 'type': 'string'}},
 'required': ['username'],
 'type': 'object'}
```

The result of function is as expected. But if I understand the warning right my fields declared in `Meta.fields` should be ignored.

Can you help me to understand what is this warning about?

/cc @frol

khorolets:
Closes #95: fields2jsonschema warning about Shema Object

If we have `ModelShema` with:

```python
class Meta:
        # pylint: disable=missing-docstring
        model = User
        fields = (
            User.id.key,
            User.username.key,
            User.first_name.key,
            User.middle_name.key,
            User.last_name.key,
            'woopy',
        )
        dump_only = (
            User.id.key,
        )
```

The field `'woopy'` doesn't exist in model.

Then in swagger.json we'll have:
```json
...
"properties": {
                        "first_name": {
                            "maxLength": 30,
                            "type": "string"
                        },
                        "id": {
                            "format": "int32",
                            "readOnly": true,
                            "type": "integer"
                        },
                        "last_name": {
                            "maxLength": 30,
                            "type": "string"
                        },
                        "middle_name": {
                            "maxLength": 30,
                            "type": "string"
                        },
                        "username": {
                            "maxLength": 80,
                            "type": "string"
                        },
                        "woopy": {
                            "type": "string"
                        }
                    },
...
```

The `'woopy'` is there as you can see.

Also if we have Shema:

```python
class User(Schema):
    class Meta:
        fields = (
            'name',
            'last_name',
        )
```

```python
>>> from app.simple import User
>>> from apispec.ext.marshmallow.swagger import schema2jsonschema
>>> from pprint import pprint
>>> pprint(schema2jsonschema(User))
... site-packages/apispec/ext/marshmallow/swagger.py:450: UserWarning: Only explicitly-declared fields will be included in the Schema Object. Fields defined in Meta.fields or Meta.additional are ignored.
  "Only explicitly-declared fields will be included in the Schema Object. "
{'properties': {}, 'type': 'object'}
```

As you can see the warning is there and the fields are not.

I think we don't need that warning if Meta has `model` and we do need it when it's not.

### DIFF
--- a/apispec/ext/marshmallow/swagger.py
+++ b/apispec/ext/marshmallow/swagger.py
@@ -454,10 +454,14 @@ def fields2jsonschema(fields, schema=None, spec=None, use_refs=True, dump=True, 
     """
     Meta = getattr(schema, 'Meta', None)
     if getattr(Meta, 'fields', None) or getattr(Meta, 'additional', None):
-        warnings.warn(
-            "Only explicitly-declared fields will be included in the Schema Object. "
-            "Fields defined in Meta.fields or Meta.additional are ignored."
-        )
+        declared_fields = set(schema._declared_fields.keys())
+        if (set(getattr(Meta, 'fields', set())) > declared_fields
+            or
+            set(getattr(Meta, 'additional', set())) > declared_fields):
+            warnings.warn(
+                "Only explicitly-declared fields will be included in the Schema Object. "
+                "Fields defined in Meta.fields or Meta.additional are ignored."
+            )
 
     jsonschema = {
         'type': 'object',


apispec/ext/marshmallow/swagger.py
```diff
@@ -453,7 +453,7 @@ class Meta:
     :rtype: dict, a JSON Schema Object
     """
     Meta = getattr(schema, 'Meta', None)
-    if getattr(Meta, 'fields', None) or getattr(Meta, 'additional', None):
+    if (getattr(Meta, 'fields', None) or getattr(Meta, 'additional', None)) and not getattr(Meta, 'model', None):
```
frol: This line is too long.

khorolets: @sloria Can you guide me?

``` python
if (
        getattr(Meta, 'fields', None) or getattr(Meta, 'additional', None)
    ) and not getattr(Meta, 'model', None):
```

or 

``` python
if (getattr(Meta, 'fields', None) or getattr(Meta, 'additional', None)) \
        and not getattr(Meta, 'model', None):
```

stasfilin: @khorolets I think 

``` python
if (getattr(Meta, 'fields', None) or getattr(Meta, 'additional', None)) \
        and not getattr(Meta, 'model', None):
```

better

You can check this https://www.python.org/dev/peps/pep-0008/

``` python
# No extra indentation.
if (this_is_one_thing and
    that_is_another_thing):
    do_something()

# Add a comment, which will provide some distinction in editors
# supporting syntax highlighting.
if (this_is_one_thing and
    that_is_another_thing):
    # Since both conditions are true, we can frobnicate.
    do_something()

# Add some extra indentation on the conditional continuation line.
if (this_is_one_thing
        and that_is_another_thing):
    do_something()
```

khorolets: @stasfilin Thank you.

@sloria I've updated PR

/cc @frol

sloria:
Thanks for the PR @khorolets . I wonder if there is a more direct way to check if there are implicitly-created fields--perhaps checking Schema._declared_fields and taking the difference with Meta.fields. 

@khorolets Would you like to investigate this?

khorolets:
@sloria I can try.

khorolets:
@sloria I've tried to check it more directly with the usage of `Schema._declared_fields` and now the warning should appear only if there at least one field that should be ignored, otherwise the warning doesn't show up.

apispec/ext/marshmallow/swagger.py
```diff
@@ -454,10 +454,14 @@ class Meta:
     """
     Meta = getattr(schema, 'Meta', None)
     if getattr(Meta, 'fields', None) or getattr(Meta, 'additional', None):
```
frol, line 456: I don't think there is much value in this check now.

apispec/ext/marshmallow/swagger.py
```diff
@@ -454,10 +454,14 @@ class Meta:
...
-        warnings.warn(
-            "Only explicitly-declared fields will be included in the Schema Object. "
-            "Fields defined in Meta.fields or Meta.additional are ignored."
-        )
+        declared_fields = set(schema._declared_fields.keys())
+        if (set(getattr(Meta, 'fields', {})) > declared_fields
```
frol: I suggest you use `set()` instead of `{}` as the latter creates a `dict`, and you end up with `dict->set` type casting.

apispec/ext/marshmallow/swagger.py
khorolets: Some tests are failing without it.

sloria:
Looks good. Thanks @khorolets for the PR and thanks @frol for reviewing!

PR #98, opened 2016-11-09
Add array type checking for field2length method

### LINKED ISSUE #97: minLength/maxLength being added to all types
Marshmallow allows me to add validation against nested schema length, doing as the follows:

```python
class FooSchema(Schema):
    bars = fields.Nested('BarSchema', validate=Length(min=1), many=True)
```

It'll validate if the `bars` array contains at least 1 object, but using apispec to generate the spec for this schema, it'll output the follow:

```json
{
  "definitions": {
    "FooSchema": {
      "properties": {
        "bars": {
          "items": {
            "$ref": "#/definitions/BarSchema"
          },
          "minLength": 1,
          "type": "array"
        }
      }
    }
  }
}
```

Based on swagger specification http://json-schema.org/latest/json-schema-validation.html#anchor26 (5.6/5.7) the `minLength` and `maxLength` can only be used for string types, in other words these fields can't be used with arrays, so the [field2length](https://github.com/marshmallow-code/apispec/blob/dev/apispec/ext/marshmallow/swagger.py#L84) needs checking if the value is an array to properly add length validators.

YuriHeupa:
field2length now validates if the field is an array instance to properly add length validators

Fixes #97

### DIFF
--- a/apispec/ext/marshmallow/swagger.py
+++ b/apispec/ext/marshmallow/swagger.py
@@ -124,6 +124,8 @@ def field2length(field):
     :param Field field: A marshmallow field.
     :rtype: dict
     """
+    attributes = {}
+
     validators = [
         validator for validator in field.validators
         if (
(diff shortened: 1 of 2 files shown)


sloria:
Thanks!

Issue #84, opened 2016-07-19
dump_only attributes passed in Meta options are not marked as readOnly


sloria:
Reported by @lafrech in https://github.com/marshmallow-code/apispec/pull/81#issuecomment-233417809:

---

I noticed this only adds readOnly if the field was provided dump_only attribute directly. If the field appears in the dump_only list in Meta, readOnly is not added.

Is this known/intended?

More precisely, it "does not work" if spec.definition is passed a marshmallow.schema.SchemaMeta but it "works" if it is passed a Schema instance:

class GistSchema(ma.Schema):
    id = ma.fields.Int()
    content = ma.fields.Str()

```
class Meta:
    dump_only = ('content',)
```
# readOnly not added

spec.definition('Gist', schema=GistSchema)
# readOnly added

spec.definition('Gist', schema=GistSchema())
I just begun experimenting with apispec and I'm not sure passing an instance is the right thing to do. In the docs, a SchemaMeta is passed.

sloria:
The workaround for this is to pass a schema instance rather than a class:

```python
spec.definition('Gist', schema=GistSchema())
```

because the `dump_only` attribute gets set on the fields upon schema instantiation.

lafrech:
Yes.

I had a look at this issue when it was submitted and didn't find any obvious solution.

This workaroud works but it breaks [schema resolving in marshmallow extension](https://github.com/marshmallow-code/apispec/blob/dev/apispec/ext/marshmallow/__init__.py#L134). The mechanism relies on Schema classes being stored, not instances.

I would like to suggest the following change to `schema_definition_helper` in marshmallow extension, so that 

- the schema instance is used to create the json description (this workaround)
- the schema class is stored (needed for further reference resolving)

```python
def schema_definition_helper(spec, name, schema, **kwargs):
    """Definition helper that allows using a marshmallow
    :class:`Schema <marshmallow.Schema>` to provide OpenAPI
    metadata.

    :param type schema: A marshmallow Schema class or instance.
    """

    if isinstance(schema, type):
        schema_cls = schema
        schema_instance = schema()
    else:
        schema_cls = type(schema)
        schema_instance = schema

    # Store registered refs, keyed by Schema class
    plug = spec.plugins[NAME]
    if 'refs' not in plug:
        plug['refs'] = {}
    plug['refs'][schema_cls] = name
    return swagger.schema2jsonschema(schema_instance, spec=spec, name=name)
```

[...]

PR #101, opened 2016-12-01
Exclude fields that are both dump_only and load_only

lafrech:
In Marshmallow, a field can be both `dump_only` and `load_only`, which means don't serialize and don't deserialize (see [nice explanation here](https://github.com/marshmallow-code/marshmallow/issues/453#issuecomment-221673132)).

Shouldn't such a field be removed from the spec just like an "exclude" field ?

Here's a quick attempt.

Note that 

- It has not unit tests (just tested in my use case).

- It only deals with `dump_only` / `load_only` properties defined in the field, not in `Meta`.

The fix to the latter might also address https://github.com/marshmallow-code/apispec/issues/84.

I'd rather have feedback about this before going any further.

Also, I'm not sure I understand the rationale behind the `dump` parameter correctly. I wouldn't like to screw it up, so any enlightenment welcome.

Thanks.

### DIFF
--- a/apispec/ext/marshmallow/swagger.py
+++ b/apispec/ext/marshmallow/swagger.py
@@ -347,7 +347,7 @@ def fields2parameters(fields, schema=None, spec=None, use_refs=True, dump=True,
             for field_name, field_obj in iteritems(fields)
             if (
                 (field_name not in exclude_fields)
-                and not (field_obj.dump_only and not dump)
+                and not (field_obj.dump_only and (not dump or field_obj.load_only))
             )
     ]
 


sloria:
@lafrech Apologies for the delay in responding to this. I *think* the behavior you are proposing is correct. 

The `dump` parameter exists because there are different field attributes to inspect depending on whether a Schema is used for dumping or loading. For example, `Field.default` (used on dumping) vs. `Field.missing` (used on loading).

Once tests are added for this, I'd be glad to merge.

lafrech:
Reworked in https://github.com/marshmallow-code/apispec/pull/112.

PR #103, opened 2016-12-16
Tornado getargspec with decorated handler method arguments.

### LINKED ISSUE #102: Tornado getargspec with decorated handler method arguments.
Current implementation uses deprecated inspect.getargspec() method to get argument names.

It causes issues having decorated handler methods and you get decorator arguments instead of method arguments which breaks swagger path generation.

matijabesednik:
Current implementation uses deprecated inspect.getargspec() method to get argument names. Fixes #102 

It causes issues having decorated handler methods and you get decorator arguments instead of method arguments which breaks swagger path generation.

### DIFF
--- a/apispec/ext/tornado.py
+++ b/apispec/ext/tornado.py
@@ -31,6 +31,7 @@ object to `add_path`.
 """
 from __future__ import absolute_import
 import inspect
+import sys
 from tornado.web import URLSpec
 
 from apispec import Path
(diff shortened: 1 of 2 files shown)


apispec/ext/tornado.py
```diff
@@ -80,7 +81,10 @@ def tornadopath2swagger(urlspec, method):
     :param method: Handler http method
     :type method: function
     """
-    args = inspect.getargspec(method).args[1:]
+    if sys.version_info >= (3, 3):
+        args = [i for i in inspect.signature(method).parameters.keys()][1:]
```
sloria: This can be simplified to 

```python
list(inspect.signature(method).parameters.keys())[1:]
```

sloria (changes requested):
Thanks for the PR! Just a minor change to make and this is good to merge. Would you mind also adding yourself to AUTHORS.rst?

matijabesednik:
Requested changes done. :)

PR #106, opened 2017-01-17
Marshmallow ext.: use Schema instance to provide API metadata

lafrech:
Workaround for https://github.com/marshmallow-code/apispec/issues/84.


sloria:
Thanks!

PR #109, opened 2017-02-19
Translate allow_none property into x-nullable

lafrech:
This makes `allow_none` property in a field translate to `x-nullable` in the spec.

As I wrote in https://github.com/marshmallow-code/apispec/issues/66#issuecomment-280878572:

It would map to `nullable`, except [`nullable` does not exist... yet](https://github.com/OAI/OpenAPI-Specification/issues/229). There seems to be a wide consensus upon the use of `x-nullable` until `nullable is added` (OpenAPI 3). It has been [integrated in ReDoc](https://github.com/Rebilly/ReDoc/pull/95), for instance.


sloria:
Looks good. Thanks!

PR #110, opened 2017-03-02
Add extra fields to a definition

lafrech:
Sometimes, I need to add a field to a definition.

Typical use case: the `discriminator` field (search for "discriminator" in [there](http://swagger.io/specification/#schemaObject), [this comment](https://github.com/OAI/OpenAPI-Specification/issues/403#issuecomment-128824351) also explains a lot and points some limitations). This can also be a vendor field (`x-my_field`).

I created a [specific definition helper](https://github.com/Nobatek/flask-rest-api/blob/master/flask_rest_api/spec/plugin.py#L111) for that, but I think it could be in apispec itself. There is nothing specific in it. So here's the PR.

I first added this to marshmallow plugin `schema_definition_helper`, but then I realized it was not specific to marshmallow schemas, so I moved it to `APISpec.definition`. I'm leaving the two commits on purpose to expose the two approaches. I can squash into one or remove the last one depending on which approach is preferred.

This feature is pretty permissive. A more conservative approach would be to allow only fields that conform to the spec: a finite hardcoded list _à la_ [`_VALID_PROPERTIES`](https://github.com/marshmallow-code/apispec/blob/dev/apispec/ext/marshmallow/swagger.py#L172) + the possibility to pass any vendor field (`x-`). I'm not sure this is needed. After all, the user adds `extra_fields` explicitly, so he should know what he is doing. It's not like we're harvesting in some Schema's metadata. BTW, I don't think `discriminator` should be defined in a field's metadata. It is schema level stuff. It shouldn't be in `_VALID_PROPERTIES`, right?

I'd be happy to hear your feedback about this.

### DIFF
--- a/apispec/core.py
+++ b/apispec/core.py
@@ -200,7 +200,8 @@ class APISpec(object):
 
         self._paths.setdefault(path.path, path).update(path)
 
-    def definition(self, name, properties=None, enum=None, description=None, **kwargs):
+    def definition(self, name, properties=None, enum=None, description=None, extra_fields=None,
+                   **kwargs):
         """Add a new definition to the spec.
 
         .. note::
@@ -227,6 +228,8 @@ class APISpec(object):
             ret['enum'] = enum
         if description:
             ret['description'] = description
+        if extra_fields:
+            ret.update(extra_fields)
         self._definitions[name] = ret
 
     # PLUGIN INTERFACE
--- a/apispec/ext/marshmallow/__init__.py
+++ b/apispec/ext/marshmallow/__init__.py
@@ -34,13 +34,12 @@ from . import swagger
 NAME = 'apispec.ext.marshmallow'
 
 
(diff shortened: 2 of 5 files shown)


sloria:
Thanks @lafrech for the contribution . I think this serves a very common use case, so lgtm. 👍

sloria:
> This feature is pretty permissive. A more conservative approach would be to allow only fields that conform to the spec: a finite hardcoded list à la _VALID_PROPERTIES + the possibility to pass any vendor field (x-). I'm not sure this is needed.

I agree. For maintainability, I think it's best that apispec not try too hard to validate inputs. Perhaps we should add some docco that mentions this...

sloria:
> BTW, I don't think discriminator should be defined in a field's metadata. It is schema level stuff. It shouldn't be in _VALID_PROPERTIES, right?

Good catch. Fixed.

PR #111, opened 2017-03-03
Allow additional properties (dump_only,...) on Nested schemas using allOf

### LINKED ISSUE #108: How to specify attributes (such as dump_only,...) on Nested fields?
Say you have a dump_only Nested field in your model.

The model reads:

```python
class Parent(Schema):

    child = Nested(Child, dump_ony=True)
```

Assuming you added Child to the definitions, the spec then looks like

```
    "Parent": {
      "properties": {
        "child": {
          "$ref": "#/definitions/Child", 
          "readOnly": true
        },
      }, 
      "type": "object"
    }, 
```

Unfortunately, this is invalid OpenAPI spec: https://github.com/swagger-api/swagger-js/issues/402

I get this warning in ReDoc:

    Other properties are defined at the same level as $ref at "#/definitions/Parent/properties/child". They are IGNORED according to the JsonSchema spec

Has anyone here been facing this already?

Should we use the AllOf trick described [here](https://github.com/darklynx/request-baskets/issues/5)?

lafrech:
Fixes #108.

Uses `allOf` as proposed [here](https://github.com/darklynx/request-baskets/issues/5) to add properties to a ref field, so as to circumvent a [limitation in JsonSchema](https://github.com/swagger-api/swagger-js/issues/402).

### DIFF
--- a/apispec/ext/marshmallow/__init__.py
+++ b/apispec/ext/marshmallow/__init__.py
@@ -146,7 +146,9 @@ def resolve_schema_dict(spec, schema, dump=True):
                 'type': 'array',
                 'items': ref_schema,
             }
-        return ref_schema
+        return {
+            'allOf': [ref_schema]
+        }
     if not isinstance(schema, marshmallow.Schema):
         schema = schema_cls
     return swagger.schema2jsonschema(schema, spec=spec, dump=dump)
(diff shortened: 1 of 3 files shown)


sloria:
Looks good. Thanks @lafrech

PR #112, opened 2017-03-04
Load-only fields must be skipped if 'dump' is True

lafrech:
Symmetric to "Dump-only fields must be skipped if 'dump=False' is passed".
Also fixes cases where dump_only or load_only is specified in Meta.
_De facto_ removes fields both dump_only and load_only.

This replaces #101. I assumed load_only fields should be removed if dump is True. Now I'm wondering the point of parameters with dump=True. If parameters are inputs of the API, those should only be used for loading, so what is the use case for dump=True?

### DIFF
--- a/apispec/ext/marshmallow/swagger.py
+++ b/apispec/ext/marshmallow/swagger.py
@@ -337,6 +337,8 @@ def fields2parameters(fields, schema=None, spec=None, use_refs=True, dump=True,
         "Schemas with many=True are only supported for 'json' location (aka 'in: body')"
 
     exclude_fields = getattr(getattr(schema, 'Meta', None), 'exclude', [])
+    dump_only_fields = getattr(getattr(schema, 'Meta', None), 'dump_only', [])
+    load_only_fields = getattr(getattr(schema, 'Meta', None), 'load_only', [])
 
     return [
         field2parameter(
@@ -348,9 +350,10 @@ def fields2parameters(fields, schema=None, spec=None, use_refs=True, dump=True,
             default_in=default_in
         )
             for field_name, field_obj in iteritems(fields)
-            if (
-                (field_name not in exclude_fields)
-                and not (field_obj.dump_only and not dump)
+            if not (
+                field_name in exclude_fields or
+                (not dump and (field_obj.dump_only or field_name in dump_only_fields)) or 
+                (dump and (field_obj.load_only or field_name in load_only_fields))
             )
     ]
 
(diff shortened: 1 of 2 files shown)


lafrech:
Hold on. I should rework the tests. I should have done them in `TestMarshmallowSchemaToParameters.test_schema_body_with_dump_only`.

@frol, since you fixed dump_only in https://github.com/marshmallow-code/apispec/pull/38/files, you might have an idea about load_only and my question in comment above...

apispec/ext/marshmallow/swagger.py
```diff
@@ -348,9 +350,10 @@ def fields2parameters(fields, schema=None, spec=None, use_refs=True, dump=True,
             default_in=default_in
         )
             for field_name, field_obj in iteritems(fields)
-            if (
-                (field_name not in exclude_fields)
-                and not (field_obj.dump_only and not dump)
+            if not (
+                field_name in exclude_fields or
+                (field_name in dump_only_fields or field_obj.dump_only) and not dump or 
+                (field_name in load_only_fields or field_obj.load_only) and dump
```
frol: I would change the order of the conditions and add more parentheses (just to avoid possible typo-bugs in future):

```python
if not (
    field_name in exclude_fields or
    (not dump and (field_obj.dump_only or field_name in dump_only_fields)) or
    (dump and (field_obj.load_only or field_name in load_only_fields))
)
```

Lazy evaluation of conditions will help to avoid unnecessary `in` checking.

Other than that, I think it is a good change. Well done!

lafrech:
Thanks @frol for your feedback. Good point about performance. I reworked the PR.

Forget my comment about the tests, I was wrong. `TestMarshmallowSchemaToParameters.test_schema_body_with_dump_only` is absolutely not the right place. I'm not sure I did them in the right place, though.

@sloria, please ask for rework if needed.

sloria:
Looks good. Thanks again @lafrech

sloria:
@lafrech I've added you as a collaborator to this repo. No pressure to invest any more time in the project than you already have; just hoping to open up the door to co-maintenance.

Thanks for all your great contributions!

lafrech:
Thanks @sloria. I don't have anything to add right now, but I probably will, sooner or later. I use apispec in [flask-rest-api](https://github.com/Nobatek/flask-rest-api) (development ongoing).

Just to make sure I understand correctly, do you have an example of a use case for `schema2parameters` with `dump=True`? Aren't parameters always inputs, therefore, loaded, but never dumped? Or am I wrong about the use of `schema2parameters`?

sloria:
@lafrech After a quick glance, I think you are right--`dump` should never be`True` when translating parameters. I've opened a new issue to remove `dump` in #114

PR #116, opened 2017-03-07
Remove dump parameter from fields2parameters and field2parameter

### LINKED ISSUE #114: Remove dump parameter from `schema2parameters`
As found by @lafrech in #112 : `schema2parameters` should only be handling client input, so `dump` should never be `True`. We should verify this and remove the `dump` parameter if possible.

lafrech:
Fixes #114.

This can be a breaking change for people calling `fields2parameters` or `field2parameter` with `dump` parameter in their code. People calling `schema2parameters` with `dump=False` are not affected as it will be silently ignored.

### DIFF
--- a/apispec/ext/marshmallow/swagger.py
+++ b/apispec/ext/marshmallow/swagger.py
@@ -317,7 +317,7 @@ def schema2parameters(schema, **kwargs):
     return fields2parameters(fields, schema, **kwargs)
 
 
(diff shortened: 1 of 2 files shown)


lafrech:
Hmmm, I was wrong. People calling `schema2parameters` with `dump=False` are also affected because `kwargs` are passed to `fields2parameters`.

Do we need to add deprecation warnings?

sloria:
Give that we are modifying API that most users will not be using, we don't need to worry too much about breaking compat with this change.

Issue #119, opened 2017-03-15
Ignore `load_only` fields if `dump=True` in `fields2jsonschema`?


luisincrespo:
Hi,

I'm making use of https://github.com/jmcarp/flask-apispec to automatically generate docs in a personal flask project. This library makes use of your apispec swagger extension to generate docs for requests and responses. I figured out that responses in the generated docs where including `marshmallow`'s `load_only` fields, which (at least in my case) is not convenient. In the line that I'm linking below you're excluding `dump_only` fields if `dump=False` when invoking that method. Do you think it would be a good idea to also ignore `load_only` fields if `dump=True`?

https://github.com/marshmallow-code/apispec/blob/dev/apispec/ext/marshmallow/swagger.py#L492

I'm opening this for discussion, and I'll be happy to create a PR for that in case you are ok with the proposed functionality.

lafrech:
We might have an issue, here, because `schema_definition_helper` calls `schema2jsonschema` which in turn calls `fields2jsonschema`.

When adding a definition, we don't want to exclude fields. The schemas in definitions can be used for both input and output, so we want all the fields, with the dump_only marked `readOnly`, and the load_only... well we can't mark them `writeOnly` because this does not exist, and in practice the only workaround I know is to create two schemas registered separately.

I suppose you could use that workaround too (yes, it sucks...).

Or we could try to modify `fields2jsonschema` for that, but `dump` parameter as a boolean is unfit as we have three cases: dump_only, load_only, or dump_and_load (which should be the default).

Maybe the whole `dump` parameter thing ought to be reviewed to support those three cases. For instance, we could have dump_only (boolean) and load_only (boolean) as separate parameters. And having both set would raise an error (no point in having a schema used for neither dumping or loading).

@luisincrespo or anyone, please correct me if I'm wrong.

luisincrespo:
> Or we could try to modify fields2jsonschema for that, but dump parameter as a boolean is unfit as we have three cases: dump_only, load_only, or dump_and_load (which should be the default).

@lafrech as you said, "dump_and_load" is the default, so I think the `dump` boolean arg suffices, I mean:

- If you're creating a response definition (`dump=True`) you only have to worry about ignoring `load_only` fields, everything else shouldn't be ignored (unless inside `exclude`).
- If you're creating a request definition (`dump=False`) you only have to worry about ignoring `dump_only` fields, everything else shouldn't be ignored (unless inside `exclude`).

lafrech:
[...]

PR #121, opened 2017-03-16
If a field is load_only, exclude it from the response definition.

luisincrespo:
Related to #119


lafrech:
I proposed a test in https://github.com/marshmallow-code/apispec/pull/123 to show what this PR breaks. I tested this PR against that test and it fails.

Let's discus this in https://github.com/marshmallow-code/apispec/issues/119.

PR #122, opened 2017-03-17
Pass current state of the definition to definition_helpers in APISpec.definition

martinlatrille:
Now pass the current state of the definition (i.e. `ret`) as a `kwarg` to all definition helpers.

It will allow helpers to do modifications of nested value in the definition (like modifying a schema's property for example) without overriding the whole top-level dict (`properties`, in our example).

### DIFF
--- a/apispec/core.py
+++ b/apispec/core.py
@@ -269,8 +269,8 @@ class APISpec(object):
         - Return a `dict` representation of the definition's Schema object.
 
         The helper may define any named arguments after the `name` argument.
-        In `kwargs`, you will find among other things:
-        - Definition: the current state of the definition of the schema.
+        ``kwargs`` will include (among other things):
+        - definition (dict): current state of the definition
 
         https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#definitionsObject
 
--- a/apispec/core.py
+++ b/apispec/core.py
@@ -219,7 +219,7 @@ class APISpec(object):
         # Execute all helpers from plugins
         for func in self._definition_helpers:
             try:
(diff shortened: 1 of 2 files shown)


lafrech:
Looks good.

Note that this relies on the order in which the helpers are executed, which is the registration order: the order of the plugins in the list you pass at APISpec init. This is also the case for path helpers. This could be an issue if you need a different order for path helpers and definition helpers. But this is more of a theoretical issue, I don't think it would be blocking in practice. Anyway, there's no reason to block this PR until this is addressed.

Care to add a test before we can merge this?

Also, I suggest this change:

```python
- Definition: the current state of the definition of the schema
+ definition (dict): current state of the definition
```

apispec/core.py
```diff
@@ -269,6 +269,8 @@ def register_definition_helper(self, func):
         - Return a `dict` representation of the definition's Schema object.
 
         The helper may define any named arguments after the `name` argument.
+        In `kwargs`, you will find among other things:
```
sloria: To piggy-back on @lafrech 's suggestion, how about:

```
``kwargs`` will include (among other things):
- definition (dict): current state of the definition
```

sloria (changes requested):
Once the test and small doc change have been made, this is good to merge. Thanks for the PR, and thanks @lafrech for reviewing.

martinlatrille:
Thanks for the review guys, I've edited the doc and added a test!

apispec/core.py
martinlatrille: Done!

sloria (approved):

sloria:
Looks good. Thanks for making those changes.

PR #123, opened 2017-03-19
Test schema2jsonschema with dump_only/load_only fields

lafrech:
This test ensure that when creating a definition from a schema, both dump_only and load_only fields appear in the output.

- dump_only fields have a readOnly attribute
- load_only fields have no attribute because writeOnly does not exist

In fact, load_only is not really supported. The right way to define a schema with load_only fields would be to create another schema without the load_only fields. schema2jsonschema should not remove those fields. It is up to the user to either create a second schema or use some sort of custom attribute to express the load_only property.


sloria:
Thanks!

PR #124, opened 2017-04-17
sort list of reqd fields

dradetsky:
apropos of #113, i was recently trying to make some changes to something-or-other, and then diffing the generated swagger output to confirm that i hadn't actually screwed anything up. I noticed that the order of the requirements lists in generated definitions changed arbitrarily between runs, making it impossible to do this check. This change just sorts the required list.

There are 5 failing tests, but they're failing on dev too, and all have to do with tornado, which I can't see this having anything to do with.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -28,3 +28,4 @@ Contributors (chronological)
 - Yuri Heupa `@YuriHeupa <https://github.com/YuriHeupa>`_
 - Matija Besednik `@matijabesednik <https://github.com/matijabesednik>`_
 - Boris Serebrov `@serebrov <https://github.com/serebrov>`_
+- Daniel Radetsky `@dradetsky <https://github.com/dradetsky>`_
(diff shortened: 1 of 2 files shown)


dradetsky:
@sloria @jmcarp although i see the tests are passing on travis, i had 5 tornado-related tests fail locally. possibly, the list of dev-requirements is missing something, but i don't know enough about tornado to guess what.

dradetsky:
fwiw, i'm seeing `AttributeError: 'URLSpec' object has no attribute '_path'` and i note that the tornado dev-requirement is not pinned to a particular version, but that's all i can say.

dradetsky:
actually, i can also say

```
(venv3) % pip show tornado
Name: tornado
Version: 4.5
```

sloria:
`dev` has fixes for compatibility with the latest Tornado release, which is why the build is passing.

dradetsky:
_merges upstream_

fuckin' a!

should i merge upstream into this pr? is it otherwise acceptable?

sloria:
No need to merge upstream. Looks good to merge. Thanks!

PR #129, opened 2017-06-06
Correct definition to match output in document.

lamiskin:
Code was malformed, with un-even brackets and output was not matching.


sloria:
Good catch. Thanks!

PR #130, opened 2017-06-23
correctly fill default parameter if default is boolean False

nebularazer:
When using `False` as a default or missing parameter the swagger spec does not include a default.
This PR fixes this.
```python
from apispec.ext.marshmallow import swagger
from marshmallow import fields

field = fields.Boolean(default=False, missing=False)
swagger.field2property(field)
# {'type': 'boolean'}  # actual response
# {'default': False, 'type': 'boolean'}  # expected
```

### DIFF
--- a/apispec/ext/marshmallow/swagger.py
+++ b/apispec/ext/marshmallow/swagger.py
@@ -12,7 +12,7 @@ from collections import OrderedDict
 import copy
 
 import marshmallow
-from marshmallow.utils import is_collection
+from marshmallow.utils import is_collection, _Missing
 from marshmallow.compat import text_type, binary_type, iteritems
 from marshmallow.orderedset import OrderedSet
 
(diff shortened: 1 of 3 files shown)


apispec/ext/marshmallow/swagger.py
```diff
@@ -231,7 +231,7 @@ def field2property(field, spec=None, use_refs=True, dump=True, name=None):
         ret['format'] = fmt
 
     default = field.default if dump else field.missing
-    if default:
+    if not isinstance(default, _Missing):
```
sloria, line 234: Since `_Missing` is private API, you can compare against the `missing` singleton instead:

```python
from marshmallow import missing

# ...

if default is not missing:
```

sloria: I made this change when I merged.

PR #131, opened 2017-06-29
Allow default value for fields to be callable.

NightBlues:
Hi there! As docstring 
https://github.com/marshmallow-code/marshmallow/blob/dev/marshmallow/fields.py#L63 says the default value may be callable. I think it's a common use case, f.e. passing list or dict constructor to field.
```
class MySchema(Schema):
    numbers = fields.List(fields.Int, default=list)
```

### DIFF
--- a/apispec/ext/marshmallow/swagger.py
+++ b/apispec/ext/marshmallow/swagger.py
@@ -232,7 +232,10 @@ def field2property(field, spec=None, use_refs=True, dump=True, name=None):
 
     default = field.default if dump else field.missing
     if default is not marshmallow.missing:
-        ret['default'] = default
+        if callable(default):
+            ret['default'] = default()
+        else:
+            ret['default'] = default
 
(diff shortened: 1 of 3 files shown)


NightBlues:
@sloria, what do you think about this?)

sloria (changes requested):
Thanks @NightBlues for the PR. I think this makes sense, but I have a slight concern about the implementation. There's a possibility of swallowing a `TypeError` raised within the callable. We should probably just mirror the implementation from marshmallow:

```python
                    if callable(self.default):
                        return self.default()
                    else:
                        return self.default
```

sloria:
Once my comment above has been addressed and tests are added, this should be good to merge.

NightBlues:
@sloria is that better?

sloria (approved):

sloria:
Looks good to me. Thanks!

Issue #125, opened 2017-05-05
`APISpecError: Could not find endpoint for view` when using flask MethodViews


boosh:
I've seen #85 which seems to suggest Flask's `MethodViews` should be supported. I've added a URL rule with:
```
app.add_url_rule('/articles/<publication>', view_func=Articles.as_view('articles'))
```

But when I try to add a path for my schema with:
```
spec.definition('Article', schema=ArticleSchema)
spec.add_path(view=Articles)
```
I get an error:
```
apispec.exceptions.APISpecError: Could not find endpoint for view <class 'mypackage.resources.articles.Articles'>
```
I've tried passing various combinations of `Articles`, `Articles.as_view('articles')`, etc to `add_path` but nothing works. I've also tried the function in #85, but it still doesn't work. 

I'm assuming this is a bug?

djanderson:
This is my current solution to this problem. If there's interest, I can do a PR to `apispec.ext.flask`.

```python
from urllib.parse import urljoin

from apispec.ext.flask import _rule_for_view, flaskpath2swagger
from apispec import Path, utils
from flask import current_app

def path_from_method_view(spec, method_view, **kwargs):
    rule = _rule_for_view(method_view)
    path = flaskpath2swagger(rule.rule)
    app_root = current_app.config['APPLICATION_ROOT'] or '/'
    path = urljoin(app_root.rstrip('/') + '/', path.lstrip('/'))
    operations = {}
    for method in method_view.methods:
        method_name = method.lower()
        view = getattr(method_view.view_class, method_name)
        docstring_yaml = utils.load_yaml_from_docstring(view.__doc__)
        operations[method_name] =  docstring_yaml or dict()

    return Path(path=path, operations=operations or None)


spec.register_path_helper(path_from_method_view)


with application.test_request_context():
    for view in views:  # where `views` is an iterable of your `MethodView.as_view` funcs
        spec.add_path(method_view=view)  # now just use `method_view=view` instead of `view=view`
```

sloria:
I would certainly review and merge a PR fixing this!

kkirsche:
Thanks for sharing this, I'm still seeing this error with the work that's been done to add support — will keep looking into it to identify what the cause is

Issue #85, opened 2016-07-19
Support for MethodView


lafrech:
I'm using Flask's [`MethodView`s](http://flask.pocoo.org/docs/0.11/views/#method-based-dispatching) for an API.

Basically, the idea is to have one class per endpoint and one class method for each HTTP method. The method name is the HTTP method, lowercase.

``` python
from flask.views import MethodView

class UserAPI(MethodView):

    def get(self):
        """User detail view.
        ---
        get:
            responses:
                200:
                    schema: UserSchema
        """
        users = User.query.all()
        ...

    def post(self):
        """User post.
        ---
        post:
            responses:
                201:
                    schema: UserSchema
        """
        user = User.from_form_data(request.form)
        ...

app.add_url_rule('/users/', view_func=UserAPI.as_view('users'))
```

AFAIU, there is currently no support for Flask's `MethodView` or Flask's `View`, only functions are supported. Is this correct?

I think I managed to get apispec to produce the correct OpenAPI file using the following patch:

https://github.com/marshmallow-code/apispec/compare/dev...Nobatek:dev_method_views

The idea is that if the `view` is a `MethodView`, we should not use the `view`'s docstring but loop over the docstrings of each method.

Thinking of it, maybe the only difference is that this allows the user to keep each docstring in each method rather than grouping all of them at the top of the class. This can be seen as cosmetic, but I think it would be better this way.

I'd be happy to get any feedback.

If there is interest for this feature, I may clean this up, add tests and all, but before I move any further, I'd rather be sure I'm on the right tracks.

lafrech:
[flask-apispec](http://flask-apispec.readthedocs.io/en/latest/usage.html) uses apispec and provides support for `Blueprint` (which @deckar01 would like to add to apispec: https://github.com/marshmallow-code/apispec/issues/68), `MethodView`, along with nice decorators that allows to specify a `Schema` to dump the data with, automatically registering the `Schema` in the spec, and the same automatic registration of input arguments `Schema`.

I'm at a crossroad, right now. flask-apispec sounds appealing, but I'm still a bit reluctant. Not only because it says "flask-apispec isn't stable yet", but also because from what I've read, its code is far from trivial ("annotation" mechanism), and if I ever want to change or customize anything, I'm afraid I might end up losing more time than I gained using it, if I'm able to do it at all.

[...]

PR #134, opened 2017-07-11
Add support for Flask MethodView to apispec.ext.flask

djanderson:
Addresses issues #85 and #125.

There are only two slight differences when using a MethodView function:

1. the method ("get", "post", etc) is inferred and is not included in the docstring
2. the path is added with `spec.add_path(method_view=view)`

Here's the patch in action: 

```python
from flask import Flask
from flask.views import MethodView

app = Flask(__name__)

class GistApi(MethodView):
    def get(self):
       """Gist view
       ---
       responses:
           200:
               schema:
                   $ref: '#/definitions/Gist'
       """

app.test_request_context().push()
method_view = GistApi.as_view('gists')
app.add_url_rule("/gists", view_func=method_view)
spec.add_path(method_view=method_view)
print(spec.to_dict()['paths'])
# {'/gists', {'get': {'responses': {200: {'schema': {'$ref': '#/definitions/Gist'}}}}}}

```

Please review!

### DIFF
--- a/apispec/ext/flask.py
+++ b/apispec/ext/flask.py
@@ -12,6 +12,7 @@ Passing a view function::
     def gist_detail(gist_id):
         '''Gist detail view.
         ---
+        x-extension: metadata
         get:
             responses:
                 200:
@@ -23,7 +24,8 @@ Passing a view function::
     app.test_request_context().push()
     spec.add_path(view=gist_detail)
     print(spec.to_dict()['paths'])
-    # {'/gists/{gist_id}': {'get': {'responses': {200: {'schema': {'$ref': '#/definitions/Gist'}}}}}}
+    # {'/gists/{gist_id}': {'get': {'responses': {200: {'schema': {'$ref': '#/definitions/Gist'}}}},
+    #                  'x-extension': 'metadata'}}
 
 Passing a method view function::
 
@@ -33,6 +35,10 @@ Passing a method view function::
     app = Flask(__name__)
 
     class GistApi(MethodView):
+        '''Gist API.
+        ---
+        x-extension: metadata
+        '''
         def get(self):
            '''Gist view
            ---
@@ -41,13 +47,20 @@ Passing a method view function::
                    schema:
                        $ref: '#/definitions/Gist'
            '''
+           pass
+
+        def post(self):
+           pass
 
     app.test_request_context().push()
     method_view = GistApi.as_view('gists')
     app.add_url_rule("/gists", view_func=method_view)
     spec.add_path(method_view=method_view)
     print(spec.to_dict()['paths'])
-    # {'/gists', {'get': {'responses': {200: {'schema': {'$ref': '#/definitions/Gist'}}}}}}
+    # {'/gists': {'get': {'responses': {200: {'schema': {'$ref': '#/definitions/Gist'}}}},
+    #             'post': {},
+    #             'x-extension': 'metadata'}}
+
 
 """
 from __future__ import absolute_import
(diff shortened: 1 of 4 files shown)


yoichi:
@djanderson your code uses parameter name "method_view" != "view" in path_helper_function, then it cannot be used with apispec.ext.marshmallow (cause TypeError). 
You can solve it by switching behavior like follows, instead of registering 2nd path_helper_function.

```
if hasattr(view, 'view_class'):
    if issubclass(view_func.view_class, MethodView):
        ....
```

I think MethodView support is useful. Thanks!

yoichi:
One more change is required  to use it with apispec.ext.marshmallow. Currently operations constructed in the preceding extension is not passed to the subsequent extensions.

djanderson:
Hi @yoichi, thanks for the feedback.

I hate to ask, but could you give me an example of something that throws a TypeError? Like maybe propose a test case I can add that shows the error?

I'm currently using this with `apispec.ext.marshmallow` and it seems to be working. In my case, I'm registering the marshmallow schemas with `spec.definition` instead of `spec.add_path(urlspec...`. If you're already registering your paths with marshmallow UrlSpec path helper, what benefit does adding it again with the MethodView give you?

yoichi:
I'm sorry for lack of explanation.
If we pass

```
 plugins=[
     'apispec.ext.flask',
     'apispec.ext.marshmallow'
]
```

to APISpec constructor, TypeError is occured and ignored in add_path(method_view=method_view) (not thrown to outside) when calling on apispec.ext.marshmallow.schema_path_helper and the method never executed since argument 'view' is not given.

As I noted, there was another problem (https://github.com/marshmallow-code/apispec/pull/138). While I fixing it, I found tornado extension uses 'urlspec' (!= 'view') argument and have the same problem as above, and I had to make 'view' argument of apispec.ext.marshmallow.schema_path_helper optional for backward compatibility.
Therefore, there is no problem in the behavior with your patch.

On the other hand, I think it is better that the implementation will not rely on the fact one of followings cause TypeError internally and not called:

* apispec.ext.flask.path_from_view (cause TypeError when 'view' is not given)
* apispec.ext.flask.path_from_method_view (cause TypeError when 'method_view' is not given)

I prefer an explicit conditional branch in path helper function instead.

Sincelery,

djanderson:
I see, thanks for the clarification.

I don't disagree with you, but I probably won't address it in this PR since I don't think I'm abusing the TypeError in a way that's not currently intended.

I appreciate the review, though!

apispec/ext/flask.py
```diff
@@ -72,6 +111,20 @@ def path_from_view(spec, view, **kwargs):
...
+        docstring_yaml = utils.load_yaml_from_docstring(view.__doc__)
+        operations[method_name] = docstring_yaml or dict()
+
+    path.operations.update(operations)
+    return path
+
 def setup(spec):
     """Setup for the plugin."""
     spec.register_path_helper(path_from_view)
+    spec.register_path_helper(path_from_method_view)
```
yoichi, line 130: Is it necessary to register multiple path helper functions in one extension? I think this breaks the locality of the conditional branch whether the view is derived from MethodView or not.

apispec/ext/flask.py
```diff
@@ -72,6 +111,20 @@ def path_from_view(spec, view, **kwargs):
     path = Path(path=path, operations=operations)
     return path
 
+def path_from_method_view(spec, method_view, **kwargs):
```
yoichi, line 114: Using 'method_view' (not 'view' as used in existing helper function) forces user to distinguish the view is derived from MethodView or  not.

yoichi (commented):
I've also put comments in the code to clarify what I mean.
But I will respect your choice if you decide that you will not deal with it.

Thanks,

tangram:
I would love to see this merged and released. Any chance you could find the time to resolve the last few things, @djanderson, @yoichi?

djanderson:
@tangram, thanks for the encouragement. I think this implementation was faithful to the "ask forgiveness, not permission" style of the existing code, but @yoichi has raised some fair criticisms and his modifications in #139 look to provide the desired capability as well as an improved user experience. I'm closing this PR so that we can focus on landing #139. @yoichi, thanks again for the critical eye and improvements.

PR #138, opened 2017-07-14
Pass operations to downstream extension

yoichi:
As I commented in https://github.com/marshmallow-code/apispec/pull/134 ,
even if preceding extension construct 'operations', it won't be used in ext.marshmallow to resolve Schema class.
I've added a test to illustrate the problem (https://github.com/yoichi/apispec/commit/833724a4d029d15b7db207817b895f28279a07a2), then fix it (https://github.com/yoichi/apispec/commit/fdd97dd9b97677dca86e3db5de307fec02348e7b).

Regards,


sloria:
Looks correct. Thanks!

PR #140, opened 2017-08-04
Custom swag field decorator

### LINKED ISSUE #120: Support custom Marshmallow fields
Currently, all custom fields automatically become `type="string", format=None` due to the [hard-coded `FIELD_MAPPING`](https://github.com/marshmallow-code/apispec/blob/bc6e0e5df8d23fd7dadedd09339f1e76910c2498/apispec/ext/marshmallow/swagger.py#L24-L42). Thus, I cannot extend any marshmallow type, e.g. `Dict` or `Number` with apispec support. Currently, the only way to extend apispec support is to patch the `FIELD_MAPPING` dict.

dradetsky:
For issue #120. Note that I do not say "Closes #120" b/c this is not exactly my finest work (in a bit of a hurry), and many other solutions are possible. Also, like I said, it only addresses marshmallow, not the other things. Does it work for them? Dunno.

Also, fun fact:

```
_______________________________________________________ summary _______________________________________________________
  py27: commands succeeded
ERROR:   py34: InterpreterNotFound: python3.4
ERROR:   py35: InterpreterNotFound: python3.5
  py36: commands succeeded
```

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -34,3 +34,4 @@ Contributors (chronological)
 - Florian Scheffler `@nebularazer <https://github.com/nebularazer>`_
 - Vadim Radovel `@NightBlues <https://github.com/NightBlues>`_
 - Yoichi NAKAYAMA `@yoichi <https://github.com/yoichi>`_
+- Daniel Radetsky `@dradetsky <https://github.com/dradetsky>`_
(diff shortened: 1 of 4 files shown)


dradetsky:
@sloria b/c of conflict I rebased. Not totally sure I did it right, so I did a separate pr #141. If that looks ok, I can push same content into this branch.

dradetsky:
I just haven't used rebase in like 2 years, and I didn't want to fuck up your nice commit history, especially since it's easy to just make more branches.

PR #141, opened 2017-08-04
Custom swag field decorator rebased

dradetsky:


dradetsky:
> dradetsky  wants to merge 5 commits into marshmallow-code:dev from dradetsky:custom-swag-field-decorator-rebased

not rly, tho. this is just a test.

apispec/ext/marshmallow/swagger.py
```diff
@@ -62,7 +88,10 @@ def _observed_name(field, name):
 
 
 def _get_json_type_for_field(field):
-    json_type, fmt = FIELD_MAPPING.get(type(field), ('string', None))
+    if hasattr(type(field), CUSTOM_FIELD_MAPPING_ATTR):
```
sloria: Minor: If you remove `type` here, this function will work for both type classes and type instances.

apispec/ext/marshmallow/swagger.py
```diff
@@ -41,6 +41,32 @@
     marshmallow.fields.List: ('array', None),
 }
 
+CUSTOM_FIELD_MAPPING_ATTR = '__custom_swagger_field_mapping'
```
sloria: Minor: How about just `__swagger_field_mapping`?

apispec/ext/marshmallow/swagger.py
```diff
@@ -41,6 +41,32 @@
...
+
+    - a core marshmallow field type (in which case we reuse that type's mapping)
+    - a pair of string|None
+    """
+    if len(args) == 1 and args[0] in FIELD_MAPPING:
+        swagger_type_field = FIELD_MAPPING[args[0]]
+    elif len(args) == 2:
+        swagger_type_field = args
+    else:
+        raise ArgumentError('pass core marshmallow field type or (type, fmt) pair')
```
sloria: I think a `TypeError` is more appropriate here. Also, uppercase the first letter of the message.

tests/test_ext_marshmallow.py
```diff
@@ -33,6 +35,42 @@ def test_can_use_schema_as_as_definition(self, spec, schema):
...
+
+        assert props_0['name']['type'] == 'string'
+        assert 'format' not in props_0['name']
+
+        assert props_a['name']['type'] == 'string'
+        assert props_a['name']['format'] == 'date-time'
+
+        assert props_b['name']['type'] == 'integer'
+        assert props_b['name']['format'] == 'int32'
+
```
sloria, line 78: Can you also add a test for the error condition, i.e. a `TypeError` (as suggested in my comment) is raised if an invalid argument is passed to `map_to_swagger_field`

sloria (changes requested):
Thanks for the PR! Just a few small changes suggested. Feel free to rebase/squash if you want, but I don't require it.

apispec/ext/marshmallow/swagger.py
dradetsky: primarily b/c one is not supposed to actually read/use the attribute by hand, so there's no percentage in making the name shorter. `__swagger_field_mapping` would definitely be better for an attribute that was part of the api. I'll change if you like tho.

apispec/ext/marshmallow/swagger.py
dradetsky: if you say so

apispec/ext/marshmallow/swagger.py
dradetsky: good point

tests/test_ext_marshmallow.py
dradetsky: depends. what's in it for me?

dradetsky:
@sloria take 2nd look?

tests/test_ext_marshmallow.py
sloria: I was concerned because you originally raised an `ArgumentError` and I didn't think `ArgumentError` was a built-in. In any case, I'd like to have that tested in order to define the expected behavior before this is merged.

apispec/ext/marshmallow/swagger.py
sloria: Even though this is a private attribute, `custom_` is redundant naming.

tests/test_ext_marshmallow.py
dradetsky: @sloria j/k; i made the change

dradetsky:
@sloria 3 tries in the circus

apispec/ext/marshmallow/swagger.py
```diff
@@ -41,6 +41,32 @@
     marshmallow.fields.List: ('array', None),
 }
 
+CUSTOM_FIELD_MAPPING_ATTR = '__swagger_field_mapping'
+
+
+def map_to_swagger_field(*args):
```
sloria: `map_to_swagger_type` might be a more appropriate name.

sloria (changes requested):
Once the name of the decorator is changed and this PR is updated with the latest `dev` branch (rebase preferred but not required), this should be good to merge.

apispec/ext/marshmallow/swagger.py
dradetsky: I chose "field" b/c the swagger obj it's mapped to has 2 keys, type & format. Still, I don't suppose it matters, so I'll change it rather than prolonging the discussion.

sloria: Fair point. Is `format` actually required? If I recall the spec correctly, it was optional.

dradetsky: It may be technically optional, but consider that the type for a datetime is `{type: string, format: date-time}` or something like that. I assume that without the format, you're gonna have a bad time.

dradetsky:
@sloria given that you seem to be reconsidering _field -> _type, i'm going to wait on actually doing the rebase until we make a call one way or the other, b/c i have a feeling the revert will get more annoying after a rebase.

sloria:
I still think that `_type` is preferred over `_field`, if only because that is the language used in the spec: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#data-types . "Field" may be conflated with a marshmallow field.

dradetsky:
something went funny with the second rebase that i don't totally understand; i can fixup things if it's too screwy

sloria:
Uh oh, looks like unrelated commits made it into this branch. Could you fix it up please?

dradetsky:
@sloria i'm not sure how to fixup current branch and/or if it's possible, but i think #143 is ok (i'm learning how to use rebase rather than merge as i go)

PR #142, opened 2017-08-08
Fix swagger location mapping for default_in param in property2parameter

decaz:
Also "cookies" -> "cookie" swagger location mapping was added.


sloria:
Thanks @decaz ! Would you mind adding a regression test for this change? That would greatly help me to understand this fix.

decaz:
@sloria done!

sloria:
Thanks! I'll review and release this fix when I'm off of work today.

PR #145, opened 2017-08-15
Plugin to support Flask-RESTful

theirix:
This PR provides a plugin to support Flask-RESTful documentation.
I am aware of existing plugin apispec_restful which unfortunately doesn't work with Blueprints and possibly with classic Flask apps. This plugin overcomes these limitations.

Plugin is tested by unittests and by usage in a real application.
I added a documentation with typical use-cases and several docstrings.

Commit message:
```
Added path-helper which accepts Flask-RESTful resources.
Resources can be linked to Path via Api instance or endpoint.
Plugin works with Flask Blueprints and Flask classic apps.
```

### DIFF
--- a/apispec/ext/flask_restful.py
+++ b/apispec/ext/flask_restful.py
@@ -1,0 +1,120 @@
+# -*- coding: utf-8 -*-
+"""Flask-RESTful plugin. Includes a path helper that allows you to pass a resource
+object to `add_path`.
+::
+
+    from pprint import pprint
+
+    from flask_restful import Api, Resource
+    from flask import Flask
+
+    class HelloResource(Resource):
+        def get(self, hello_id):
+            '''A greeting endpoint.
+                   ---
+                   description: get a greeting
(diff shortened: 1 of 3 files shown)


sloria:
Flask-RESTful seems to be in a deprecated state (source: https://github.com/flask-restful/flask-restful/issues/335#issuecomment-314096887) and I'm not willing to take on further maintenance burden. Therefore, I'm not going to merge this for now.

Thanks anyway for the PR. I recommend either sending a PR to `apispec_restful` or publishing a separate package.

theirix:
Thanks for the review and for addressing concerns about Flask-RESTful. Maybe it's better to use pure Flask solution.

I'll try to make a separate package from this PR because apispec_restful is not at pypi yet.

theirix:
Just for information, I extracted the proposed plugin to the autonomous [package](https://github.com/theirix/apispec-flask-restful) and uploaded it to pypi.

Issue #135, opened 2017-07-11
apispec.utils.load_yaml_from_docstring should let YAMLError propagate


djanderson:
* I think the "---" prefix is enough to keep from accidentally parsing random docstrings as YAML
* the YAML parser raises very useful errors
* swallowing the YAML parser error silently is *probably* not doing the user a favor

```python
>>> docstring_with_no_yaml = """Some function docstring.
... 
... Nothing to see here.
... """
>>> docstring_with_yaml = """Some function docstring.
... ---
... get:
...   responses:
...     200:
...       schema: $ref: '#/definitions/Schema'
... """
>>> from apispec.utils import *
>>> load_yaml_from_docstring(docstring_with_no_yaml)
>>> # ok
... load_yaml_from_docstring(docstring_with_yaml)
>>> # huh?
... import yaml
>>> yaml.load(dedent('\n'.join(docstring_with_yaml.split('\n')[1:])))
Traceback (most recent call last):
[...]
yaml.scanner.ScannerError: mapping values are not allowed here
  in "<unicode string>", line 5, column 19:
          schema: $ref: '#/definitions/Schema'
                      ^ 
```

I propose changing

```python
def load_yaml_from_docstring(docstring):
    """Loads YAML from docstring."""
    split_lines = trim_docstring(docstring).split('\n')

    # Cut YAML from rest of docstring
    for index, line in enumerate(split_lines):
        line = line.strip()
        if line.startswith('---'):
            cut_from = index
            break
    else:
        return None

    yaml_string = "\n".join(split_lines[cut_from:])
    yaml_string = dedent(yaml_string)
    try:
        return yaml.load(yaml_string)
    except yaml.YAMLError:
        return None
```
to
```python
def load_yaml_from_docstring(docstring):
    """Loads YAML from docstring."""
    split_lines = trim_docstring(docstring).split('\n')

    # Cut YAML from rest of docstring
    for index, line in enumerate(split_lines):
        line = line.strip()
        if line.startswith('---'):
            cut_from = index
            break
    else:
        return None

    yaml_string = "\n".join(split_lines[cut_from:])
    yaml_string = dedent(yaml_string)
    return yaml.load(yaml_string)
```

sloria:
I think you are correct @djanderson . I would certainly review and merge a PR for this!

PR #146, opened 2017-08-16
Don't catch/pass YAMLError when parsing docstring fails.

djanderson:
The current behavior is to swallow any parse errors silently, which makes
trying to add more than a trivial apispec docstring confusing and frustrating.

See issue #135 for more discussion.


sloria:
Thanks! This is now released.

PR #147, opened 2017-08-20
Handle load_only marshmallow schema

theirix:
A small fix that handles `load_only` schema parameter as well as `dump_only`. It allows to correctly handle schemas with `Meta.load_only` field.

### DIFF
--- a/apispec/ext/marshmallow/swagger.py
+++ b/apispec/ext/marshmallow/swagger.py
@@ -533,7 +533,8 @@ def fields2jsonschema(fields, schema=None, spec=None, use_refs=True, dump=True, 
     exclude = set(getattr(Meta, 'exclude', []))
 
     for field_name, field_obj in iteritems(fields):
-        if field_name in exclude or (field_obj.dump_only and not dump):
+        if field_name in exclude or (field_obj.dump_only and not dump) or \
+                                    (field_obj.load_only and dump):
             continue
 
         observed_field_name = _observed_name(field_obj, field_name)


theirix:
Sorry, missed a test. It's an interesting question whether the dump_only field should appear when `dump=False` specified and load_only field for `dump=True`. Previously they appeared with a `read_only=True` swagger attribute and I doubt this is an expected behaviour.

sloria:
@theirix I'm not sure about the failing test either. Intuitively, it does seem like `load_only` fields should be excluded if `dump=True`. Perhaps if the `dump` parameter isn't explicitly passed, both `load_only` and `dump_only` fields should appear. If `dump=True` is passed, `load_only` fields should not appear.

theirix:
Yes, tri-state logic that you have described could help if we change signature for patched functions and a bunch of others to:
```python
def schema2jsonschema(schema, spec=None, use_refs=True, dump=None, name=None):
```
And it is not a fully backward-compatible change. A patch can broke a library for those who uses `load_only` fields and expects these fields to be in result anyway. If apispec is used as a request/response serializer when dump parameter is explicitly specified or implied as a default parameter, a patch can be useful as it properly excludes load-only fields.

I don't insist in accepting this PR in a minor release. It's totally up to you to decide is it worth to change a behaviour in a such way.

sloria:
Closing as this has gone stale, and I'm not sure this is an issue anymore. Feel free to re-open if it's still a valid issue.

lafrech:
This is still an issue. It is discussed in https://github.com/marshmallow-code/apispec/issues/119.

There are three possible cases:
- Load only: parameter schema
- Dump only: response schema
- Load and dump: definition

A boolean `dump` parameter is not enough. We need either two parameters or a tri-state parameter.

sloria:
OK, so it sounds like we'll need a different solution from this PR anyway.

lafrech:
Exactly.

PR #150, opened 2017-08-30
Recover context in example

yoichi:
In flask example code, test_request_context is pushed but not poped. It may affect the subsequent operation.
I think it is better to use "with" statement which recover the context automatically.
http://flask.pocoo.org/docs/0.12/reqcontext/

Thanks,


sloria:
Thanks!

PR #151, opened 2017-09-02
Skip non-dict elements on resolve to avoid AttributeError

yoichi:
Fix against "x-extension: value" cause following error:

```
>           for response in operation.get('responses', {}).values():
E           AttributeError: 'str' object has no attribute 'get'

apispec/ext/marshmallow/__init__.py:131: AttributeError
```


sloria:
Thanks!

Issue #136, opened 2017-07-11
Order of plugins affects output


CausticYarn:
The example shows:

```
# Create an APISpec
spec = APISpec(
    title='Swagger Petstore',
    version='1.0.0',
    plugins=[
        'apispec.ext.flask',
        'apispec.ext.marshmallow',
    ],
)
```
 
And the output is:

```
#         "responses": {
#           "200": {
#             "schema": {
#               "$ref": "#/definitions/Pet"
#             },
#             "description": "A pet to be returned"
#           }
```

However, if the order of the plugins to spec is (marshmallow, flask), then the output is:

```
#         "responses": {
#           "200": {
#             "schema": "PetSchema"
#             "description": "A pet to be returned"
#           }
```

Which doesn't resolve correctly in the Swagger doc.  If the order is relevant by design, it should be documented.

sloria:
This looks to be a bug; order should not matter, at least not for the flask and marshmallow plugins. I would definitely review and merge a PR fixing this.

yoichi:
Order dependency comes from the fact that operation resolve process is done inside path helpers.
It can be solved by separating them https://github.com/marshmallow-code/apispec/pull/152

We have to introduce incompatible change to extension mechanism to solve it. I'm not sure we should fix this issue or not. What do you think? @sloria

sloria:
This is fixed in 0.27.0

PR #152, opened 2017-09-02
Extract operation resolvers and execute them after path helpers

yoichi:
Solve https://github.com/marshmallow-code/apispec/issues/136
NOTE: This fix introduce incompatible change to extension mechanism.

Thanks,


sloria:
@yoichi What about this PR is backwards-incompatible? Isn't this just exposing a new hook, or am I missing something?

yoichi:
I thought extracting a part of functionality of path_helper introduce some incompatibility. But that was  my misunderstanding.
Extensions can keep resolver functionality in path_helper, or change themselves to use new hook.

docs/writing_plugins.rst
```diff
@@ -12,6 +12,7 @@ There are three types of helper functions:
```
sloria, line 12: Above line will need to be changed to "four".

sloria:
@yoichi OK, thanks for the clarification. If this doesn't introduce backwards incompatibilities, then I think we should go ahead and merge this.

docs/writing_plugins.rst
```diff
@@ -8,10 +8,11 @@ Helper Functions
 
 Plugins are comprised of "helper" functions that augment the behavior of `apispec.APISpec` methods.
 
-There are three types of helper functions:
+There are four types of helper functions:
 
 * Definition helpers
 * Path helpers
+* Operation resolvers
```
sloria: @yoichi Would it make sense to call these "Operation helpers" for consistency?

sloria (commented):
Just a suggestion for a name change. Other than that, this looks good to merge.

docs/writing_plugins.rst
yoichi: Yes. They don't have to resolve something and call them helpers like existing ones seems better. I'll push a change.
Thanks for the suggestion

sloria:
Looks good to me. Thanks!

Issue #75, opened 2016-05-18
schema2parameters - body + header


shaicantor:
Hey,

I have a schema and the default_in = body, but one of the fields has a headers location.
schema2parameters inserts the header field inside the body instead of returning one body param and one header param.

What do I do?

``` python
class AccountsSchema(Schema):
    user_id = fields.Int(required=True, location="headers")
    name = fields.Str(required=True, validate=validate.Length(min=1, max=255), location="json")
    status = fields.Str(required=True, validate=validate.OneOf(choices=["Active", "Paused"]), location="json")
```

``` python
def fields2parameters(fields, schema=None, spec=None, use_refs=True, dump=True,
                      default_in='body', name='body', required=False):
    """Return an array of OpenAPI parameters given a mapping between field names and
    :class:`Field <marshmallow.Field>` objects. If `default_in` is "body", then return an array
    of a single parameter; else return an array of a parameter for each included field in
    the :class:`Schema <marshmallow.Schema>`.

    https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameterObject
    """
    if default_in == 'body':
        if schema is not None:
            # Prevent circular import
            from apispec.ext.marshmallow import resolve_schema_dict
            prop = resolve_schema_dict(spec, schema, dump=dump)
        else:
            prop = fields2jsonschema(fields, spec=spec, use_refs=use_refs, dump=dump)

        return [{
            'in': default_in,
            'required': required,
            'name': name,
            'schema': prop,
        }]

    assert not getattr(schema, 'many', False), \
        "Schemas with many=True are only supported for 'json' location (aka 'in: body')"

    exclude_fields = getattr(getattr(schema, 'Meta', None), 'exclude', [])

    return [
        field2parameter(
            field_obj,
            name=_observed_name(field_obj, field_name),
            spec=spec,
            use_refs=use_refs,
            dump=dump,
            default_in=default_in
        )
            for field_name, field_obj in iteritems(fields)
            if (
                (field_name not in exclude_fields)
                and not (field_obj.dump_only and not dump)
            )
    ]
```

PR #153, opened 2017-09-10
Handle multiple fields with location='json' correctly.

yoichi:
Fixes https://github.com/marshmallow-code/apispec/issues/75
* Merge results from field2parameter.
* Modify input data for test_field_required since required
  fields in body are treated differently.


sloria:
Thanks @yoichi . Would you mind looking into the build failure?

sloria:
Thank again! This is released in 0.25.3. 

By the way, I've given you write access to this project. No pressure to do any more work than you already have; just wanted to open the door to further collaboration. =)

PR #158, opened 2017-10-08
Add aiohttp plugin

0lmer:


sloria:
This is great! Thanks @0lmer . I would be more comfortable with this as a 3rd-party plugin for now. I am trying to keep this repo small and easy to maintain.

If you do end up releasing this as a separate library, please add to the Ecosystem page here: https://github.com/marshmallow-code/apispec/wiki/Ecosystem

PR #160, opened 2017-10-09
Enum from equal validator

Bangertm:
Swagger generated from fields with a Marshmallow fields.Equal validator now contain an enum with a single entry.


sloria:
Thanks!

PR #162, opened 2017-10-19
Add "options" method to allowed methods

buxx:
Hello,

Why "OPTIONS" method is not listed in allowed methods ? I use apispec for application who need to describe it's OPTIONS roads.

Is it possible to add it officially ?

Regards,
Bux.

### DIFF
--- a/apispec/core.py
+++ b/apispec/core.py
@@ -13,6 +13,7 @@ VALID_METHODS = [
     'patch',
     'delete',
     'head',
+    'options',
 ]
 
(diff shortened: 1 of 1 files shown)


yoichi:
OPTIONS is listed in
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#pathItemObject
and this change looks good to me.

BTW, Travis CI error is caused by another change and will be fixed by  https://github.com/marshmallow-code/apispec/pull/163

sloria:
Thanks!

PR #167, opened 2017-11-08
Permit auto generation of definition schema names

lebouquetin:
The case we faced is the following one:

- we have a self referencing schema, lets call it NodeSchema.
- this schema is included in another schema - NodeList

When we try to define NodeList in APIspec, the NodeSchema is not yet reference, so its structure is inlined. As it is self referencing, the inline results in an infinite recursion, so it crashes

The global solution is to add definitions of our Schema in the right order. As we use APIspec in [hapic]( https://github.com/algoo/hapic/) (a generic auto-documented REST API framework), we can't assume that schema will be defined in the right order for APIspec.

There are two ways to get schemas defined in the right order:
1. first solution is to preprocess all the schema we have to document in order to add thems to apispec in the right order. It's not very elegant (but it works)
2. second solution is to define the rule to define name of a schema in the auto-generated documentation. This is quite easy to integrate in APIspec - this is the current PR.

How it works:
- if you do nothing special, APIspec keep on working as it already do.
- if you define a callable `schema_name_resolver_callable` at construction time of `APISpec`, then it will be used for auto-referencing required schemas.

We'd be happy this PR to be accepted fast, let us know if you are ok and/or if we should change some stuff for the PR to be accepted (naming, etc).

Damien

### DIFF
--- a/apispec/core.py
+++ b/apispec/core.py
@@ -91,11 +91,11 @@ class APISpec(object):
         See https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#infoObject
     :param \*\*dict options: Optional top-level keys
         See https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#swagger-object
-    :param callable schema_name_resolver_callable: callable who generate the
+    :param callable schema_name_resolver: callable who generate the
         schema definition name. The schema will be given to this callable and
         definition name must be returned by callable, ex:
 
-            def schema_name_resolver_callable(schema):
+            def schema_name_resolver(schema):
                 return schema.__name__
     """
 
@@ -105,7 +105,7 @@ class APISpec(object):
         version,
         plugins=(),
         info=None,
-        schema_name_resolver_callable=None,
+        schema_name_resolver=None,
         **options
     ):
         self.info = {
@@ -114,7 +114,7 @@ class APISpec(object):
         }
         self.info.update(info or {})
         self.options = options
-        self.schema_name_resolver_callable = schema_name_resolver_callable
+        self.schema_name_resolver = schema_name_resolver
         # Metadata
         self._definitions = {}
         self._parameters = {}
--- a/apispec/core.py
+++ b/apispec/core.py
@@ -91,7 +91,12 @@ class APISpec(object):
         See https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#infoObject
     :param \*\*dict options: Optional top-level keys
         See https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#swagger-object
-    :param callable schema_name_resolver_callable: callable who generate the schema definition name
+    :param callable schema_name_resolver_callable: callable who generate the
+        schema definition name. The schema will be given to this callable and
+        definition name must be returned by callable, ex:
+
+            def schema_name_resolver_callable(schema):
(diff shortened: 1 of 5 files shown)


apispec/ext/marshmallow/__init__.py
```diff
@@ -162,6 +162,17 @@ def resolve_schema_dict(spec, schema, dump=True, use_instances=False):
         schema_cls = schema
     else:
         schema_cls = resolve_schema_cls(schema)
+
+    # Auto reference schema if available
+    if schema_cls not in plug.get('refs', {}):
+        if spec.schema_name_resolver_callable:
```
yoichi: spec can be None here (travis-ci test fails)

apispec/core.py
```diff
@@ -91,15 +91,25 @@ class APISpec(object):
         See https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#infoObject
     :param \*\*dict options: Optional top-level keys
         See https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#swagger-object
+    :param callable schema_name_resolver_callable: callable who generate the schema definition name
```
yoichi: What is the input and output of this callable? Could you add a test code to illustrate its usage?

yoichi (changes requested):

apispec/ext/marshmallow/__init__.py
lebouquetin: @buxx could you upgrade our code for fix ?

apispec/core.py
lebouquetin: @yoichi good question and query. @buxx you upgrade the documentation and add a test code to illustrate?

buxx: @lebouquetin I do that.

apispec/ext/marshmallow/__init__.py
buxx: @lebouquetin done

apispec/core.py
buxx: done

yoichi (approved):

yoichi:
Thanks for clarifying the spec. I think this feature is useful.

* It would be nice if we have a test code for a new feature in order to avoid future regression. 
* "schema_name_resolver_callable" is a bit long. Does it make sense if simply "schema_name_resolver"?
* How about skipping spec.definition() call if "not definition_name"?

buxx:
@yoichi For your differents points: no problem, i  will do these changes next week :-)

buxx:
@yoichi It's seems we have a problem with python2.7: when json.dumps (used in tests) we got an `RuntimeError: dictionary changed size during iteration`. I guess it's because we add definitions when reading them (i don't know why it is working in python 3.x).
I don't found a smart solution. Maybe try to execute the code who parse definitions before ? Or disable this feature for python < 3.x ?

buxx:
@yoichi I found another way to implement this feature: auto-reference schemas when add a definition. Solve the problem of dynamic evolution of definition dict. Tests are green and all nested schemas are auto referenced correctly.

buxx:
Hello @yoichi @sloria we have no news since last month about this pull request. any problem with it ?

sloria:
@buxx Thanks for your patience and for getting this up to date. I'll try to take a look at this this weekend.

sloria:
Thanks for the PR!

PR #173, opened 2017-12-01
Filter view methods with rule methods

ukaratay:
This commit makes sure that same documentation is not generated twice when a view used in different rules with different methods.


sloria:
Thanks @ukaratay . Would you mind adding a test for this?

ukaratay:
@sloria Sure, it may take a little while but I will.

sloria:
@ukaratay No worries. Thanks a lot!

ukaratay:
@sloria Done.

sloria:
Thanks!

Issue #165, opened 2017-11-02
Support OpenAPI version 3.0


rdehouss:
Hi there,

We're using APISpec with Marshmallow and aiohttp or sanic and it works really great.

There are some new great changes with OpenAPI (version 3) vs Swagger API (version 2).
Any plan to implement Open API version 3?

Thanks

Kind regards,

Raphaël

zikphil:
+1 for this

benturnerdev:
+1

lafrech:
Guys, please use the thumbs up emoticon rather than posting +1 comments. This is just noise for those being notified by email.

If you want to help tackle this issue, you may dig in the OpenAPI 2 and 3 specs to spot differences, or maybe find a "what's new ?" or "migration howto" like this:

https://www.blazemeter.com/blog/whats-new-in-swagger-open-api3

Here's what I noticed, already.

#### Polymorphism

The use of `discriminator` has been clarified:

https://github.com/OAI/OpenAPI-Specification/blob/3.0.0/versions/3.0.0.md#composition-and-inheritance-polymorphism

(see [this commit](https://github.com/OAI/OpenAPI-Specification/issues/403).)

I don't think this is a breaking change. AFAIK, there was nothing about this in apispec and anyway, v3 is backwards compatible with v2 IIUC.

#### Parameter locations

https://github.com/OAI/OpenAPI-Specification/blob/3.0.0/versions/3.0.0.md#parameter-locations

> Possible values are "query", "header", "path" or "cookie".

It differs slightly from OpenAPi v2:

https://github.com/OAI/OpenAPI-Specification/blob/3.0.0/versions/2.0.md#fixed-fields-7

> Possible values are "query", "header", "path", "formData" or "body".

This may have an impact on `__location_map__` in [marshmallow swagger extension](https://github.com/marshmallow-code/apispec/blob/38e3cd481c3d4d903fc69ff3d048e798bb9284ae/apispec/ext/marshmallow/swagger.py#L581), which looks like a mix of the two, already:

```python
__location_map__ = {
    'query': 'query',
    'querystring': 'query',
    'json': 'body',
    'headers': 'header',
    'cookies': 'cookie',
    'form': 'formData',
    'files': 'formData',
}
```

It seems, though, that it is not just about updating that dictionary. From [the link above](https://www.blazemeter.com/blog/whats-new-in-swagger-open-api3):

[...]

PR #174, opened 2017-12-05
adds optional config param for OpenAPI version 3

Bangertm:
This is an intial attempt to add basic support for OpenApi 3.
Based on the value of the openapi_version field, the APISpec.todict
places schemas either directly in "definitions" (v2) or in
"components/schemas" (v3)

References for nested Marshmallow schemas are also handled

Begins work on #165

See [https://blog.readme.io/an-example-filled-guide-to-swagger-3-2/](https://blog.readme.io/an-example-filled-guide-to-swagger-3-2/)

### DIFF
--- a/apispec/core.py
+++ b/apispec/core.py
@@ -2,6 +2,7 @@
 """Core apispec classes and functions."""
 import re
 from collections import OrderedDict
+from distutils import version
 
 from apispec.compat import iterkeys
 from .exceptions import APISpecError, PluginError
@@ -15,8 +16,6 @@ VALID_METHODS = [
     'head',
     'options',
 ]
-
-OPENAPI_VERSIONS = ('2.0', '3.0.0')
 
 
 def clean_operations(operations, openapi_version):
@@ -131,11 +130,7 @@ class APISpec(object):
         }
         self.info.update(info or {})
 
-        if openapi_version not in OPENAPI_VERSIONS:
-            raise APISpecError(
-                'Not a valid OpenAPI version number: {}'.format(openapi_version)
-            )
-        self.openapi_version = openapi_version
+        self.openapi_version = validate_openapi_version(openapi_version)
 
         self.options = options
         self.schema_name_resolver = schema_name_resolver
@@ -370,3 +365,21 @@ class APISpec(object):
         if method not in self._response_helpers:
             self._response_helpers[method] = {}
         self._response_helpers[method].setdefault(status_code, []).append(func)
+
+def validate_openapi_version(openapi_version_str):
+    """Function to validate the version of the OpenAPI passed into APISpec
+
+    :param str openapi_version_str: the string version of the desired OpenAPI
+        version used in the output
+    """
+    min_inclusive_version = version.LooseVersion('2.0')
+    max_exclusive_version = version.LooseVersion('4.0')
+
+    openapi_version = version.LooseVersion(openapi_version_str)
+
+    if not min_inclusive_version <= openapi_version < max_exclusive_version:
+        raise APISpecError(
+            'Not a valid OpenAPI version number: {}'.format(openapi_version)
+        )
+
+    return openapi_version.vstring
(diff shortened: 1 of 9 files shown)


Bangertm:
These changes plus updating the docstrings on our view functions to support OpenAPI 3 was able to generate valid OpenAPI 3 output.  

I used this [this tool](https://mermade.org.uk/openapi-converter) to convert the Swagger 2 output from APISpec to OpenAPI 3 format.  Then updated the docstrings and made changes to APISpec to generate similar output that I verified in [Swagger Editor](https://editor.swagger.io/).

lafrech:
Great. Thanks for tackling this.

Shouldn't it be

```python
       elif self.openapi_version == '3.0.0':
           ret['openapi'] = self.openapi_version
           ret['components'] = {'schemas': self._definitions, 'parameters': self._parameters}
```

?

Bangertm:
Ultimately yes, parameters should be put in the 'components' dictionary.  

I was going to do that, but noticed there is some functionality [here](https://github.com/marshmallow-code/apispec/blob/cb974e44a9781481f0b40749c98864beb2829fe8/apispec/core.py#L32) that generates refs to parameters when the parameters is specified as just a name in the docstring.  See this [commit](https://github.com/marshmallow-code/apispec/commit/3556849f815d42e28f84ee60cc518af00c988cc1).  

It looks like to generate valid OpenAPI 3 refs we would have to pass in the version to Path.__init__.  Didn't want to make that change until I understood a little better how things were working, which I think I do now.  

Also I should note, the assumption that is being made with this commit is that the only definitions being added to the spec are schemas.  That's apparently not strictly true in OpenAPI 2, but it is the way I use APISpec.

apispec/core.py
```diff
@@ -16,20 +16,30 @@
     'options',
 ]
 
-OPENAPI_VERSION = SWAGGER_VERSION = '2.0'
+OPENAPI_VERSIONS = ('2.0', '3.0.0')
```
sloria: It would be nice if users could specify just the `major` or `major.minor` version , e.g. `2`, `3`, `3.0`.

Maybe you could use `distutils.version.LooseVersion` to parse the version.

sloria:
Is there a way we can validate against the 3.0 spec in a test? Can `utils.validate_swagger` be modified to support OpenAPI 3?

sloria:
This is looking great! Just a few suggestions above. Also, please get this up to date with the `dev` branch.

Once those are done, I think this good to merge.

Bangertm:
Using `distutils.version.LooseVersion` to validate the version input by the user.  The check I put in is to make sure the version is greater than or equal to 2.0 and less than 4.0.0.  That leaves it up the user input the version that they want in the output dictionary.  

It looks like the swagger validation tool we were previously using does not support OpenAPI 3.  I found a list of validation tools [here](https://github.com/Mermade/awesome-openapi3#parsersmodelsvalidators), but haven't had a chance to take a look at any of them.

Bangertm:
Added a test for an OpenAPI 3 spec.  To do this I swapped out the `swagger-tools` npm package for `check_api`, which supports both OpenAPI 2 and 3 (and uses swagger-tools as a dependency).

Also added some additional functionality to resolve Marshmallow Schemas from their new locations in the `requestBody` and `response` dictionaries.  Right now the resolver is not raising an exception if it finds a schema in the wrong place based on the version, the output will just be invalid.

The one other place that schemas can be in OpenAPI 3 are in parameters query, path, etc. (just not body).  But these schema dictionaries need to be formatted differently in 3 than they were in 2.  See query and path parameters on [this blog](https://dev.to/mikeralphson/comparing-openapiswagger-20-and-300-rc1).  At this point I haven't worked out a good way to make this change.  So I'd like to defer that.

sloria:
This looks great. Thanks for your hard work and patience!

sloria:
This is released in 0.29.0.

I've also given you write access to the repo. No pressure to do any more work than you already have; just want to open the door to further collaboration. =)

sloria:
@Bangertm Since you're a frequent contributor to apispec, I'd like to invite you to the marshmallow-code contributor Slack. If you're interested, send me an email at sloria1 AT gmail DOT com.

PR #175, opened 2017-12-05
Fix an error running utils.validate_swagger on Windows

Bangertm:
The result was a WindowsError.

The arguments for subprocess.check_output are hard coded so running
in shell is safe.


sloria:
Thanks!

PR #180, opened 2018-01-17
Feature/auto ref autogenerated class

buxx:


buxx:
Miss click, sorry for that !

PR #182, opened 2018-01-19
OneOf Polymorphism

Bangertm:
One of the nice features in OpenAPI 3 is support for [oneOf polymorphism](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#discriminator-object).  In one of our API's we are using [marshmallow-oneofschema](https://github.com/maximkulkin/marshmallow-oneofschema) to enable serialization/deserialization/validation of multiple schemas on an endpoint or in a nested field.  This merge request enables API's using marshmallow-onofofschema to be described in OpenAPI 3 format.  

A couple notes;

- I'm [testing](https://github.com/marshmallow-code/apispec/compare/dev...Bangertm:polymorphic?expand=1#diff-36818c01919e762abcc4b32dfdd1b61fR484) to see if the schema matches the marshmallow-onofschema API as a determination of if the schema is a oneOf schema.  This seemed cleaner than testing for inheritance. 
- I think the most user friendly behavior when apispec encounters a marshmallow-oneofschea is to add a oneOf schema object to the spec _and also_ add each of the referenced schemas to the spec.  This way the user doesn't have to add them separately.  The potential concern I encountered is that by default nested fields are not resolved right away and `json.dumps` serialization of the spec results in a `RuntimeError: dictionary changed size during iteration`.  (Oddly this doesn't happen with print)  I was able to work around this by using the new `schema_name_resolver` functionality, but I'm not 100% sure this is the right thing to do.
- I added most of the new code and tests to new modules because swagger.py was getting pretty long.  
- [Marshmallow-Polyfield](https://github.com/Bachmann1234/marshmallow-polyfield) provides similar functionality, but would require additional work to represent in a spec.

### DIFF
--- a/apispec/ext/marshmallow/polymorphic.py
+++ b/apispec/ext/marshmallow/polymorphic.py
@@ -1,0 +1,35 @@
+# -*- coding: utf-8 -*-
+
+def is_oneof(schema):
+    """Function to determine if a schema matches the oneOf api
+
+    :param Schema schema: A marshmallow Schema instance or a class object
+    :rtype: bool, True if the schema matches the oneOf API False otherwise
+    """
+    return hasattr(schema, 'type_field') and hasattr(schema, 'type_schemas')
+
+
+def oneof_schema_2_schema_object(spec, oneof_schema):
+    """Function to generate an OpenAPI schema object based on a oneOf Schema
+    This function will add all of the Schemas to the spec and then return a
+    dictionary containing references to those Schemas
+
+    :param APISpec spec: Current `APISpec` instance
+    :param OneOfSchema oneof_schema: A OneOfSchema instance or a class object
+    """
+    mapping = {}
+    oneof = []
+    for name, schema in oneof_schema.type_schemas.items():
+        spec.definition(name, schema=schema)
+        ref = '#/components/schemas/{}'.format(name)
+        mapping.update({name: ref})
+        oneof.append({'$ref': ref})
+
+    ret = {
+        'oneOf': oneof,
+        'discriminator': {
+            'propertyName': oneof_schema.type_field,
+            'mapping': mapping
+        }
+    }
+    return ret
--- a/apispec/ext/marshmallow/polymorphic.py
+++ b/apispec/ext/marshmallow/polymorphic.py
@@ -20,6 +20,7 @@ def oneof_schema_2_schema_object(spec, oneof_schema):
     mapping = {}
     oneof = []
     for name, schema in oneof_schema.type_schemas.items():
+        name = name.replace(' ', '_')
         spec.definition(name, schema=schema)
         ref = '#/components/schemas/{}'.format(name)
         mapping.update({name: ref})
--- a/apispec/ext/marshmallow/polymorphic.py
+++ b/apispec/ext/marshmallow/polymorphic.py
@@ -20,9 +20,9 @@ def oneof_schema_2_schema_object(spec, oneof_schema):
     mapping = {}
     oneof = []
     for name, schema in oneof_schema.type_schemas.items():
-        name = name.replace(' ', '_')
-        spec.definition(name, schema=schema)
-        ref = '#/components/schemas/{}'.format(name)
+        component_name = name.replace(' ', '_')
+        spec.definition(component_name, schema=schema)
+        ref = '#/components/schemas/{}'.format(component_name)
         mapping.update({name: ref})
         oneof.append({'$ref': ref})
 
--- a/apispec/ext/marshmallow/swagger.py
+++ b/apispec/ext/marshmallow/swagger.py
@@ -17,6 +17,7 @@ from marshmallow.compat import text_type, binary_type, iteritems
 from marshmallow.orderedset import OrderedSet
 
 from apispec.lazy_dict import LazyDict
+from . import polymorphic
 
 ##### marshmallow #####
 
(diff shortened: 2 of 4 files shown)


lafrech:
Hi.

Thanks for this. I'd really like such a feature added to apispec.

I've been doing this on a project of mine. I'm also using marshmallow-oneofschema.

From what I remember, OpenAPI v2 also allows polymorphism, but it is "not well defined" (https://github.com/OAI/OpenAPI-Specification/issues/403). OpenAPI v3 makes it better: you don't have to use the class name (which should be hidden implementation detail) as the discriminator. Overall, I think it makes sense to support only OpenAPI v3 polymorphisme, like you did.

> I'm testing to see if the schema matches the marshmallow-onofschema API as a determination of if the schema is a oneOf schema. This seemed cleaner than testing for inheritance.

Why do you think the inheritance test wouldn't be clean? Do you just want to avoid importing marshmallow-oneofschema? Checking `type_field` and `type_schema` restricts the test to marshmallow-oneofschema anyway. Unless you want to allow another class to propose this interface?

I think this PR adds an optional dependency on marshmallow-oneofschema. It would be nice if a polymorphic field was added to Marshmallow, but last time I asked (can't find the source anymore) @sloria didn't want to integrate that into marshmallow and recommended using marshmallow-oneofschema.

If we go that way, we could add marshmallow-oneofschema as an optional dependency explicitly, try/except its import, only use the feature if import succeeds and test inheritance using isinstance. Just thoughts. I'm totally open to discussion about this.

> I think the most user friendly behavior when apispec encounters a marshmallow-oneofschea is to add a oneOf schema object to the spec and also add each of the referenced schemas to the spec. This way the user doesn't have to add them separately. 

Yes, this is what I wanted to achieve but I didn't go that far yet. My client code registers all schemas in a loop.

>  The potential concern I encountered is that by default nested fields are not resolved right away and json.dumps serialization of the spec results in a RuntimeError: dictionary changed size during iteration. (Oddly this doesn't happen with print) I was able to work around this by using the new schema_name_resolver functionality, but I'm not 100% sure this is the right thing to do.

This should be investigated. Unfortunately, I have no time to look into this right now.

> I added most of the new code and tests to new modules because swagger.py was getting pretty long.

Good.

This is definitely a feature worth having. It is a recurrent request.

We'd rather get it right the first time, though.

Bangertm:
Regarding the test, if there was a canonical way of supporting polymorphism with Marshmallow this would be easier...  I don't feel strongly about it, testing for the API still seems simpler than adding an optional dependency, try/excepting the import, optioning the feature based on the import result, and testing for inheritance. 

If one were to rewrite marshmallow-onofschema in light of OpenAPI 3, the attributes would probably be `property_name` and `mapping`. In which case, sure, multiple classes could have that interface.

The issue with adding nested marshmallow-oneofschema's to a spec comes down to [this line](https://github.com/marshmallow-code/apispec/blob/9b364208180f3f1f9757981bfbfd19557e872ca6/apispec/ext/marshmallow/swagger.py#L557) where an unevaluated function is added to a `LazyDict`.  By default the function is not evaluated until it is referenced in the `LazyDict`.  Using the `schema_name_resolver` works around the issue because it adds a hook to immediately evaluate the function instead of waiting for something like `json.dumps` to resolve the function as part of a loop over the entire `LazyDict`.

lafrech:
> Regarding the test, if there was a canonical way of supporting polymorphism with Marshmallow this would be easier...

Yes, totally. I wish a polymorphic schema feature was added to Marshmallow. AFAIK, @sloria prefers to keep oneofschema as an external library. Maybe it is not deemed canonical enough to be part of the core. But this makes it more complicated for external libs like apispec (well, I use the plural "libs", but apispec is the only case I know of).

timakro:
Until this is ready you can use my plugin [apispec-oneofschema](https://github.com/timakro/apispec-oneofschema) for this.

lafrech:
Thanks @timakro.

I just noticed @timakro's plugins are GPL, not MIT. No license war intended but I thought it would be worth mentioning here because it is not that usual in this ecosystem so anyone could use them inadvertently and because this means it is still worth having in apispec (or as an MIT plugin).

timakro:
@lafrech I moved to the less restrictive LGPL.

sloria:
I'm still lukewarm about adding first-class support for a oneofschema. It feels like more complexity than is worth adding to both apispec and marshmallow.

This isn't to say we'll never support it, but I don't think it's a priority right now.

Closing this, since it's gone stale. Thanks anyway @Bangertm for the proposed implementation and thanks @timakro for releasing the plugin.

deckar01:
What if this package provided `OneOf` as a custom field? It should be possible to implement apispec's "discriminator" strategy as a subclass of `Nested`. The discriminator map isn't technically required by OpenAPI, but it would be needed so that the field could figure out which schema to use.

```py
from apispec.ext.marshmallow import OneOf

class PetSchema(Schema):
    animal = OneOf('species', {'turtle': TurtleSchema, 'dog': DogSchema, 'cat': CatSchema})
    name = fields.Str()
```

**Edit:** Updated the example to specify the discriminator field name.

lafrech:
@deckar01 do you mean create a polymorphic field _à la_ oneofschema in apispec?

The original author of oneofschema is not interested in it anymore (https://github.com/maximkulkin/marshmallow-oneofschema/pull/17#issuecomment-428647052) but offered to transfer ownership so it would be nice to either have some polymorphic field implementation either in marshmallow or in apispec (marshmallow would be a better place IMHO), or if openofschema is the external lib we recommend, at least keep an eye on the status of the project.

deckar01:
I was suggesting a `OneOf` field rather than a `OneOf` schema. It wouldn't support `OneOf` for the response level schema, but it could be used for nested fields and non-homogeneous lists. It would also be much easier to implement.

lafrech:
Oh, sorry. I didn't notice the field vs. schema distinction.

Serialization would also require a class -> schema mapping.

I don't see right now how this is easier than oneof schema, but I wouldn't mind the lack of response level polymorphism and I see why you're suggesting to put it in apispec rather than marshmallow.

(Yet, a polymorphic schema in marshmallow is a recurrent request and would be great.)

Issue #184, opened 2018-03-05
Body parameters renamed "body"


mathewmarcus:
When adding a body parameter, i.e. a parameter where `in="body"`, via a docstring, the parameter is automatically renamed to `"body"`. 

For example:

```python
from apispec import APISpec
from flask import Flask
from marshmallow import Schema, fields
import json

spec = APISpec(
    title='Swagger Petstore',
    version='1.0.0',
    plugins=[
        'apispec.ext.flask',
        'apispec.ext.marshmallow',
    ],
)

app = Flask(__name__)


class CategorySchema(Schema):
    id = fields.Int()
    name = fields.Str(required=True)


class PetSchema(Schema):
    category = fields.Nested(CategorySchema, many=True)
    name = fields.Str()


@app.route('/random', methods=['POST'])
def random_pet():
    """A cute furry animal endpoint.
    ---
    post:
        description: Get a random pet
        parameters:
            - name: pet
              required: true
              in: body
              schema: PetSchema
    """
    pass


with app.test_request_context():
    spec.add_path(view=random_pet)

print(json.dumps((spec.to_dict())))
```

this yields the following swagger 

```
{
    "info": {
        "title": "Swagger Petstore",
        "version": "1.0.0"
    },
    "paths": {
        "/random": {
            "post": {
                "description": "Get a random pet",
                "parameters": [
                    {
                        "in": "body",
                        "required": false,
                        "name": "body",
                        "schema": {
                            "type": "object",
                            "properties": {
                                "category": {
                                    "type": "array",
                                    "items": {
                                        "type": "object",
                                        "properties": {
                                            "name": {
                                                "type": "string"
                                            },
                                            "id": {
                                                "type": "integer",
                                                "format": "int32"
                                            }
                                        },
                                        "required": [
                                            "name"
                                        ]
                                    }
                                },
                                "name": {
                                    "type": "string"
                                }
                            }
                        }
                    }
                ]
            }
        }
    },
    "tags": [],
    "swagger": "2.0",
    "definitions": {},
    "parameters": {}
}
```

Essentially, even though the parameter is explicity named `pet`, it is automatically renamed `body`. 

This appears to stem from the fact that when calling [schema2parameters](https://github.com/marshmallow-code/apispec/blob/dev/apispec/ext/marshmallow/swagger.py#L347), 
[...]

PR #185, opened 2018-03-06
Fix marshmallow body param

### LINKED ISSUE #168: Marshmallow-schema'd body parameters lose 'required' and 'description' field spec
Consider this code:

```python
from apispec import APISpec

spec = APISpec(
        title='bug',
        version="0.0.1",
        plugins=(
            'apispec.ext.marshmallow',
        )
    )

class SomeBodyParam(Schema):
    name = fields.String(required=True)
    city = fields.String()
    age = fields.Integer()

spec.add_path(
    'foo/',
    {
        'post': {
            'parameters': [
                { 'name': 'some body param',
                  'in': 'body',
                  'schema': SomeBodyParam(),
                  'description': 'Description disapears and so does required',
                  'required': True
                  }
            ]
        }
    }
)

spec.to_dict()
```

The output is:

```
{'definitions': {},
 'info': {'title': 'bug', 'version': '0.0.1'},
 'parameters': {},
 'paths': OrderedDict([('foo/',
               {'post': {'parameters': [{'in': 'body',
                   'name': 'body',
                   'required': False,
                   'schema': {'properties': {'name': {'type': 'string'}, 'city': {'type': 'string'}, 'age': {'type': 'integer', 'format': 'int32'}},
                    'required': ['name'],
                    'type': 'object'}}]}})]),
 'swagger': '2.0',
 'tags': []}
```

Bug:
- The `required` field switched from True to False
- The `description` field got lost

The `name` parameter also gets renamed, but I believe this is per-spec.

I believe the cause is here: https://github.com/marshmallow-code/apispec/blob/0.27.0/apispec/ext/marshmallow/__init__.py#L148

Adding a `, required=parameter['required'] if 'required' in parameter else False` ought to fix the required bit.  Do something similar for `description`, but need to add it to [swagger.py:fields2parameters()](https://github.com/marshmallow-code/apispec/blob/0.27.0/apispec/ext/marshmallow/swagger.py#L367) signature and include it in the dict constructor in there.

That sound like a reasonable approach?

mathewmarcus:
I believe this fixes #168 #184. I tried to introduce the fewest possible changes to function signatures, but I am definitely open to any suggestions.

### DIFF
--- a/apispec/ext/marshmallow/__init__.py
+++ b/apispec/ext/marshmallow/__init__.py
@@ -203,8 +203,9 @@ def resolve_parameters(spec, parameters):
         if not isinstance(parameter.get('schema', {}), dict):
             schema_cls = resolve_schema_cls(parameter['schema'])
             if issubclass(schema_cls, marshmallow.Schema) and 'in' in parameter:
+                del parameter['schema']
                 resolved += swagger.schema2parameters(
-                    schema_cls, default_in=parameter['in'], spec=spec)
+                    schema_cls, default_in=parameter.pop('in'), spec=spec, **parameter)
                 continue
         resolved.append(parameter)
     return resolved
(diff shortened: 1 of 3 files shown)


sloria:
This looks like the right behavior. Thanks for the PR!

PR #186, opened 2018-03-15
Identify local object definitions and recurse over the properties

lphuberdeau:
This adds a little bit of consistency allowing object definitions as part of the response - not only arrays. My primary case for this is to wrap all of our responses in a data property without having to declare a custom schema for each data type individually.

### DIFF
--- a/apispec/ext/marshmallow/__init__.py
+++ b/apispec/ext/marshmallow/__init__.py
@@ -238,8 +238,11 @@ def resolve_schema_in_response(spec, response):
 
 def resolve_schema_dict(spec, schema, dump=True, use_instances=False):
     if isinstance(schema, dict):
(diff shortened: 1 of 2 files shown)


lafrech:
IIUC, you're stepping into the object to identify properties using known schemas. If we go this route, why stop at the first level and not recurse through the whole schemas?

Also, it looks like your use case is similar to http://marshmallow.readthedocs.io/en/latest/extending.html#example-enveloping. I'm not sure how to use this pre_load/post_dump trick with apispec, though

lphuberdeau:
The patch does recurse through the entire structure of object definitions as the same function is called for each matching child entry.

lafrech:
Oh, right. Sorry.

I don't have any objection to this PR.

sloria:
Good change. Thanks @lphuberdeau!

PR #188, opened 2018-03-21
Specify custom bottle app

dtaniwaki:
I think it'd be great if I can specify custom app instead of default app for Bottle.

```py
from apispec import APISpec
spec = APISpec(title='', version='', plugins=['apispec.ext.bottle'])

from bottle import Bottle
app = Bottle()
@app.route('/')
def index():
    pass

spec.add_path(view=index, app=app)
```

Could you consider to merge this PR?


sloria:
Makes sense to me. Thanks!

PR #189, opened 2018-03-29
Maintain order in which methods are added to endpoint

lafrech:
Similar to https://github.com/marshmallow-code/apispec/pull/86 and https://github.com/marshmallow-code/apispec/pull/87.

### DIFF
--- a/apispec/core.py
+++ b/apispec/core.py
@@ -70,7 +70,7 @@ class Path(object):
 
     def __init__(self, path=None, operations=None, openapi_version='2.0'):
         self.path = path
-        operations = operations or {}
+        operations = operations or OrderedDict()
         openapi_version = validate_openapi_version(openapi_version)
         clean_operations(operations, openapi_version.version[0])
         invalid = {key for key in
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -1,4 +1,6 @@
 # -*- coding: utf-8 -*-
+from collections import OrderedDict
+
 import pytest
 import mock
 import yaml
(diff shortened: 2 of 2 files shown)


lafrech:
Rebasing this on https://github.com/marshmallow-code/apispec/pull/190 assuming it would be merged as is. If not, I'll rebase again.

This change is useful to me because I'm enforcing the order of the methods for each endpoint in the spec:

    ['OPTIONS', 'HEAD', 'GET', 'POST', 'PUT', 'PATCH', 'DELETE', 'TRACE']

I find it nice to have a consistent method order.

It works because I'm iterating on the methods in my lib, not in apispec.

----------------------------------------------------------------------------------------

For people using `apispec.ext.flask.path_from_view`, it would take another modification in there:

```python
HTTP_METHODS = ['OPTIONS', 'HEAD', 'GET', 'POST', 'PUT', 'PATCH', 'DELETE', 'TRACE']

def path_from_view(spec, view, **kwargs):
    [...]
    operations = OrderedDict()
    for method in HTTP_METHODS:
        if method in view.methods and method in rule.methods:
            [...]
```

Likewise for tornado and the others.

This means hardcoding the method order. The user could mutate the global method list to control the order, though.

I think this can be done in another PR without holding this one.

sloria:
Thanks @lafrech . #190 is merged, but it looks like this needs to be rebased one more time. Also, would you mind opening an issue for the Flask/Tornado issue you mentioned?

lafrech:
Great. I just rebased and opened #193.

PR #190, opened 2018-03-29
Don't let Path inherit from dict

lafrech:
While investigating https://github.com/marshmallow-code/apispec/pull/189, I've been wondering why `Path` inherits from `dict` and in the meantime holds content in its `operations` attribute.

It looks like most content is duplicated in `operations` and in `Path` itself. In fact, inheriting from `dict` seems kinda strange.

This PR removes this inheritance to simplify the code. It makes more sense to me this way. It brings no enhancement, but tries to improve maintainability.

What was the reason for the original design? Any objection to this simplification?

I had to modify 2 tests to get this going. This could mean I'm totally mistaken, but from the name of the tests, I don't think the behaviour I modified is what was being tested, rather a side-effect.

Feedback welcome.

------------------------------------------------------

Note that if this PR is applied, we'll probably need to change

    self.operations = operations or {}

into

    self.operations = operations or OrderedDict

to get the behaviour proposed in https://github.com/marshmallow-code/apispec/pull/189.

### DIFF
--- a/apispec/core.py
+++ b/apispec/core.py
@@ -54,7 +54,7 @@ def clean_operations(operations, openapi_major_version):
                                        for p in parameters]
 
 
-class Path(dict):
+class Path(object):
     """Represents an OpenAPI Path object.
 
     https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#pathsObject
@@ -68,7 +68,7 @@ class Path(dict):
         standard.
     """
 
-    def __init__(self, path=None, operations=None, openapi_version='2.0', **kwargs):
+    def __init__(self, path=None, operations=None, openapi_version='2.0'):
         self.path = path
         operations = operations or {}
         openapi_version = validate_openapi_version(openapi_version)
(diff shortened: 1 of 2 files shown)


sloria:
Hm, I don't recall why I made `Path` a dict... I wonder if that decision predated the `to_dict` method and we relied on `dict(path)` for serializing Paths.

This PR is probably the right way to go.

lafrech:
Great.

How shall we proceed? Should I merge this?

Considering the modifications in the tests, do you consider this a breaking change? I would say it isn't and those tests were wrongish, but maybe I'm focusing on my own experience of apispec.

sloria:
Yeah, I think this can be merged as is. It looks like a breaking change, but it shouldn't affect most users.

@lafrech Is this still WIP? If not, can you rewrite the commit message without "WIP"?

lafrech:
Alright, I rebased, reviewed and modified the commit message.

I checked those tests again and I don't think anyone should be affected.

Thanks.

PR #192, opened 2018-04-04
Update .gitignore

lafrech:


.gitignore
```diff
@@ -92,3 +92,7 @@ ENV/
 
 # Rope project settings
 .ropeproject
+
+# Node stuff:
```
sloria: There shouldn't be Node.js-related files in this repo. Why is this necessary?

lafrech: Those files are generated when running the spec validation tests that use Node JS.

I'm not a JS guy. I tired to follow https://github.com/marshmallow-code/apispec/blob/dev/CONTRIBUTING.rst. Maybe I did it wrong. Since there is no npm package on Debian, I had to use Node repos and install check_api as root. I have little idea how the whole thing works.

Bottom line, once I run the tests, I end up with a `node_modules` directory and a `package-lock.json` file. I figured I'd add that to .gitignore, but if this happened due to me not proceeding correctly, we might as well reword the documentation. Or perhaps I should follow some Node JS for dummies tutorial.

sloria: Ah, I totally forgot that the spec validation was done with a node module. OK, then this is correct. 👍

PR #194, opened 2018-04-04
Noting support for OpenAPI v3

earth2marsh:
… and some cleanup to other OpenAPI references.

(note: because this is a docs change, I didn't bother opening an issue—hope that's ok!)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -18,7 +18,7 @@ A pluggable API specification generator. Currently supports the `OpenAPI Specifi
 Features
 ========
 
-- Supports OpenAPI Specification vesions 2 and 3 (f.k.a. the Swagger specification)
+- Supports OpenAPI Specification version 2 (f.k.a. the Swagger specification) with limited support for version 3.
 - Framework-agnostic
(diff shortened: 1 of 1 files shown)


lafrech:
(No pb, you don't have to open an issue when you open a PR.)

OpenAPI v3 support is being worked on, but I'm not sure it should be advertised as "supported" already. @Bangertm, what do you think?

BTW, the work on this can be followed in https://github.com/marshmallow-code/apispec/issues/165.

Bangertm:
It's advertised as: 

> A pluggable API specification generator. Currently supports the OpenAPI (versions 2.0 and 3.0) spec.

On marshmallow-code.  The 3.0 support is fairly rudimentary though.  It's basically what you get after running a 2.0 spec through a 2 to 3 converter tool.  No support for any of the new features of 3.0 (such as polymorphism :wink:).

sloria:
Maybe the docs should read "Currently supports OpenAPI Specification version 2, with limited support for version 3" ?

earth2marsh:
Thanks for clarifying. I've taken a pass to encorporate the suggested changes, HTH!

When more complete v3 support lands, would love to add you to the OpenAPI Implementations list.

sloria:
Thanks!

PR #197, opened 2018-04-06
Add appropriate nullable flag depending on OpenAPI version

Bangertm:
Addresses issue identified by @lafrech in #165

### DIFF
--- a/apispec/ext/marshmallow/swagger.py
+++ b/apispec/ext/marshmallow/swagger.py
@@ -216,6 +216,23 @@ def field2length(field):
     return attributes
 
 
+def field2nullable(field, spec):
+    """Returns the dictionary with an appropriate nullable key depending on
+    the OpenAPI version.  Returns an empty dict if the field is not nullable
+
+    :param Field field: A marshmallow field.
+    :param spec: apispec.core.APISpec instance
+    :rtype: dict
+    """
+    if not field.allow_none:
+        return {}
+
+    if spec and spec.openapi_version.version[0] == 3:
+        return {'nullable': True}
+
+    return {'x-nullable': True}
+
+
 # Properties that may be defined in a field's metadata that will be added to the output
 # of field2property
 # https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schemaObject
@@ -293,8 +310,7 @@ def field2property(field, spec=None, use_refs=True, dump=True, name=None):
     if field.dump_only:
         ret['readOnly'] = True
 
-    if field.allow_none:
-        ret['x-nullable'] = True
+    ret.update(field2nullable(field, spec))
 
     ret.update(field2range(field))
     ret.update(field2length(field))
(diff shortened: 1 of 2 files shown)


apispec/ext/marshmallow/swagger.py
```diff
@@ -216,6 +216,23 @@ def field2length(field):
...
+    the OpenAPI version.  Returns an empty dict if the field is not nullable
+
+    :param Field field: A marshmallow field.
+    :param spec: apispec.core.APISpec instance
+    :rtype: dict
+    """
+    if not field.allow_none:
+        return {}
+
+    if spec and spec.openapi_version.version[0] == 3:
```
lafrech, line 230: `spec.openapi_version.version[0]`

This is accessed in two places, and more to come. Maybe we should add a shortcut for it.
For instance properties `spec.major_version`, `spec.minor_version`.

Nothing blocking, just a thought.

apispec/ext/marshmallow/swagger.py
```diff
@@ -293,8 +310,7 @@ def field2property(field, spec=None, use_refs=True, dump=True, name=None):
     if field.dump_only:
         ret['readOnly'] = True
 
-    if field.allow_none:
-        ret['x-nullable'] = True
+    ret.update(field2nullable(field, spec))
```
lafrech, line 313: I'm a bit bothered by the fact that the `spec` attribute to the function is optional and defaults to `None`.

Same concern for `get_ref_path`, BTW. Even worse then because we don't check it is not `None`.

Maybe we should make `spec` mandatory. It breaks a lot of tests, but maybe not real-life use cases (otherwise, you'd get an `AttributeError` in the call to `get_ref_path`).

We should give this a thought and maybe refactor a bit. Check what `spec` is used for and decide what needs to be passed. From what I gathered, we only use OpenAPI version, the list of schema definitions (`refs`) and an optional `schema_name_resolver` callable). Maybe we don't need to carry the whole spec object all around.

Anyway, this issue is much wider than this specific PR, so I suppose we can merge this and open a separate issue for the `spec` attribute refactor.

Bangertm: Making spec mandatory everywhere is probably the right first step.  Practically the entry points to all of these functions (`schema_definition_helper` `schema_path_helper` and `schema_operation_resolver`) have spec as mandatory so it's only the tests that break.  

I think there is definitely some opportunity for refactoring.  We have quite a bit of complexity around multiple ways of handling nested schemas (using refs, inlining, waiting to see if a ref gets added later, immediately resolving a name).  Simplifying that would probably help in other areas.

lafrech: Yes, totally.

I'm not sure the whole spec needs to be carried about, but I think the version should be everywhere, as it may affect anything, so it should be provided to all helpers.

Anyway, we can merge this PR and focus on the refactor later on.

lafrech:
I just proposed a PR that does this and more: https://github.com/marshmallow-code/apispec/pull/199.

sloria:
`nullable` is supported in #199. Thanks anyway for the PR!

PR #199, opened 2018-04-09
field2property rework

lafrech:
This PR reworks field2property a bit to provide more consistency to the field2attribute methods. This makes it easier to manage different OpenAPI versions.

It also adds the `x-nullable`/`nullable` and `writeOnly` attributes depending on the version.

It does not address the optional spec attribute issue discussed in https://github.com/marshmallow-code/apispec/pull/197 but it is a first step.

@sloria, @Bangertm Feedback welcome.

### DIFF
--- a/apispec/ext/marshmallow/swagger.py
+++ b/apispec/ext/marshmallow/swagger.py
@@ -110,7 +110,7 @@ def _get_json_type_for_field(field):
     return json_type, fmt
 
 
-def field2choices(field):
+def field2choices(field, **kwargs):
     """Return the dictionary of swagger field attributes for valid choices definition
 
     :param Field field: A marshmallow field.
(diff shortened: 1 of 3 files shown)


sloria (approved):
I think this approach is a big improvement. +1 to moving OpenAPI 3 work forward, too. Nice!

apispec/ext/marshmallow/swagger.py
```diff
@@ -110,29 +110,69 @@ def _get_json_type_for_field(field):
...
+    :param Field field: A marshmallow field.
+    :rtype: dict
+    """
+    attributes = {}
+    if field.dump_only:
+        attributes['readOnly'] = True
+    return attributes
+
+
+def field2write_only(field, **kwargs):
```
lafrech, line 150: If this was Python 3 only, I'd write

```python
def field2write_only(field, *, openapi_major_version, **kwargs):
```

lafrech:
Great. Shall I merge this, then?

sloria:
LGTM 🚢 🇮🇹

Issue #201, opened 2018-05-04
Why doesn't Marshmallow Dict field resolve to additionalProperties?


UrKr:
Any Dict field in a schema seems to resolve to just 'type': 'object' in the spec output. Why not to additionalProperties? 

```

class ASchema(MM.Schema):
    data = MM.fields.Dict(keys=MM.fields.String(), values=MM.fields.Nested(OtherSchema), required=True)
```

the yaml output for:
`
spec.definition('A', ASchema, schema=ASchema)
`
is this: 
`data: {type: object}`
Why not this:
```
data: {
  type: object,
  additionalProperties: {
    $ref: '#/definitions/A'
  }
}
```

lafrech:
`keys` and `values` were added in marshmallow 3 and are not yet documented by apispec. Indeed, it would be nice to have them documented.

Not sure how that would look like, though.

(Thinking out loud.) In marshmallow, keys can be anything, including stuff that would generate wrong JSON. (I mean can you use arrays as dict indexes in JSON? I guess not.) It is fine if we ignore pathological cases. I mean if the doc feature does not work on cases that can't be functional anyway, no big deal.

UrKr:
Yes, I suppose it would only have to work with string keys. Though the `keys` value could actually just be ignored, no? 

Are you taking pull requests for this? I don't have one ready or anything, I would have to look at the code and see what I can do. Is there any other ambiguity apart from the keys? Naively, I would think it's just a matter of adding additional `Dict` field handling.

lafrech:
Yes, I suppose we can ignore `keys`.

We welcome PRs. When the change involved is a bit tricky, it can be better to discuss it here before actually working on it, to avoid the frustration of seeing the PR not merged.

From this [SO post](https://stackoverflow.com/questions/41239913/why-additionalproperties-is-the-way-to-represent-dictionary-map-in-swagger-ope), it seems your fix is correct, so it's simpler than I thought. You can send a PR implementing this.

Thanks!

lafrech:
Fixed in #202.

PR #202, opened 2018-05-04
Add handling for marshmallow Dict values metadata

UrKr:
For #201 
Previously, even if a Dict field had 'values' metadata, it resolved to:
```
{ type: object }
```
Now the field is checked for the presence of 'values' metadata and an `additionalProperties` property is added to the definition. The vlaue of `additionalProperties` is obtained by a call to field2property

### DIFF
--- a/apispec/ext/marshmallow/swagger.py
+++ b/apispec/ext/marshmallow/swagger.py
@@ -380,7 +380,8 @@ def field2property(field, spec=None, use_refs=True, dump=True, name=None):
     elif isinstance(field, marshmallow.fields.List):
         ret['items'] = field2property(field.container, spec=spec, use_refs=use_refs, dump=dump)
     elif isinstance(field, marshmallow.fields.Dict):
-        ret['additionalProperties'] = field2property(field.metadata['values'], spec=spec, use_refs=use_refs, dump=dump)
+        if 'values' in field.metadata:
+            ret['additionalProperties'] = field2property(field.metadata['values'], spec=spec, use_refs=use_refs, dump=dump)
         
     # Dasherize metadata that starts with x_
     metadata = {
--- a/apispec/ext/marshmallow/swagger.py
+++ b/apispec/ext/marshmallow/swagger.py
@@ -380,8 +380,9 @@ def field2property(field, spec=None, use_refs=True, dump=True, name=None):
     elif isinstance(field, marshmallow.fields.List):
         ret['items'] = field2property(field.container, spec=spec, use_refs=use_refs, dump=dump)
     elif isinstance(field, marshmallow.fields.Dict):
-        if MARSHMALLOW_VERSION_INFO[0] >= 3:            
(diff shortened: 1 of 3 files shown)


lafrech:
Thanks. Could you please add a test?

Also, why not check `field.value_container` like with the `List` field above?

UrKr:
Oh, right. Sorry, I'll add a test. 
As for the `field.value_container` instead of `field.metadata['values']`. It doesn't seem that Dict has a field `.value_container`. I get `'Dict' object has no attribute 'value_container'`

Which is strange, because looking at the marshmallow code https://github.com/marshmallow-code/marshmallow/blob/dev/marshmallow/fields.py#L1106 it seems it would have to be there

UrKr:
I've added the test, though I'm not entirely sure I did it in line with the others.

lafrech:
OK, I get it.

Apparently, you're using marshmallow 2. And the Travis tests run on marshmallow 2 as well.

The `values` and `keys` parameters to `Dict` were introduced in marshmallow 3 (still beta). With marshmallow 2, `values` parameter just is added to metadata and ignored.

What you need to do is test the marshmallow version. There is an example in the code already. Something like this:

    if MARSHMALLOW_VERSION_INFO[0] >= 3

Then, only if using marshmallow 3, check `value_container` and add the `additionalProperties` (it makes no sense with MA2).

And you can skip the test if running marshmallow < 3. Search for a pattern similar to this in `test_swagger.py`:

    @pytest.mark.skipif(swagger.MARSHMALLOW_VERSION_INFO[0] < 3,

I just pushed a commit to let the tests run on MA3 as well, so please rebase you branch so that MA3 behaviour gets tested.

Thanks.

UrKr:
I don't use git for version control often, so I hope I did it correctly.

lafrech:
Great. This looks pretty good.

To keep the test as minimal as possible, I think you should test the "simple string to string mapping" described here: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#model-with-mapdictionary-properties.

The test you're proposing also goes through schema definition resolution, which is out of scope for this unit test IMHO (I'm open to contradiction, here).

To do so, you can remove `AnalysisWithDictSchema` and just create within the test function a simple Schema with a `Dict(values=String())` field.

Thanks.

UrKr:
Ok, I changed the test to check a simple, non-nested schema with a Dict field. Sorry for the multiple Travis checks.

lafrech:
Great. One last thing. The case where `values` is not specified (equals `None`) is not covered. Please add a condition in swagger.py and complete the test function.

UrKr:
Ok, I added that, and made the tests more in line with the others by comparing against dict literal instead of accessing keys. 
```
assert result == {'type': 'object', 'additionalProperties': {'type': 'string'}}
assert result == {'type': 'object'}
```

lafrech:
Perfect. Thanks!

PR #203, opened 2018-05-05
Run tests on marshmallow 3

lafrech:
I just noticed the tests don't run on marshmallow 3.

I copied `travis.yml` file from webargs: https://github.com/sloria/webargs/blob/dev/.travis.yml.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,16 +6,25 @@ python:
 - '3.4'
 - '2.7'
 - pypy
+
+env:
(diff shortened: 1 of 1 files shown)


lafrech:
I checked all tests pass, and I checked in the test reports that MA3 was used.

@sloria, you may want to double-check. Especially the part that does the PyPI upload.

sloria:
Good catch. Thanks @lafrech

PR #205, opened 2018-05-07
Code cleanup + use flake8 3.5.0

lafrech:
Update flake8 and fix pylint errors when it makes sense.


lafrech:
Added a commit to use flake8-quotes.

sloria:
Thanks!

Issue #207, opened 2018-05-08
RFC: Use classes for plugins


lafrech:
I think things would be simpler if plugins were classes.

I read in the comments that the plugin system is copied on Sphinx. I don't know why Sphinx does it like this, but I don't see the benefit in apispec.

Since plugins are only referred to by their path as string, there is no direct access to them. So their configuration can be a bit awkward:

1/ Mutating a global variable in the plugin file.

2/ Calling a function in the plugin file that will have to store either in a global in the file (which is equivalent to case 1 above) or somewhere else (like `map_to_swagger_type` stores the mapping in the field's `__swagger_field_mapping` attribute).

3/ Acting upon the spec object. This is the case for `schema_name_resolver`, which is used only in marshmallow plugin, so shouldn't be exposed to `spec`.

Likewise, storage is more complicated as it is done in the `spec` object. The list of schema definitions is stored in `spec.plugins['apispec.ext.marshmallow']['refs']` which is the reason why the `spec` object is passed to all functions in the plugin.

With plugin classes, the plugin object would keep its storage internal. It wouldn't need globals. It could be configured at instantiation before (or even after, if needed) being passed to the `spec` object.

A plugin would just have to define such methods:

```python
class MyApispecPlugin(BaseApispecPlugin):
    def definition_helper(...):
        [...]

    def path_helper(...):
        [...]

    def operation_helper(...):
        [...]
```

Of course, `BaseApispecPlugin` would provide no-op methods so that child classes would only add needed methods.

Then, no need to store stuff in the spec object. Variables would be plugin class attributes, modifiable by direct access or plugin object method calls (this is plugin implementation detail). No need to expose plugin internal details to `spec` (`schema_name_resolver`). No need to add attributes to the field classes (schema -> swagger type mapping can be stored in a plugin object attribute).

Here's what setup would look like:

```python
# Instantiate and configure plugin
ma_plugin = apispec.ext.marshmallow.MarshmallowPlugin(
    schema_name_resolver = my_func
)

# Create an APISpec
spec = APISpec(
    title='Swagger Petstore',
    version='1.0.0',
    plugins=[
        apispec.ext.flask.FlaskPlugin(),  # No specific config to do beforehand
        ma_plugin,  # Configured above
    ],
)
```

@sloria, is there a reason I'm missing for the current plugin interface?

This refactor would solve part of #206.

sloria:
[...]

PR #211, opened 2018-05-14
Generalize schema resolution for arrays

cvlong:
Updates the marshmallow plugin to resolve an array of schema items in the request parameters. Generalizes `resolve_schema` to handle arrays defined in `parameters` or `responses`.

### DIFF
--- a/tests/test_ext_marshmallow.py
+++ b/tests/test_ext_marshmallow.py
@@ -466,6 +466,42 @@ class TestOperationHelper:
         assert op['parameters'][0]['schema'] == resolved_schema
         assert op['responses'][200]['schema'] == resolved_schema
 
+    def test_schema_array_in_docstring_uses_ref_if_available_v3(self, spec_3):
+        def pet_view():
+            """Not much to see here.
+
+            ---
+            get:
+                parameters:
+                    - in: body
+                      content:
+                        application/json:
+                            schema:
+                                type: array
+                                items: tests.schemas.PetSchema
+                responses:
+                    200:
+                        content:
+                            application/json:
+                                schema:
+                                    type: array
+                                    items: tests.schemas.PetSchema
+            """
+            return '...'
+
+        spec_3.add_path(path='/pet', view=pet_view)
+        p = spec_3._paths['/pet']
+        assert 'get' in p
+        op = p['get']
+        resolved_schema = {
+            'type': 'array',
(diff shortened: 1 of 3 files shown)


sloria (approved):
Looking good. I think this is the correct approach, and +1 to OAI 3 support.

It would be great to have a regression test for the OpenAPI 3 format. @cvlong Would you be up for adding it?

cvlong:
Thanks, @sloria. Added the test you suggested.

sloria:
Thanks @cvlong !

tests/test_ext_marshmallow.py
```diff
@@ -451,11 +456,51 @@ def pet_view():
...
-        assert op['responses'][200]['schema'] == {
+        resolved_schema = {
             'type': 'array',
             'items': {'$ref': '#/definitions/Pet'}
         }
+        assert op['parameters'][0]['schema'] == resolved_schema
+        assert op['responses'][200]['schema'] == resolved_schema
+
+    def test_schema_array_in_docstring_uses_ref_if_available_v3(self, spec_3):
+        def pet_view():
```
lafrech, line 470: I think this tests passes because `spec3.definition('Pet', schema=PetSchema)` is done in an earlier test.

lafrech: Sorry, I take that back. It passes even if ran alone.

It differs from the v2 test in that the v2 checks for the reference `{'$ref': '#/definitions/Pet'}` explicitly. AFAIU, this test does not check that the reference is used. Or it does, if the ref exists, but we're not sure about that.

I rebased the "classes Plugin" branch (https://github.com/marshmallow-code/apispec/issues/207) on this and I think it is fixed there.

cvlong: @lafrech let me know if you'd like me to change something here. I can adjust the test to check the reference with the following change, 

```python
        spec_3.definition('Pet', schema=PetSchema)
        spec_3.add_path(path='/pet', view=pet_view)
        ...
        resolved_schema = {
            'type': 'array',
            'items': {'$ref': '#/components/schemas/Pet'},
        }
        ...
```

lafrech: Yes, exactly. Like the v2 test above. This is what I did in https://github.com/marshmallow-code/apispec/tree/dev_207_plugin_classes.

PR #213, opened 2018-05-20
extensions: support pyramid view inspection

ergo:
Hi,

I've added pyramid web framework extension.

The PR includes tests (inspired by flask extension tests) and documentation updates.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -49,3 +49,4 @@ Contributors (chronological)
 - Louis-Philippe Huberdeau `@lphuberdeau <https://github.com/lphuberdeau>`_
 - Urban `@UrKr <https://github.com/UrKr>`_
 - Christina Long `@cvlong <https://github.com/cvlong>`_
+- Marcin Lulek `@ergo <https://github.com/ergo>`_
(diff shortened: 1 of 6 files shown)


sloria:
This is great! Thanks @ergo . I would be more comfortable with this as a 3rd-party plugin, as we're trying to keep this repo small and easy to maintain.

If you do end up releasing this as a separate library, please add to the Ecosystem page here: https://github.com/marshmallow-code/apispec/wiki/Ecosystem

peteclark3:
@ergo - Did this plugin end up living anywhere else?  I don't see it in the ecosystem.  Thanks!

ergo:
Hi @peteclark3 I need to publish it on github :) I'll try to do it ASAP.

@sloria would you accept a PR where existing web framework plugins are split into separate package, and then we import them into apispec back for backward compatibility? This way a nice separation could be maintained and at same time no existing application would break.

sloria:
I've considered splitting the plugins into separate repos/packages, but I don't have plans to do that in the immediate future, especially while the plugin API is not yet stable (see #219). Having the "officially-supported" plugins within this repo reduces maintenance burden, since we only have one release schedule to worry about. Also, most apispec users are also marshmallow users, and I think it's nice to have marshmallow support out of the box.

ergo:
@peteclark3 https://github.com/ergo/pyramid_apispec - there you go, I would love to get some feedback.

PR #217, opened 2018-05-27
Openapi 3.0.X support for shcemas in paramters in marshmallow extension

fMeow:
Currently, marshmallow extension only partially supports openapi 3.0.X . In my case, I would like to assign marshmallow schema object in parameters. 

The code below is a demonstration for my case.

```python

class PetSchema(Schema):
    id = fields.Int(dump_only=True)
    name = fields.Str()
    age = fields.Int()

def pet_view():
    """Not much to see here.
    ---
    get:
        parameters:
            - in: query
              schema: PetSchema
    """
    return '...'
```

The expected result in Openapi Specification 3.0.X is as below:

```yaml
get:
  parameters:
  - in: query
    name: name
    required: false
    schema: 
      type: string
  - in: query
    name: age
    required: false
    schema: 
      format: int32
      type: integer

```
In short, **format** and **type** field must be inside **schema** object. A parameter object do not allow **format** field nor **type** field.

However, current result is as follow:

```yaml
get:
  parameters:
  - in: query
    name: name
    required: false
    type: string
  - in: query
    name: age
    required: false
    format: int32
    type: integer
```

Thus, I modify the code to fix this behavior. Also, I notice that **collectionFormat** field is now deprecated in Openapi 3.0.X . Instead, an equivalent of collection set to **multi** is that **style** set to **form**, with **explode** set to true.
See [style values](https://swagger.io/specification/#style-values). Below is some important description or examples copied directly from openapi specification.

style | explode | empty | string | array | object
-- | -- | -- | -- | -- | --
form | true | color= | color=blue | color=blue&color=black&color=brown | R=100&G=200&B=150

```yaml
name: id
in: query
description: ID of the object to fetch
required: false
schema:
  type: array
  items:
    type: string
style: form
explode: true
```

### DIFF
--- a/apispec/ext/marshmallow/swagger.py
+++ b/apispec/ext/marshmallow/swagger.py
@@ -543,7 +543,7 @@ def property2parameter(prop, name='body', required=False, multiple=False, locati
         elif openapi_major_version == 3:
             if multiple:
                 ret['explode'] = True
-                ret['style'] = dict(form='array')
+                ret['style'] = 'form'
             ret['schema'] = prop
     return ret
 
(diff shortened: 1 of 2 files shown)


sloria:
This looks great. Thanks!

PR #219, opened 2018-05-27
Use classes for plugins

### LINKED ISSUE #207: RFC: Use classes for plugins

lafrech:
Closes #207.

I think I'm almost done with this.

There is still a few things to do:

- [x] Decide whether or not to provide a base class for plugin classes. @sloria, you seemed to prefer not to, but I don't see why. I think it would simplify plugin creation.
- [x] Be sure to complete tests to test Plugin interface
- [x] Document changes

### DIFF
--- a/apispec/core.py
+++ b/apispec/core.py
@@ -132,7 +132,11 @@ class APISpec(object):
         self._path_helpers = []
         self._operation_helpers = []
         # {'get': {200: [my_helper]}}
-        self._response_helpers = []
+        self._response_helpers = {}
+        self._plugin_definition_helpers = []
+        self._plugin_path_helpers = []
+        self._plugin_operation_helpers = []
+        self._plugin_response_helpers = []
 
         for plugin in self.plugin_objects:
             self.setup_plugin(plugin)
@@ -204,7 +208,7 @@ class APISpec(object):
                         operations=operations,
                         openapi_version=self.openapi_version.vstring)
         # Execute plugins' helpers
-        for func in self._path_helpers:
+        for func in self._plugin_path_helpers:
             try:
                 ret = func(path=path, operations=operations, **kwargs)
             except TypeError:
@@ -213,9 +217,24 @@ class APISpec(object):
                 ret.path = normalize_path(ret.path)
                 path.update(ret)
                 operations = ret.operations
+        # Deprecated interface
+        for func in self._path_helpers:
+            try:
+                ret = func(
+                    self, path=path, operations=operations, **kwargs
+                )
+            except TypeError:
+                continue
+            if isinstance(ret, Path):
+                ret.path = normalize_path(ret.path)
+                path.update(ret)
+                operations = ret.operations
         if operations:
-            for func in self._operation_helpers:
+            for func in self._plugin_operation_helpers:
                 func(path=path, operations=operations, **kwargs)
+            # Deprecated interface
+            for func in self._operation_helpers:
+                func(self, path=path, operations=operations, **kwargs)
 
         if not path.path:
             raise APISpecError('Path template is not specified')
@@ -223,10 +242,21 @@ class APISpec(object):
         # Process response helpers for any path operations defined.
         # TODO: optimize this?
         for method, operation in iteritems(path.operations):
-            if method in VALID_METHODS:
-                for status_code, response in iteritems(operation):
-                    for func in self._response_helpers:
+            if method in VALID_METHODS and 'responses' in operation:
+                for status_code, response in iteritems(operation['responses']):
+                    for func in self._plugin_response_helpers:
                         response.update(func(method, status_code, **kwargs))
+        # Deprecated interface
+        # Rule is that method + http status exist in both operations and helpers
+        methods = set(iterkeys(path.operations)) & set(iterkeys(self._response_helpers))
+        for method in methods:
+            responses = path.operations[method]['responses']
+            statuses = set(iterkeys(responses)) & set(iterkeys(self._response_helpers[method]))
+            for status_code in statuses:
+                for func in self._response_helpers[method][status_code]:
+                    responses[status_code].update(
+                        func(self, **kwargs)
+                    )
 
         self._paths.setdefault(path.path, path.operations).update(path.operations)
 
@@ -247,9 +277,15 @@ class APISpec(object):
         """
         ret = {}
         # Execute all helpers from plugins
-        for func in self._definition_helpers:
+        for func in self._plugin_definition_helpers:
             try:
                 ret.update(func(name, definition=ret, **kwargs))
+            except TypeError:
+                continue
+        # Deprecated interface
+        for func in self._definition_helpers:
+            try:
+                ret.update(func(self, name, definition=ret, **kwargs))
             except TypeError:
                 continue
         if properties:
@@ -267,13 +303,67 @@ class APISpec(object):
     def setup_plugin(self, plugin):
         plugin.init_spec(self)
         if hasattr(plugin, 'definition_helper'):
-            self._definition_helpers.append(plugin.definition_helper)
+            self._plugin_definition_helpers.append(plugin.definition_helper)
         if hasattr(plugin, 'path_helper'):
-            self._path_helpers.append(plugin.path_helper)
+            self._plugin_path_helpers.append(plugin.path_helper)
         if hasattr(plugin, 'operation_helper'):
-            self._operation_helpers.append(plugin.operation_helper)
+            self._plugin_operation_helpers.append(plugin.operation_helper)
         if hasattr(plugin, 'response_helper'):
-            self._response_helpers.append(plugin.response_helper)
+            self._plugin_response_helpers.append(plugin.response_helper)
+
+    def register_definition_helper(self, func):
+        """Register a new definition helper. The helper **must** meet the following conditions:
+
+        - Receive the `APISpec` instance as the first argument.
+        - Receive the definition `name` as the second argument.
+        - Include ``**kwargs`` in its signature.
+        - Return a `dict` representation of the definition's Schema object.
+
+        The helper may define any named arguments after the `name` argument.
+        ``kwargs`` will include (among other things):
+        - definition (dict): current state of the definition
+
+        https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#definitionsObject
+
+        :param callable func: The definition helper function.
+        """
+        self._definition_helpers.append(func)
+
+    def register_path_helper(self, func):
+        """Register a new path helper. The helper **must** meet the following conditions:
+
+        - Receive the `APISpec` instance as the first argument.
+        - Include ``**kwargs`` in signature.
+        - Return a `apispec.core.Path` object.
+
+        The helper may define any named arguments in its signature.
+        """
+        self._path_helpers.append(func)
+
+    def register_operation_helper(self, func):
+        """Register a new operation helper. The helper **must** meet the following conditions:
+
+        - Receive the `APISpec` instance as the first argument.
+        - Receive ``operations`` as a keyword argument.
+        - Include ``**kwargs`` in signature.
+
+        The helper may define any named arguments in its signature.
+        """
+        self._operation_helpers.append(func)
+
+    def register_response_helper(self, func, method, status_code):
+        """Register a new response helper. The helper **must** meet the following conditions:
+
+        - Receive the `APISpec` instance as the first argument.
+        - Include ``**kwargs`` in signature.
+        - Return a `dict` response object.
+
+        The helper may define any named arguments in its signature.
+        """
+        method = method.lower()
+        if method not in self._response_helpers:
+            self._response_helpers[method] = {}
+        self._response_helpers[method].setdefault(status_code, []).append(func)
 
 def validate_openapi_version(openapi_version_str):
     """Function to validate the version of the OpenAPI passed into APISpec
--- a/apispec/core.py
+++ b/apispec/core.py
@@ -8,7 +8,7 @@ import yaml
 
 from apispec.compat import iterkeys, iteritems, PY2, unicode
 from apispec.lazy_dict import LazyDict
-from .exceptions import APISpecError
+from .exceptions import PluginError, APISpecError
 
 VALID_METHODS = [
     'get',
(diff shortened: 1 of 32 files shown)


apispec/core.py
```diff
@@ -125,24 +124,38 @@ def __init__(self, title, version, plugins=(), info=None,
...
         self._path_helpers = []
         self._operation_helpers = []
         # {'get': {200: [my_helper]}}
         self._response_helpers = {}
-
-        for plugin_path in plugins:
+        old_plugins = list(p for p in plugins if isinstance(p, str))
+        if old_plugins:
+            warnings.warn(
+                'Old style plugins are deprecated. Use classes instead.', DeprecationWarning)
```
sloria: Note for the future: it'd be nice if this linked to the relevant docs.

```
Old-style plugins are deprecated. Use plugin classes instead. See https://apispec.readthedocs.io/en/latest/using_plugins.html.
```

apispec/core.py
```diff
@@ -319,6 +366,8 @@ def register_definition_helper(self, func):
 
         :param callable func: The definition helper function.
         """
+        warnings.warn(
+            'Helper functions are deprecated. Use plugin classes.', DeprecationWarning)
```
sloria: Same here: once docs are written, you can link to them:

```
Helper functions are deprecated. Use plugin classes instead. See https://apispec.readthedocs.io/en/latest/writing_plugins.html
```

sloria (commented):
I like the direction this is going. I don't see any major design issues.

I think you're right about the base plugin class simplifying plugin creation, since there's shared initialization logic across plugins. It would also make the plugin interface more "self-documenting". IDEs would be able to do static analysis, etc.

lafrech:
I moved `BasePlugin` out of `ext` and made it importable from `apispec`.

I also rebased. I hope to get this done and merged soon because the change is so important every rebase is a nightmare. In fact, I had to merge rather than rebase as the rebase was too complicated.

lafrech:
I added tests for the plugin interface, and docstrings in the `BasePlugin` class to explain what the helper methods should do.

I also reworked `add_path` a bit to fix some inconsistencies. Since this is a huge rework, it is tempting to change everything. I'm still not perfectly happy about the whole thing but this is out of scope so I'd rather postpone to later (potentially breaking) changes. For instance the `path` and `operations` parameters passed from `add_path` to the plugin helpers are redundant as operations is an attribute of path. Also, I don't find the API really consistent. Some helpers return values while some mutate their inputs, some may return `None`, some must return a `dict`... I tried to stay close to current interface.

The last missing part is the documentation.

I'm happy if I get some feedback before writing the docs so that I don't have to rewrite if some changes are needed in the code.

There is no "upgrading" section in the docs. May I stick to a few lines in the changelog explaining what to do when upgrading (along with a rewrite of the "writing plugins" section)?

sloria:
Re: API inconsistencies: I don't doubt there's some things to be cleaned up. But those don't necessarily need to block getting this merged, especially if they're risky to change. As for low-risk changes: feel free to clean things up as you see fit.

I'll try to review this more over the weekend if I have time, though I can't make any promises (I'm currently in the process of moving between houses).

> There is no "upgrading" section in the docs. May I stick to a few lines in the changelog explaining what to do when upgrading (along with a rewrite of the "writing plugins" section)?

Yep, that's perfect.

lafrech:
Rebased again.

I really hope to merge this soon because each merge is a PITA. Not only because there are so many changes, but also because this PR improves the test coverage quite a lot: all tests (or almost) are run against both OpenAPI v2 and v3 (while in dev, only a few specific tests are run against v3).

The difficulty here was not so much the merge itself, but fixing two tricky bugs that had just been added to dev that the tests in dev wouldn't show. Those are fixed in the merge commits, which is not so nice in terms of traceability.

I'm not so happy with `fields2parameters`. It has `default_in` defaulting to `body` which is invalid in OpenAPIv3, and its internal logic is a bit twisted. It might be worth splitting in two methods for v2 and v3. Anyway, this can be reworked later on.

apispec/ext/marshmallow/swagger.py
```diff
@@ -48,211 +50,15 @@
 CUSTOM_FIELD_MAPPING_ATTR = '__swagger_field_mapping'
 
 
-def map_to_swagger_type(*args):
```
sloria: This was part of the public API. Perhaps this should be exposed `MarshmallowPlugin` as an instance method?

apispec/core.py
```diff
@@ -213,35 +227,61 @@ def normalize_path(path):
...
-
         if not path.path:
             raise APISpecError('Path template is not specified')
 
-        # Process response helpers for any path operations defined.
+        # Execute operation helpers
+        for plugin in self.plugins:
+            try:
+                plugin.operation_helper(path=path, operations=path.operations, **kwargs)
+            except NotImplementedError:
```
sloria: It's possible for this to swallow `NotImplementedError` errors that aren't raised directly by plugin methods. I suppose `BasePlugin` could raise a plugin-specific exception, e.g. `PluginMethodNotImplementedError` to avoid this. Might not be absolutely necessary, though.

sloria (commented):
This is looking great. Thanks for all your work keeping this up to date.

Before we release this change, I think `swagger` and `Swagger` should be renamed `open_api` and `OpenAPIConverter`. But that doesn't need to happen before this PR is merged.

apispec/core.py
lafrech: Yes, I'm not sure sure it is that bad. I'm more uncomfortable with the `TypeError` caught above. I already experienced issues because it was hiding real problems in my code. `NotImplementedError` is more specific. But I don't mind changing this if you wish.

apispec/ext/marshmallow/swagger.py
lafrech: It is already exposed. Sort of. It is in the `Swagger` object.

```python
ma_plugin  = MarshmallowPlugin()

APISpec(
        title='My super API',
        version='0.1',
        plugins=(ma_plugin, ),
        openapi_version='3.0.1')

@ma_plugin.swagger.map_to_swagger_type('string', 'uuid')
class UuidField(StrField):
    [...]
```

I could add a `map_to_swagger_type` method to `MarshmallowPlugin` to proxy this one, but unless I add complicated or awkward logic, it would still have to be called after registration in `APISpec` instance, as this is when the `Swagger` instance is created.

I can't think of any simpler way to do this. We could create the `Swagger` object at init of the MA plugin, but it still wouldn't be totally usable until a spec object is set (just like the MA plugin) anyway.

Or maybe we can merge the MA plugin and the swagger object. Using composition, we'd keep them in separate files.

```python
class MarshmallowPlugin(BasePlugin, Swagger):
    """APISpec plugin handling marshmallow schemas
    [...]
    """
```

This way, all methods are accessible at MA plugin level.

Looks weird to me, but perhaps because I too rarely use composition.

lafrech:
> Before we release this change, I think `swagger` and `Swagger` should be renamed `open_api` and `OpenAPIConverter`.

Done. Except I chose `openapi` over `open_api`. This is the name of the version field in the spec (https://swagger.io/specification/#oasVersion).

lafrech:
Only `MarshmallowPlugin` uses the spec object.

I'm thinking of emptying `BasePlugin.init_spec` and store spec object as `spec` attribute only in `MarshmallowPlugin.init_spec`.

apispec/core.py
sloria: Yeah, I'm also concerned about TypeErrors being swallowed. I don't recall why that exception-handling was added. Would have to dig more into it.

lafrech: Perhaps was it made to avoid breaking on helper functions that would not have **kwargs in their signature?

I could remove it while I'm at it (no test broken on this branch if I do).

sloria: Ah, that seems likely. Yeah, go ahead and remove it.

lafrech: OK, I removed `TypeError` catching for the new plugin interface and left it on the legacy helpers.

I also introduced `PluginMethodNotImplementedError`.

lafrech:
Documentation updated.

TODO:

- [ ] Decide if `spec` should be stored as attribute in `BasePlugin.init_spec`. We can keep it as it is. No big deal.
- [x] Changelog. Should I wait until it is merged into `dev`?
- [ ] Perhaps add some `.. deprecated:: 0.38.0` directive in every plugin module's `setup` function. Version number unkown until merged into `dev`.

lafrech:
CHANGELOG updated.

Also, I'm hesitant to make `[bottle|flask]_path_to_swagger/openapi` Plugin class static method or leave them as global module functions. That's a minor detail.

@sloria, I think I'm done with this for now. Thanks (and good luck !) for the review.

sloria:
>  Decide if spec should be stored as attribute in BasePlugin.init_spec. We can keep it as it is. No big deal.

I suppose making `init_spec` no-op makes plugin classes more general-purpose. I don't have strong feelings about it either way, though. 

>  Perhaps add some .. deprecated:: 0.38.0 directive in every plugin module's setup function. Version number unkown until merged into dev.

Good idea. Feel free to assume 0.39.0 (i.e. `.. deprecated:: 0.39.0`) for now. I can always change those before I release.

> Also, I'm hesitant to make [bottle|flask]_path_to_swagger/openapi Plugin class static method or leave them as global module functions. That's a minor detail.

Yeah, it's personal preference. I tend to eschew static methods in favor of functions.

> I think I'm done with this for now. Thanks (and good luck !) for the review.

Great! I'll try to give this a final pass some time within the next week.

lafrech:
> I suppose making init_spec no-op makes plugin classes more general-purpose.

Done.

> add some .. deprecated:: 0.39.0 directive in every plugin module's setup function.

Done. 

> make [bottle|flask]_path_to_swagger/openapi Plugin class static method or global module functions

Left as static methods. I like to have all methods accessible from the class or object, rather than "scattered" in the module. You're right, that's a matter of personal preference.

> I'll try to give this a final pass some time within the next week.

Great, thanks.

apispec/exceptions.py
```diff
@@ -4,16 +4,15 @@
 
 class APISpecError(Exception):
     """Base class for all apispec-related errors."""
-    pass
 
 class PluginError(APISpecError):
     """Raised when a plugin cannot be found or is invalid."""
-    pass
+
+class PluginMethodNotImplementedError(APISpecError):
```
sloria: Minor: This should probably inherit from `NotImplementedError`.

docs/using_plugins.rst
```diff
@@ -195,18 +193,33 @@ If your API uses `method-based dispatching <http://flask.pocoo.org/docs/0.12/vie
...
+        version='1.0.0',
+        openapi_version='2.0',
+        info=dict(
+            description='A minimal gist API'
+        ),
+        plugins=[ma_plugin]
+    )
+
+
+    @ma_plugin.openapi.map_to_swagger_type('string', 'uuid')
```
sloria: This breaks the [Law of Demeter](https://www.wikiwand.com/en/Law_of_Demeter) by accessing a method on an internal attribute. Also, it's a lot of typing. Perhaps we should expose `map_to_swagger_type` on `MarshmallowPlugin`?

docs/writing_plugins.rst
```diff
@@ -3,12 +3,15 @@
 Writing Plugins
 ===============
 
-Helper Functions
-----------------
+A plugins is a child class of `apispec.plugin.BasePlugin`.
```
sloria: Revision: "A plugin is a subclass of `apispec.plugin.BasePlugin`."

docs/writing_plugins.rst
```diff
@@ -17,67 +20,73 @@ There are four types of helper functions:
...
+            return Path(path=path, operations=operations)
 
-The ``setup`` Function
-----------------------
 
-All plugins **must** define a ``setup`` function that receives an `apispec.APISpec` instance as its only argument. The ``setup`` function registers your plugin's helper functions with the ``spec``.
+The ``init_spec`` Method
+------------------------
 
+`BasePlugin` has an `init_spec` method that `APISpec` calls on each plugin at initialization with the spec object itself as parameter. It is a no-op, but it can be overriden by a plugin to get useful information from the spec object.
```
sloria: Revision suggestion: "It is no-op by default, but a plugin may override it to access and store useful information from the spec object."

docs/writing_plugins.rst
```diff
@@ -17,67 +20,73 @@ There are four types of helper functions:
...
-        """
-        operations = load_operations_from_docstring(func.__doc__)
-        return Path(path=path, operations=operations)
+    class DocPlugin(BasePlugin):
+
+        def path_helper(self, path, func, **kwargs):
+            """Path helper that parses docstrings for operations. Adds a
+            ``func`` parameter to `apispec.APISpec.add_path`.
+            """
+            openapi_major_version = self.spec.openapi_version.major
```
sloria: This would require `spec` to be stored in `init_spec`.

tests/conftest.py
```diff
@@ -0,0 +1,41 @@
...
+from collections import namedtuple
+
+import pytest
+
+from apispec import APISpec
+from apispec.ext.marshmallow import MarshmallowPlugin
+
+
+@pytest.fixture(params=(
+    ('2.0', True),
```
sloria, line 13: Nice! 👍

sloria (changes requested):
This looks excellent. Just a few minor changes suggested, but otherwise this looks good to go.

apispec/exceptions.py
lafrech: Done.

```python
class PluginMethodNotImplementedError(APISpecError, NotImplementedError):
    """Raised when calling an unimplemented helper method in a plugin"""
```

docs/using_plugins.rst
lafrech: Done.

lafrech:
Thanks for the review. That's a huge change. We may have left a few glitches. Nothing big, hopefully.

I addressed all your comments.

sloria (approved):

sloria:
I'll cut a release after work today.

lafrech:
No rush. Thanks.

PR #223, opened 2018-06-01
fix description for parameters in OpenAPI 3

fMeow:
In OpenAPI3, desired output for query parameters is like this:
```yaml
parameters:
- description: Password
  in: query
  name: password
  required: true
  schema: 
    type: string
```

However,  only properties *in*, *name* and *required* are defined in the top level, while all other properties is moved under property *schema*, which conflicts with OpenAPI 3 specification.

Current output:
```yaml
parameters:
- in: query
  name: password
  required: true
  schema: 
    type: string
    description: Password
```

So before moving all properties to schema, it's reasonable to check for *description* property and define description in top level of parameter object.

### DIFF
--- a/apispec/ext/marshmallow/swagger.py
+++ b/apispec/ext/marshmallow/swagger.py
@@ -427,6 +427,13 @@ def fields2parameters(fields, schema=None, spec=None, use_refs=True,
     the :class:`Schema <marshmallow.Schema>`.
 
     https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameterObject
+
+    For OpenAPI 3, only "query", "header", "path" or "cookie" is allowed for the location
+    of parameters. In OpenAPI 3, Request Body Object is used when parameters is in body.
+    So this function always return an array of a parameter for each included field in
+    the :class:`Schema <marshmallow.Schema>`.
+
+    https://swagger.io/specification/#parameterObject
     """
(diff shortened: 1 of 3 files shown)


sloria:
Thanks again @Guoli-Lyu . Can you add a test for this, too? Once that's added, this can be merged.

fMeow:
I add test for descriptions in parameters and also requestBody, which in OpenAPI 2 is like this:

```yaml
parameters:
- in: path
  name: name
```
@sloria

apispec/ext/marshmallow/swagger.py
```diff
@@ -427,14 +427,21 @@ def fields2parameters(fields, schema=None, spec=None, use_refs=True,
     the :class:`Schema <marshmallow.Schema>`.
 
     https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameterObject
+
+    For OpenAPI 3, only "query", "header", "path" or "cookie" is allowed for the location
```
sloria, line 431: Minor revisions:

```
For OpenAPI 3, only "query", "header", "path" or "cookie" are allowed for the location of parameters. "requestBody" is used when fields are in the body.

This function always returns a list, with a parameter for each included field in the :class:`Schema <marshmallow.Schema>`.
```

sloria (commented):
Looks good. Thanks for adding the test. Just some minor revisions to the docs needed. I can make those myself before I merge.

PR #224, opened 2018-06-09
Use a Python library for validating specs

sloria:
Switch from check_api to prance for validating specs to avoid
having to install a global npm package for validation.

Validation requirements can be installed by adding the [validation]
extras, i.e. pip install 'apispec[validation]'

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -32,11 +32,6 @@ Setting Up for Local Development
 
     # After activating your virtualenv
     $ pip install -r dev-requirements.txt
-    $ npm install -g check_api
-
-3. Install apispec in develop mode. ::
-
-   $ pip install -e .
 
 Git Branch Structure
 ++++++++++++++++++++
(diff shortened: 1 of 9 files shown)


CONTRIBUTING.rst
```diff
@@ -32,11 +32,6 @@ Setting Up for Local Development
 
     # After activating your virtualenv
     $ pip install -r dev-requirements.txt
-    $ npm install -g check_api
-
-3. Install apispec in develop mode. ::
```
sloria, line 37: This is no longer necessary because `pip install -r dev-requirements.txt` takes care of this.

lafrech:
Sounds great.

Why not just add it to `dev-requirement.txt` and leave `setup.py` unchanged?

sloria:
This allows users to install the validation requirements so that they can use `apispec.utils.validate_swagger`.

lafrech:
You're right. I'll probably be using this in my framework to emit a warning when the generated spec is invalid. Or to validate the spec in my units tests. This is a real improvement.

sloria:
I'm also planning to rename `validate_swagger` to `validate_spec` (`validate_swagger` will raise a DeprecationWarning). I'll do that after merging this.

sloria:
Went ahead with the rename as a part of this PR.

PR #228, opened 2018-06-13
Extension to merge documentation through the decorator stack

brunoais:
This extension allows spreading excepts of pydoc documentation throughout decorators which are then used to merge back into a complete documentation (see the py file's pydoc).
It requires using the `doc_wraps()` included and documented in the extension, in place of `functools.wraps()`. See the included pydoced code for an example.

I'm another person who is not good with naming variables but I hope the sames I chose were good enough. Feel free to discuss variable name changes as you see fit.
Ask anything that is not clear enough, so I can understand which parts are not as easy to read. I hope this can also help other people and not just me.

Is this a good feature to be later accepted?
I need the confirmation before I do code preparing for a merge that won't happen

TODO (in order):

- [X] Make the code functional
- [X] Make PR with some explanation
- [ ] Get author's approval to continue
- [ ] Change the code to use the helpers and normalizers so it works with both python 2.7 and python 3.6 
- [ ] Change the code to comply with coding guidelines required.
- [ ] Create tests
- [ ] Discuss changes and adjustments (again)
- [ ] Get author's approval to continue
- [ ] Add myself to `AUTHORS.rst`
- [ ] Add related documentation in `docs/using_plugins.rst`
- [ ] Register the addition in `docs/changelog.rst`
- [ ] Create followup issue/PR to add compatibility with OpenAPI 3.0

PR #229, opened 2018-06-17
Invalid data removal operation helper

brunoais:
This code cleans up elements that are not usable for the current setup. E.g. a body parameter in a `GET` request or a path parameter that is not specified in the path of the function itself.

This is specially useful when the same function is reused for different paths (which uses different path variables) or even for different methods (serving a POST and a GET, at the same time, while the GET one doesn't receive an Entity)

TODO (in order):

- [X] Make the code functional
- [X] Make PR with some explanation
- [ ] Get author's approval to continue
- [ ] Change the code to use the helpers and normalizers so it works with both python 2.7 and python 3.6 
- [ ] Change the code to comply with coding guidelines required.
- [ ] Create tests
- [ ] Discuss changes and adjustments (again)
- [ ] Get author's approval to continue
- [ ] Add myself to `AUTHORS.rst`
- [ ] Add related documentation in `docs/using_plugins.rst`
- [ ] Register the addition in `docs/changelog.rst`
- [ ] Create followup issue/PR to add compatibility with OpenAPI 3.0

### DIFF
--- a/apispec/ext/remove_incoherent_data.py
+++ b/apispec/ext/remove_incoherent_data.py
@@ -1,0 +1,116 @@
+# -*- coding: utf-8 -*-
+"""Operations helper that allows you to remove unusable configurations given the rules of OpenAPI and the rules of HTTP
+This extension removes:
+
+- Path parameters if the parameter is not in the path
+- Body parameters if the method is entityless (bodyless)
+- Response schemas if the method requires an answer without entity (body)
+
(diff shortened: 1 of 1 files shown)


sloria:
See https://github.com/marshmallow-code/apispec/pull/228#issuecomment-397920736 . In the interest of keeping the core small, this should be released as a separate plugin. Thanks anyway!

PR #238, opened 2018-07-06
Refactor add_path even more

lafrech:
This is based on https://github.com/marshmallow-code/apispec/pull/237 and goes further. `Path` is removed.

Clearer API (IMHO), and almost no test broken.

Probably not perfect, but I think it is a nice improvement.

@sloria, I'd understand your reluctance to introduce breaking changes. Would it make you feel more comfortable to publish a 1.0 version then break everything in 2.x? I don't mind either way. The breaking changes introduced here are not much worse than those introduced with the class interface. They will break every custom path helper, and possibly some operation helpers, marginally, but the fix is easy. It is not as if it made some feature impossible.

### DIFF
--- a/apispec/__init__.py
+++ b/apispec/__init__.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 """Contains the main `APISpec` class.
 """
-from .core import APISpec, Path
+from .core import APISpec
 from .plugin import BasePlugin
 
 __version__ = '0.39.0'
@@ -11,6 +11,5 @@ __license__ = 'MIT'
 
 __all__ = [
     'APISpec',
-    'Path',
     'BasePlugin',
 ]
--- a/apispec/core.py
+++ b/apispec/core.py
@@ -219,7 +219,7 @@ class APISpec(object):
 
         https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#pathsObject
 
-        :param str|Path|None path: URL Path component or Path instance
+        :param str|None path: URL path component
         :param dict|None operations: describes the http methods and options for `path`
         :param dict kwargs: parameters used by any path helpers see :meth:`register_path_helper`
         """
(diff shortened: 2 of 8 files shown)


lafrech:
I shouldn't have labelled this WIP.

@sloria, I don't mean to rush you if you saw that but didn't have the time to look at it. If it was non-breaking I would merge it myself and wait for you to release, but since it is a bit more than that, I'm requesting your review.

sloria (approved):
Yeah, I saw this and just hadn't had the time to review it in earnest.

I'm good with this change. It was admittedly pretty ugly to allow operations to be passed both in `Path` and as a kwarg to `add_path`, so I'm all for cleaning that up. And +1 for reducing the API surface.

How about we release this in 0.40.0 then remove all deprecated API in 1.0.0?

sloria:
Actually, let's make this 1.0.0b1 then remove all deprecated API in 1.0.0 final.

lafrech:
> Yeah, I saw this and just hadn't had the time to review it in earnest.

Sorry for rushing you on this. I wasn't sure how much you rely on @ ping and/or review requests to reach your attention. (I don't.)

> Actually, let's make this 1.0.0b1 then remove all deprecated API in 1.0.0 final.

As you wish. We can remove all deprecated stuff right now since we're now in beta: it won't be pulled unless asked explicitly with `--pre`.

I had a few other things in mind. I'm not sure they'll make it to v1.0 but if these are breaking changes, that'll be a 2.x.

sloria:
I want to give users an easy transition to 1.0, hence allowing for one beta release with the deprecated API. We can remove it in 1.0.0b2 if we end up needing another beta.

Feel free to tag any issues with the 1.0 milestone. Otherwise, I'll just plan on releasing 1.0.0 as is, sans deprecated API, in a week or two.

alysivji:
@sloria I'm writing a plugin for falcon which is using the new `add_path`. When will the 1.0.0bx release be up on PyPI?

Thanks for this great library!

sloria:
@alysivji I've gone ahead and cut the 1.0.0b1 release.

PR #247, opened 2018-07-16
Rework documentation of fields default values

### LINKED ISSUE #196: Documenting callable default
When a default/missing value is a callable, it is called when generating the documentation.

```python
    if default is not marshmallow.missing:
        if callable(default):
            ret['default'] = default()
        else:
            ret['default'] = default
```

This can be wrong if the value of the callable depends on the moment it is called. Typically, if it is `dt.datetime.now`. There should be an opt-out for this allowing to specify an alternative text to write in the docs.

Not sure how to do that. Maybe a `default_doc` attribute in the Schema that would take precedence over default/missing.

lafrech:
Closes #196.

I couldn't come up with a better name than `doc_default`.

### DIFF
--- a/apispec/ext/marshmallow/openapi.py
+++ b/apispec/ext/marshmallow/openapi.py
@@ -319,10 +319,7 @@ class OpenAPIConverter(object):
             ret['format'] = fmt
 
         default = field.missing
-        if default is not marshmallow.missing:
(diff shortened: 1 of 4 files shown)


sloria:
This needs docs, but this is good to merge as is. 🚢 🇮🇹

Issue #246, opened 2018-07-16
Refactor: remove MarshmallowPlugin.path_helper, make YAML optional,...


lafrech:
We have two types of plugins:

- web framework plugins:
  - get path from view object (with a [limitation to one path per view](https://github.com/marshmallow-code/apispec/blob/dev/apispec/ext/flask.py#L107) in Flask, see #181)
  - populate `operations` with info parsed from YAML docstrings

- marshmallow plugin
  - populate `operations` with info parsed from YAML docstrings
  - resolve marshmallow schemas in operations

Several remarks:

The path helper in marshmallow plugin is a generic path helper that expects a view function with a docstring. I'm wondering if it is of any use, since anyone using a web framework would use the dedicated path helper. Assuming a web framework is simple enough not to need more than this raw helper, then it should be `SimpleWebFrameworkPlugin.path_helper` anyway. This is not marshmallow specific at all. I'd be tempted to remove it. (It would break a few tests.) In my framework, I use `MarshmallowPlugin` but I override its path helper with a method that raises `PluginMethodNotImplementedError` to remove that feature.

The `WebFrameworkPlugin`s (flask,...) are more like `YAMLWebFrameworkPlugin`s. I'd be happy to make the YAML requirement optional. I don't use it, as I get all necessary information from the decorators I use to pass serialization and deserialization schemas. I don't think `flask-apispec` uses this either, although I didn't check. If we got rid of `MarshmallowPlugin.path_helper` already, then it's only about importing YAML stuff only in web framework plugins modules.

BTW, `load_yaml_from_docstring` and `load_operations_from_docstring` could be reworked to return `{}` rather than `None` and avoid that `or` test in each plugin: 

```python
    utils.load_operations_from_docstring(view.__doc__) or {}
```

Once we're there, we'll have `MarshmallowPlugin` with only an `operation_helper`, and a few web framework plugins with only a `path_helper`. And only one `path_helper` returning a path string should be registered at a time, as the last one speaking overrides the others. Anyway, only one should be needed.

Also the web framework path helpers deal with operations quite a lot, and perhaps this logic should be split into `path_helper` and `operation_helper`. From a quick glance, the coupling between the two seems loose. Could be a bit trickier with flask due to the `if method in rule.methods` test, which covers what looks like a twisted but legit case (method declared in the view but not in the rule).

[...]

PR #249, opened 2018-07-17
Remove MarshmallowPlugin.path_helper

lafrech:
Addresses part of https://github.com/marshmallow-code/apispec/issues/246.

### DIFF
--- a/tests/test_ext_marshmallow.py
+++ b/tests/test_ext_marshmallow.py
@@ -232,38 +232,147 @@ class TestOperationHelper:
             return '#/definitions/'
         return '#/components/schemas/'
 
+    @pytest.mark.parametrize('pet_schema', (PetSchema, PetSchema(), 'tests.schemas.PetSchema'))
     @pytest.mark.parametrize('spec_fixture', ('2.0', ), indirect=True)
-    def test_schema_v2(self, spec_fixture):
-        def pet_view():
-            return '...'
-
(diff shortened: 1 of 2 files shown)


tests/test_ext_marshmallow.py
```diff
@@ -232,201 +232,110 @@ def ref_path(spec):
             return '#/definitions/'
         return '#/components/schemas/'
 
+    @pytest.mark.parametrize('pet_schema', (PetSchema, PetSchema(), 'tests.schemas.PetSchema'))
```
sloria, line 235: 👍 Nice!

sloria (approved):
This looks good. It might be good to have an integration test that combines the marshmallow plugin with a framework plugin to make sure Schema names in docstrings can be parsed.

Not necessarily a blocker though. This otherwise looks good to merge. Feel free to add the integration test now or in a followup PR; I'll plan to merge this over the weekend.

lafrech:
> It might be good to have an integration test that combines the marshmallow plugin with a framework plugin to make sure Schema names in docstrings can be parsed.

This is tested in `test_ext_order.py`. In https://github.com/marshmallow-code/apispec/issues/246 I suggest we remove this test as the order may matter, but we could keep the part that tests the right order. Basically, rename the test to "test_web_framework_and_marhsmallow" and test only [framework plugin, marshmallow] order.

sloria:
Sounds good to me.

PR #259, opened 2018-07-31
Remove deprecated features

lafrech:


sloria:
This will need a changelog update, but that can happen in a future PR/commit.

PR #292, opened 2018-09-21
Rework APISpec signature

lafrech:
Remove `info`: this is already covered by `**options` anyway, so it is redundant. This change is almost not breaking (only if all parameters are passed as positional).

Make `openapi_version` mandatory, no default version. There is no reason the version should default to `2.0`.

In the tests, I also removed `description=...` in the calls to `APISpec` because this is wrong. It should go in `info={'description': ...}`. Anyway, it is not needed for the test.

TODO:

- [x] Changelog
- [x] Docs: almost ready, just double-check. Bottom of writing_plugins the call to `APISpec` misses `openapi_version`...

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,15 @@
 Changelog
 ---------
 
+1.0.0b3 (unreleased)
+++++++++++++++++++++
+
+Features:
+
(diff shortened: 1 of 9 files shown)


sloria (approved):
I like these changes. I wasn't sure about making `openapi_version` a required argument, but I think it will make more sense as OpenAPI 3.0 becomes more widely used and supported.

Issue #302, opened 2018-10-05
Split out web framework plugins into separate package


ergo:
Hi,  

Now that Apispec nears 1.0 release, and given how many issues are open right now I was thinking that maybe web framework plugins should be living in a separate `apispec_plugins` package.
For bw. compatibility existing namespace `apispec.ext.$framework` could pull them back from separate package.

This would allow plugin authors to fix bugs in implementations or support newer versions of frameworks without the need to release apispec itself, or even add new plugins.

I will be attending a hackathon soon and we could probably tackle this.

sloria:
I think this could be a good idea. @lafrech Any objections?

lafrech:
Thanks @ergo for offering your help on this.

I find it useful to be able to test all bundled frameworks while modifying apispec, so it is practical to have them in the tests. I also agree that once apispec reaches 1.0, it should not be as necessary.

You know what? Go for it, and let's keep those plugins as dev-requirements and keep the tests for now. Besides, some internal features may be only covered by web framework plugins tests. Not sure but we could check that.

No need to ensure backward compatibility for 1.0 so no need to republish.

I think we can keep marshmallow in.

Should those plugins be hosted in marshmallow-code organization? @sloria, could you create the repos when code is ready?

lafrech:
I just added [*plugins* label](https://github.com/marshmallow-code/apispec/labels/plugins) to issues that should be forwarded to plugins bugtrackers.

I also updated [1.0 milestone](https://github.com/marshmallow-code/apispec/milestone/2).

ergo:
Hi I've added #304 , the repo for the package is https://github.com/ergo/apispec-ext-webframeworks - should I change ownership to marshmallow organisation?
I've tried to do things making the change transparent to end users. There probably is a way to move framework tests to separate package and test both in single run.

ergo:
I did not want to register any packages yet to solve the chicken/egg problem for test runner since I don't know the conventions Marshmallow project wants to follow here.

ergo:
I've made `invoke test` run tests from both packages at once, so testing all the frameworks during apispec development is still possible @lafrech.

lafrech:
Thanks @ergo for tackling this.

Thoughts, open to discussion:

- Shouldn't we make an extension per web framework? (apispec-ext-flask,...) Each extension would hard-depend on its framework, with a minimal compatible framework version (E.g. flask>=0.12).  

[...]

PR #304, opened 2018-10-07
#302 Move all web framework plugins to apispec-ext-webframeworks package

ergo:
This moves all the code to `apispec-ext-webframeworks`, apispec will pull it in as dependency during install, at this point nothing should be different from 3rd party perspective and everything should be backwards compatible. All tests pass.


ergo:
The tests fail because chicken and egg problem, pip tries to fetch a package that does not live in pypi yet.

Issue #245, opened 2018-07-16
Create methods to register all component objects


lafrech:
Currently, definitions can be registered to be referenced to in the spec.

The spec allows to register other objects like parameters or responses.

OpenAPI 3: https://swagger.io/specification/#componentsObject

Some of them also exist in OpenAPI 2. Search for "Swagger Object" in https://swagger.io/specification/v2/.

We'd rather get this right with all those objects in mind rather than address them independently each time everyone needs one of them.

It should be easy to create methods to register each object type. Or should we create a single method and pass the type as argument (and deprecate `definition`)?

Then it's less obvious how to help the developer to use references. Referencing schema definition is automatic. Referencing responses probably wouldn't be, so the user would have to enter the reference response name. Ideally, we'd provide a function that at least takes care of the `$ref` syntax allowing the user to just enter the name of the ref.

lafrech:
For instance, https://github.com/marshmallow-code/apispec/commit/3556849f815d42e28f84ee60cc518af00c988cc1 added the possibility to pass parameters a while ago.

We'd need to either do the same with each component type (response, security, etc.), or create a single method to manage them all. Since it's less than a ten of them, a method for each is reasonable. Besides, it makes it easier to do step-by-step.

Also, there should be hooks to allow plugins to provide helpers. For instance, there could be parameter helpers so that marshmallow schemas can be parsed when registered as parameters, response helpers, header helpers,...

Referencing could be done as was done for parameters: the developer passes the parameter name as a string and the `clean_operation` function blindly uses a ref when it encounters a string rather than a dict, assuming it has to be a reference: `{'$ref': '#/parameters/' + x}`, without any check. That's good enough for me. At least for a first step. This just takes a modification of `clean_operation` for each component type.

Overall, this issue is two-fold:

- Add methods to register components
- Add hooks for plugin helpers in those methods (including already existing `add_parameter` method)

lafrech:
Currently, existing methods are

- `definition`
- `add_parameter`

It would be nice to have consistent naming.

Should it be

- `definition`, `parameter`,...
- `add_definition`, `add_parameter`,...
- `register_definition`, `register_parameter`,...

Or a generic `register_component` method that accepts a type as first argument?

lafrech:
#305 proposes another approach: use a `Components` class to hold those components. I'm not convinced it is the best approach. I found in nicer when I did it, but I don't have a strong feeling about it.

[...]

PR #305, opened 2018-10-07
Rework spec: add Components class

lafrech:
I've been trying to work on https://github.com/marshmallow-code/apispec/issues/245.

This PR adds a `Components` class that stores the components. Components is v3 terminology. In v2, those are just top-level fields.

For now, it only has `add_schema` (formerly `spec.definition`) and `add_parameter` (formerly `spec.add_parameter`). It is a first step before adding other methods like `add_response`, `add_security_scheme`, etc (the point of #245).

I'm using v3 terminology because it is the future (hopefully...), but this is v2 compatible. I could even alias `add_definition` to `add_scheme` but I don't think it is necessary.

There's nothing mandatory in adding this `Components` class, we could keep all `add_xxx` methods in `APISpec`. I just thought it would be nicer this way.

From user perspective, the only difference is that

    spec.add_parameter(...)
    spec.definition(...)

become

    spec.components.add_parameter(...)
    spec.components.add_schema(...)

~The first commit is just a modification to the test to make sure we test the result (the doc that is produced) and not the internals (spec._definitions,...). It can be merged independently.~ (Done.)

Waiting for feedback about this before addressing the other components (reponses,...).

TODO:

- [x] Update changelog
- [x] Update docs

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,37 @@ Changelog
 1.0.0b5 (unreleased)
 ++++++++++++++++++++
 
+Features:
+
+- ``apispec.core.Components`` is added. Each ``APISpec`` instance has a
+  ``Components`` object used to define components such as schemas, parameters
+  or reponses. "Components" is the OpenAPI v3 terminology for those reusable
+  top-level objects.
+- ``apispec.core.Components.parameter`` and ``apispec.core.Components.response``
+  are added.
+- *Backwards-incompatible*: ``apispec.APISpec.add_path`` and
+  ``apispec.APISpec.add_tag`` are renamed to ``apispec.APISpec.path`` and
+  ``apispec.APISpec.tag``.
+- *Backwards-incompatible*: ``apispec.APISpec.definition`` is moved to the
+  ``Components`` class and renamed to ``apispec.core.Components.schema``.
+
+::
+
+    # apispec<1.0.0b5
+    spec.add_tag({'name': 'Pet', 'description': 'Operations on pets'})
+    spec.add_path('/pets/', operations=...)
+    spec.definition('Pet')
+    def PetSchema(Schema)
+        ...
+
+    # apispec>=1.0.0b5
+    spec.tag({'name': 'Pet', 'description': 'Operations on pets'})
+    spec.path('/pets/', operations=...)
+    spec.components.schema('Pet')
+    def PetSchema(Schema)
+        ...
+
+
 Deprecations/Removals:
 
 - The ``response_helper`` feature is removed. The same can be achieved from
@@ -12,7 +43,7 @@ Deprecations/Removals:
 1.0.0b4 (2018-10-28)
 ++++++++++++++++++++
 
-* *Backwards-incompatible*: ``apispec.ext.flask``,
+- *Backwards-incompatible*: ``apispec.ext.flask``,
   ``apispec.ext.bottle``, and ``apispec.ext.tornado`` are moved to
   a separate package, `apispec-webframeworks <https://github.com/marshmallow-code/apispec-webframeworks>`_.
   (:issue:`302`).
--- a/apispec/core.py
+++ b/apispec/core.py
@@ -181,19 +181,6 @@ class APISpec(object):
         ret = deepupdate(ret, self.options)
         return ret
 
-    # Backward compatigility
-    def add_parameter(self, param_id, location, **kwargs):
-        self.components.add_parameter(param_id, location, **kwargs)
-
-    # Backward compatigility
-    def definition(
-            self, name, properties=None, enum=None, description=None, extra_fields=None,
-            **kwargs
-    ):
-        self.components.add_schema(
-            name, properties=properties, enum=enum, description=description, extra_fields=extra_fields,
-            **kwargs)
-
     def to_yaml(self):
         """Render the spec to YAML. Requires PyYAML to be installed."""
         from .yaml_utils import dict_to_yaml
--- a/apispec/core.py
+++ b/apispec/core.py
@@ -102,7 +102,7 @@ class Components(object):
             schemas_key: self._schemas,
         }
 
-    def add_schema(
+    def schema(
             self, name, properties=None, enum=None, description=None, extra_fields=None,
             **kwargs
     ):
@@ -140,7 +140,7 @@ class Components(object):
             ret.update(extra_fields)
         self._schemas[name] = ret
 
-    def add_parameter(self, param_id, location, **kwargs):
+    def parameter(self, param_id, location, **kwargs):
         """ Add a parameter which can be referenced.
 
         :param str param_id: identifier by which parameter may be referenced.
@@ -158,7 +158,7 @@ class Components(object):
                 continue
         self._parameters[param_id] = ret
 
-    def add_response(self, ref_id, **kwargs):
+    def response(self, ref_id, **kwargs):
         """Add a response which can be referenced.
 
         :param str ref_id: ref_id to use as reference
@@ -227,14 +227,14 @@ class APISpec(object):
         from .yaml_utils import dict_to_yaml
         return dict_to_yaml(self.to_dict())
 
-    def add_tag(self, tag):
+    def tag(self, tag):
         """ Store information about a tag.
 
         :param dict tag: the dictionary storing information about the tag.
         """
         self._tags.append(tag)
 
-    def add_path(self, path=None, operations=None, **kwargs):
+    def path(self, path=None, operations=None, **kwargs):
         """Add a new path object to the spec.
 
         https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#pathsObject
--- a/apispec/core.py
+++ b/apispec/core.py
@@ -127,7 +127,7 @@ class Components(object):
         # Execute all helpers from plugins
         for plugin in self._plugins:
             try:
-                ret.update(plugin.definition_helper(name, definition=ret, **kwargs))
+                ret.update(plugin.schema_helper(name, definition=ret, **kwargs))
             except PluginMethodNotImplementedError:
                 continue
         if properties:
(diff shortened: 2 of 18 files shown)


lafrech:
I went ahead with response definitions.

I also added parameter/response components helpers to `BasePlugin`. I had to call them xxx_component helpers because response_helper already exists (see https://github.com/marshmallow-code/apispec/issues/315).

We don't need all types of component register methods and plugin helpers implemented to merge this, but those are the most important and they illustrate the point.

@sloria do you like the direction this is taking? Anything I should rework before this can be merged?

sloria:
@lafrech Yes, this is looking good. I was a bit hesitant about it at first, but following your logic in https://github.com/marshmallow-code/apispec/issues/245#issuecomment-427745748 , it makes sense. I do like how it follows the v3 terminology. 

Kinda reminds me of elasticsearch-dsl's [aggregation API](https://elasticsearch-dsl.readthedocs.io/en/latest/search_dsl.html).

```python
s = Search()
s.aggs.bucket('tags_per_day', 'terms', field='tags')
```

It might be because I'm used to elasticsearch-dsl, but I have a slight preference towards the noun forms, e.g. `spec.components.definition`, for their terseness. It allows for more natural chaining, like so:

```python
spec.components\
   .definition('Gist', schema=GistSchema)\
   .parameter('GistParams', schema=GistParamSchema)\
    # ...
```

Thoughts?

lafrech:
Nouns are fine. Let's go for nouns.

lafrech:
Questions

- Rename `definition` (current apispec and OpenAPI v2 terminology) into `schema` (OpenAPI v3 terminology)? The change is breaking already in both cases.

- OK for removing `add_` from `add_parameter` and other `add_component` methods. Should I do the same with `add_tag` and `add_path`, which are methods of spec, not of spec.components, due to OpenAPIv3 terminology?

Assuming the answers are _keep definition_ and  _remove `add_` everywhere_, we'd get:

```py
spec.path(...)
spec.tag(...)
spec.components.definition(...)
spec.components.parameter(...)
spec.components.response(...)
```

- Rename `definition_helper` into `schema_helper`? The former is more explicit, but it could be because we're used to it. We should be consistent with the component name above.

- Remove `component` from the component helpers? It makes the name a bit long, but I had to do that for the response component helper, because a `response_helper` already exists. In https://github.com/marshmallow-code/apispec/issues/315, I suggest the existing `response_helper` could be removed...

sloria:
> Rename definition (current apispec and OpenAPI v2 terminology) into schema (OpenAPI v3 terminology)? The change is breaking already in both cases.

If we're moving towards OpenAPI v3 terminology (which I think is a good idea), then we should use "schema" rather than "definition."

> OK for removing add_

Yes, remove add_ for consistency in all the method names.

> Rename definition_helper into schema_helper? The former is more explicit, but it could be because we're used to it.

"schema_helper" makes sense for consistency. Why do you say "definition_helper" is more explicit?

> Remove component from the component helpers? It makes the name a bit long, but I had to do that for the response component helper, because a response_helper already exists. In #315, I suggest the existing response_helper could be removed...

Since I can't recall the use case for the current `response_helper`, I'm fine getting rid of it. Then we can go with the more concise `parameter_helper`, `response_helper`, etc.

lafrech:
> "schema_helper" makes sense for consistency. Why do you say "definition_helper" is more explicit?

I meant by itself `definition_helper` was more explicit than `schema_helper`, but consistency first and it must be the same as the name chosen above, of course.

I found `definition` could be clearer than `schema` in both places but here again, consistency first, so if we choose OpenAPIv3 terminology, then let's do it completely.

> Since I can't recall the use case for the current response_helper, I'm fine getting rid of it.

Good. Maybe it used to be practical for some use case, but since the plugin classes, it has become awkward to use and it has no benefit over the operation helper.

lafrech:
I think I addressed your concerns.

It should be ready to merge but it could use one last check from another pair of eyes.

CHANGELOG.rst
```diff
@@ -4,6 +4,20 @@ Changelog
...
+  ``Components`` object used to define components such as schemas, parameters
+  or reponses. "Components" is the OpenAPI v3 terminology for those reusable
+  top-level objects.
+- *Backwards-incompatible*: ``apispec.APISpec.definition`` is moved to the
+  ``Components`` class and renamed to ``apispec.core.Components.schema``.
+- ``apispec.core.Components.parameter`` and ``apispec.core.Components.response``
+  are added.
+- *Backwards-incompatible*: ``apispec.APISpec.add_path`` and
+  ``apispec.APISpec.add_tag`` are renamed to ``apispec.APISpec.path`` and
+  ``apispec.APISpec.path``.
```
sloria:
```suggestion
  ``apispec.APISpec.tag``.
```

CHANGELOG.rst
```diff
@@ -4,6 +4,20 @@ Changelog
 1.0.0b5 (unreleased)
 ++++++++++++++++++++
 
+Features:
+
+- ``apispec.core.Components`` is added. Each ``APISpec`` instance has a
+  ``Components`` object used to define components such as schemas, parameters
+  or reponses. "Components" is the OpenAPI v3 terminology for those reusable
+  top-level objects.
+- *Backwards-incompatible*: ``apispec.APISpec.definition`` is moved to the
```
sloria: Not a blocker, but it would be helpful to have a code snippet here comparing "<1.0.0" vs ">=1.0.0b5".

apispec/core.py
```diff
@@ -92,60 +201,40 @@ def __init__(self, title, version, openapi_version, plugins=(), **options):
...
-        kwargs['in'] = location
-        self._parameters[param_id] = kwargs
-
-    def add_tag(self, tag):
+    def tag(self, tag):
         """ Store information about a tag.
 
         :param dict tag: the dictionary storing information about the tag.
         """
         self._tags.append(tag)
```
sloria, line 236: These methods should `return self` in order to support chaining.

apispec/core.py
```diff
@@ -92,60 +201,40 @@ def __init__(self, title, version, openapi_version, plugins=(), **options):
...
 
-    def add_path(self, path=None, operations=None, **kwargs):
+    def path(self, path=None, operations=None, **kwargs):
```
sloria, line 237: These methods should also be made chainable.

docs/quickstart.rst
```diff
@@ -19,22 +19,22 @@ First, create an `APISpec <apispec.APISpec>` object, passing basic information a
         )
     )
 
-Add definitions to your spec using `definition <apispec.APISpec.definition>`.
+Add schemas to your spec using `schema <apispec.core.Components.schema>`.
```
sloria, line 22: The link to `schema` isn't working here because `Components` isn't being autodoc'd.

We're currently generating autodoc for `apispec`, not `apispec.core`. So you can fix this by import `Components` in `apispec/__init__.py` and changing this to

```suggestion
Add schemas to your spec using `schema <apispec.Components.schema>`.
```

tests/schemas.py
```diff
@@ -2,9 +2,10 @@
 
 
 class PetSchema(Schema):
-    description = dict(id='Pet id', name='Pet name')
-    id = fields.Int(dump_only=True, description=description['id'])
-    name = fields.Str(description=description['name'], required=True, deprecated=False, allowEmptyValue=False)
+    description = dict(id='Pet id', name='Pet name', password='Password')
+    id = fields.Int(dump_only=True, description='Pet id')
```
sloria:
```suggestion
    id = fields.Int(dump_only=True, description=description['id'])
```

sloria (commented):
Just a few minor changes. Feel free to merge after those are addressed.

lafrech:
Thanks @sloria for looking into this. I'll apply your suggestions.

I don't get the point about chaining. Do you mean

```py
    spec.tag('Tag 1').tag('Tag 2').tag('Tag 3').path(...)
```

It is unrelated and I'm not convinced it is that useful but it doesn't hurt, I can do that.

sloria:
Chaining certainly isn't necessary, but it can arguably make the code more readable because it allows grouping of the mutations (on either `spec` or `components`) into single statements.

lafrech:
So `schema`, `parameter` and `response` should also return self (= `Components` instance).

It will need tests. May I postpone this to another PR?

sloria:
> It will need tests. May I postpone this to another PR?

Yep, no problem

CHANGELOG.rst
lafrech: Good catch. Fixed.

CHANGELOG.rst
lafrech: Done.

tests/schemas.py
lafrech: Good catch. Fixed.

apispec/core.py
lafrech: Opened https://github.com/marshmallow-code/apispec/issues/335.

docs/quickstart.rst
lafrech: I didn't like the idea of exposing `Components` in __init__.py just for documentation sake, so I added it to the API docs through core.py.

I also removed duplicate `BasePlugin` documentation.

lafrech:
Merged. Thans for double-checking this.

BTW, we'll need to update apispec-webframeworks.

CHANGELOG.rst
```diff
@@ -4,6 +4,37 @@ Changelog
...
+  ``Components`` class and renamed to ``apispec.core.Components.schema``.
+
+::
+
+    # apispec<1.0.0b5
+    spec.add_tag({'name': 'Pet', 'description': 'Operations on pets'})
+    spec.add_path('/pets/', operations=...)
+    spec.definition('Pet')
+    def PetSchema(Schema)
```
sloria, line 27: The `def PetSchema(Schema)` is a bit confusing. Remove?

lafrech: Oops. Right. I got mixed up with flask-rest-api where I use this as a decorator around `Schema` definitions.

PR #310, opened 2018-10-09
Fix #302 Webframeworks split to new package

### LINKED ISSUE #302: Split out web framework plugins into separate package

ergo:


setup.py
```diff
@@ -41,6 +41,7 @@ def read(fname):
     packages=find_packages(exclude=('test*', )),
     package_dir={'apispec': 'apispec'},
     include_package_data=True,
+    install_requires=['apispec-webframeworks'],
```
sloria: We shouldn't require installing this package.

ergo: I wanted to preserve backwards compatibility, but if we don't want that it can be removed.

sloria: If we're preserving compatibility, then there should be a DeprecationWarning raised when importing from `apispec.ext`.

I don't think it's strictly necessary, though. Unless @lafrech disagrees, I think we can just document the migration path to using apispec-webframeworks and break backwards compat.

sloria:
@ergo OK, apispec-webframeworks is now published to PyPI. Can you add it as a dev dependency here and remove the backwards compatibility code?

setup.py
lafrech: Absolutely. This is what I wrote in https://github.com/marshmallow-code/apispec/issues/302.

Drop compatibility and document the change.

ergo:
Done, two entries on travis do not pass, but those do not seem related to my changes.

sloria:
@ergo The build failures appear to be due to how you're testing against apispec-webframeworks. 

Should it be

```python
    webframeworks_test_dir = pkg_resources.resource_filename(
        'apispec_webframeworks'
    )
```

instead?

ergo:
Hm, we can try some other solution, how did you come to conclusion it is `resource_filename` issue?
The testrunner output is a bit cryptic for me in this scenario.

sloria:
@ergo From the traceback, it appears that pytest is trying to run tests in `site-packages/tests`.

ergo:
Interesting, seems to happen only on 3.5 and I can't replicate locally, I will try to come up with a solution.
It seems that the currently released packages for `apispec-webframeworks` do not contain tests directory in the first place, since tests there do not contain `__init__` they need to explicitly included to the egg.

ergo:
All tests pass, I don't have a good idea how to solve the problem of testing `apispec-webhelpers` inside tox - there are some tests via `test_ext_combination` but I'm thinking about tests from webhelpers package directly.

That would require distributing tests with pypi package (I know how to do this for source package, but not for wheel) and then doing some directory sniffing with `pkg_resources`. 

Right now the easy way to test both packages in a single run is via `invoke tests` and that works just fine in my venv when `apispec-webhelpers` is installed via `pip install -e .` .

sloria:
How about inlining the tests, as documented here? https://docs.pytest.org/en/latest/goodpractices.html#tests-as-part-of-application-code

That way you'd just need to pass `--pyargs apispec_webframeworks`.

ergo:
OK, I've pushed to both repos, with new release of `apispec_webframeworks` all tests should pass, 
Invoke tests works for me and correctly imports tests from external package.

tasks.py
```diff
@@ -22,6 +22,11 @@ def test(ctx, watch=False, last_failing=False):
     if last_failing:
         args.append('--lf')
     args.append('tests')
+    try:
```
sloria: This try-except should not be necessary, since we're installing apispec-webframeworks in dev requirements.

apispec/ext/__init__.py
```diff
@@ -0,0 +1 @@
+from __future__ import absolute_import
```
sloria: This can be removed.

sloria (changes requested):
Just a few drive-by comments.

sloria:
It's a bit suspicious that tests are passing without apispec-webframeworks having been release yet. @ergo Are you sure that the tests are running?

sloria:
Oh, I see, it's probably trying to run tests in the `apispec_webframeworks` package but it's not finding any. Disregard previous comment.

lafrech:
Do we really need to run webframeworks tests from apispec?

It's nice to be able to run them using invoke if we want to, but I'm not sure we need to do that in the CI tests.

If apispec introduces breaking changes, it's nice to know that webframeworks are broken, but it shouldn't block apispec from releasing. webframeworks will have to adapt and release a fix that can only be released  if the tests pass, which may require that apispec's new versions is released. If the CI prevents apispec from releasing due to the breaking changes, we're facing a chicken and egg problem, aren't we?

When marshmallow introduces a breaking change, apispec detects it as soon as its CI tests are run (which incidentally is pretty soon due to dependabot).

Or am I totally mistaken?

ergo:
@lafrech turning on/off of this behavior is matter of configuration, so we have two options:
- do not install webframeworks by default as requirements
- make an env var for turning off the tests

Though right now we still have "combined" plugin test in apispec that requires both packages to be present.

tasks.py
ergo: fixed

apispec/ext/__init__.py
ergo: fixed

PR #322, opened 2018-10-24
Fix flake8 issues

lafrech:
There were new flake8 issues.

This PR fixes E731 and configures flake8 to ignore W504.

Regarding W503 / W504, it is still a bit blurry to me but in [this bug report](https://gitlab.com/pycqa/flake8/issues/463), @asottile says pycodestyle now reports both. Since they are incompatible, I think it is up to us to choose our convention.

---------------------------------------------------------------------------------------------------------------------------------

And now we have another CI failure because flake8 did not exclude the docs directory.

I don't get it. Here's the config.

```
# E127: continuation line over-indented for visual indent
# E128: continuation line under-indented for visual indent
# E265: block comment should start with #
# E302: expected 2 blank lines, found 0
# E266: too many leading ‘#’ for block comment
# W504: line break after binary operator
[flake8]
ignore = E127,E128,E265,E302,E266,W504
max-line-length = 110
exclude = .git,docs,env,venv,.ropeproject,build,.tox
```

I can't even reproduce the failure on my local setup.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -8,7 +8,9 @@ universal=1
 # E128: continuation line under-indented for visual indent
 # E265: block comment should start with #
 # E302: expected 2 blank lines, found 0
+# E266: too many leading ‘#’ for block comment
+# W504: line break after binary operator
 [flake8]
-ignore = E127,E128,E265,E302,E266
+ignore = E127,E128,E265,E302,E266,W504
 max-line-length = 110
-exclude=.git,docs,env,venv,.ropeproject,build,.tox
+exclude = .git,docs,env,venv,.ropeproject,build,.tox
--- a/setup.cfg
+++ b/setup.cfg
@@ -13,4 +13,3 @@ universal=1
 [flake8]
 ignore = E127,E128,E265,E302,E266,W504
 max-line-length = 110
-exclude = .git,docs,env,venv,.ropeproject,build,.tox
--- a/apispec/ext/marshmallow/openapi.py
+++ b/apispec/ext/marshmallow/openapi.py
@@ -603,10 +603,12 @@ class OpenAPIConverter(object):
             ):
                 continue
 
+            def prop_func(field_obj=field_obj):
+                return self.field2property(
+                    field_obj, use_refs=use_refs, dump=dump, load=load, name=name,
+                )
+
             observed_field_name = self._observed_name(field_obj, field_name)
-            prop_func = lambda field_obj=field_obj: self.field2property(  # flake8: noqa
-                field_obj, use_refs=use_refs, dump=dump, load=load, name=name,
-            )
             jsonschema['properties'][observed_field_name] = prop_func
 
             partial = getattr(schema, 'partial', None)
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -6,7 +6,7 @@ import os
 import sys
 
 sys.path.insert(0, os.path.abspath('..'))
-import apispec  # flake8: noqa
+import apispec  # noqa: E402
 
 extensions = [
     'sphinx.ext.autodoc',


setup.cfg
```diff
@@ -8,7 +8,9 @@ universal=1
 # E128: continuation line under-indented for visual indent
 # E265: block comment should start with #
 # E302: expected 2 blank lines, found 0
+# E266: too many leading ‘#’ for block comment
+# W504: line break after binary operator
 [flake8]
-ignore = E127,E128,E265,E302,E266
+ignore = E127,E128,E265,E302,E266,W504
```
asottile, line 14: fwiw, if you're happy with the default `ignore` list:

>                         (Default: E121,E123,E126,E226,E24,E704,W503,W504)

you can use `extend-ignore = ...` to add to it instead of overriding it with `ignore =`

lafrech: Thanks @asottile for the tip. I'll defer code style decisions to @sloria.

asottile:
as for the docs thing, that's a bit of an odd interaction that I've seen before.

tl;dr, if a filename is passed to `flake8`, `flake8` (intentionally) does not apply the `exclude` list to it.  What's happening is `pre-commit` is finding all python files and passing them on to `flake8`.  The easiest fix is to add `exclude: ^docs/` to https://github.com/marshmallow-code/apispec/blob/bd95afbafd30cb6911c582f6fcea7a97af5974ae/.pre-commit-config.yaml#L12

asottile:
Particularly https://github.com/pre-commit/pre-commit-hooks/issues/112 https://gitlab.com/pycqa/flake8/issues/10

.pre-commit-config.yaml
```diff
@@ -10,3 +10,4 @@ repos:
   - id: double-quote-string-fixer
   - id: end-of-file-fixer
   - id: flake8
+    exclude: .git,docs,env,venv,.ropeproject,build,.tox
```
asottile: pre-commit will only ever run on checked in files, and takes a [regular expression](https://pre-commit.com/#regular-expressions) you want `exclude: ^docs/`

lafrech:
I copied the whole line like this and the tests still fail:

```
  - id: flake8
    exclude: .git,docs,env,venv,.ropeproject,build,.tox
```

Is it the wrong syntax?

They pass on my machine, even the pre-commit hook passed when I committed this.

asottile:
`pre-commit` will only run on the files you're comitting.  If you want to try it against all the files I'd suggest `pre-commit run --all-files`

setup.cfg
sloria: Oh, interesting. Yeah, `extend-ignore` is probably the way to go. @lafrech can you make that change?

sloria:
@lafrech For the docs issue, I think you can just add `# noqa: E402` to that one import line.

sloria:
> They pass on my machine, even the pre-commit hook passed when I committed this.

@lafrech You're probably using an old flake8 version in the cached pre-commit environment. I ran into this this morning. Try running

```
pre-commit clean
pre-commit run --all-files
```

That will clear the cache and re-install the environment.

lafrech:
Alright, I added `# noqa: E402` to the offending line in docs and removed the flake8 exclude line. IIUC, flake8 only works on python files anyway so much of the excludes were useless.

@sloria you were right about pre-commit cache.

Regarding `extend-ignore`, I don't know. The default list is almost orthogonal to ours, so this would ignore checks that we currently pass, while we'd still need to specify almost all of those we currently ignore.

Let's merge this already. We can change `extend-ignore` later on if we feel the need.

lafrech:
Thanks @asottile for the support!

asottile:
No problem! glad to help

PR #323, opened 2018-10-24
Split webframeworks into a separate package

lafrech:
Built upon @ergo's https://github.com/marshmallow-code/apispec/pull/310.

I rebased and removed all tests involving webframeworks. Rationale in https://github.com/marshmallow-code/apispec/pull/310#issuecomment-432550087.

I think the "combined" test is covered indirectly by tests in test_core.py. It could be moved to webframeworks if we want to test specifically the web frameworks actions in it.

Then I figured let's add a flag to invoke test to test the webframeworks (but don't use it in CI, see rationale above).

Web frameworks dependencies are added in dev-requirements.txt.

Web frameworks tests are run using

    invoke test --webframeworks


sloria:
> If apispec introduces breaking changes, it's nice to know that webframeworks are broken, but it shouldn't block apispec from releasing. 

Fair point. Perhaps we can have a Travis job that runs the apispec-webframeworks test suite but is allowed to fail: https://docs.travis-ci.com/user/customizing-the-build/#rows-that-are-allowed-to-fail .

sloria:
This is fine to merge as is. I'm planning to rework the dev environment the same way as I did with marshmallow (with tox, etc.). I'll see if I can make my above proposal happen.

sloria:
This still needs docs updates. Opened a new issue: #324 . @lafrech Do you have time to do this?

sloria:
Actually, I have some time to update the docs. I'll take care of it.

Issue #307, opened 2018-10-08
RFC: remove schema auto-reference feature from MarshmallowPlugin


lafrech:
See [comment](https://github.com/marshmallow-code/apispec/pull/179#issuecomment-405185455).

Rationale:

This is something that is easy to do in a custom plugin. Just override `definition_helper`. But it is hard to do in a totally generic way.

To make it more generic, the authors added a `schema_name_resolver` callback, so that the user can provide a function that returns a name from a Schema (for instance, it can be the schema name with Schema suffix removed, and lowercased: `PetSchema` -> `pet`).

The `inspect_schema_for_auto_referencing` method hardcodes the field types that have containers. Currently, only `Nested` and `List`. `Dict` is missing for Marshmallow 3 (this could be easily fixed). And there is no mechanism to allow custom container fields. We could add another parametrization mechanism for that, but it looks like overparametrization. I mean from a user perspective, it looks just as easy to reimplement the whole feature (the recipe can be provided in the docs) than to awkwardly parametrize the existing feature. OTOH, I realize the custom field with container is a corner case.

The PR I linked to above (#179) adds more complexity to fix issues in the feature.

I'm thinking it might be better to drop it. Maybe we could add instructions in the docs to show how to achieve it in a custom plugin. Of course, if the original authors manage to do what they need in a plugin and publish this plugin, I'd be happy to link to it. The point is not to trash other people's work (NIH syndrome), but to make it the most practical both from user and maintainer perspective.

Maybe I'm overthinking it. It's mostly a matter of opinion. I'm open to discussion about it. Since it is a breaking change, I'd like to have it settled for 1.0.

@inkhey, @lebouquetin and @buxx, since you added the feature, I'd be happy to get your feedback. I think you can easily fit this in hapic. It might be even easier for you if you want to modify it. You tell me.

Note: If we keep it, let's at least manage the `Dict` case for marshmallow 3.

Bangertm:
I like the concept behind the auto-reference feature: add a single schema to the spec and all sub schemas get added and referenced.  This keeps the resulting yaml relatively flat and give names to nested schemas instead of nesting them as an anonymous object. 

[...]

Issue #172, opened 2017-12-01
Support for Hyperlinks schema field type (provided by flask-marshmallow)


Ch00k:
Currently if I have a `Hyperlinks` field type in my schema, it is being interpreted as `string`, and of course all the complex nested structure of the field is not preserved. Is there any way I can make `Hyperlinks` field type work with apispec?
https://flask-marshmallow.readthedocs.io/en/latest/#flask_marshmallow.fields.Hyperlinks

Ch00k:
Just realised that the same happens to a `Function` field.

lafrech:
This happens in `OpenAPIConverter.field2property`.

There is currently no way to extend this method.

As a workaround, you could make your own `OpenAPIConverter` child class and override `field2property`, then override `MarshmallowPlugin` to use it. This implies a lot of duplication in `field2property`, so it is far from perfect.

A better solution would be to let the user specify `field2property` methods to map to field types, assuming that's even feasible.

We could keep a `FIELD_TO_PROPERTY` mapping:

```py

FIELD_TO_PROPERTY = {
    flask_marshmallow.fields.Hyperlinks: hyperlink_field2property,
}
```
Each methods in the mapping would return a dict.

Then in `OpenAPIConverter.field2property`

```py
        [...]
        elif isinstance(field, marshmallow.fields.Dict):
            [...]
        # Manage custom fields
        elif type(field) in FIELD_TO_PROPERTY:
            ret.update(FIELD_TO_PROPERTY[type(field)](field, use_refs=use_refs, dump=dump, name=name))
        [...]
```

Then, you'd "just" have to create `hyperlink_field2property` so that it returns what you expect.

By definition, `Function` fields can return anything. What would you expect apispec to do?

Issue #169, opened 2017-11-11
Marshmallow schema parameters are resolved using schema class, not schema instance


dlopuch:
When apispec resolves parameters using marshmallow schemas, it appears to use the Schema *class*, not the Schema *instance*:

https://github.com/marshmallow-code/apispec/blob/dev/apispec/ext/marshmallow/__init__.py#L146

I believe this is problematic because Marshmallow allows for schema modifiers on instantiation, but these modifiers are therefore ignored by apispec.  This is important because different uses of a schema may require different instantiations.

For example, say you define a schema like so:

```python
class SomeEntitySchema(Schema):
    id = fields.Integer(required=True)
    name = fields.String(required=True)
```

On a get-item, either the class or the generic instantiation `SomeEntitySchema()` works.  

However, on a post, you would use it as a body param like `SomeEntitySchema(exclude=('id',)` -- ie when you're creating a new entity, you don't pass in the ID (this would be assigned by the server).

So the bug is since apispec uses the *class* of the schema not the *instance* in the parameter case, instance-specific modifiers are ignored.  The apispec tells you that `id` is required on a post, which you specifically said it's not.

Any insight into why https://github.com/marshmallow-code/apispec/blob/dev/apispec/ext/marshmallow/__init__.py#L146 looks up the class instead of using the instance?

inkhey:
Just trying something to support exclude/only for auto_ref : https://github.com/inkhey/apispec/commit/e584f26433bcc388267a71292c61add7ee94ae76 .

lafrech:
@dlopuch, for this specific use case, I just mark `id` as `dump_only` in the schema and I can use the same schema class in GET, POST and PUT.

lafrech:
This is a complicated design choice. Indeed it can be misleading to accept instances but resolve on the class. Maybe resolving on instances wouldn't make things worse.

The issue with resolving by instance is that if you use a schema with the same modifiers in multiple places, it won't be the same instance (unless you instantiate it once and re-use the instance) so the definition resolving won't work. But it is not as bad as current situation.

I don't see any obvious blocking issue with resolving by instance. Doesn't mean there is none.

lafrech:
Resolving on class is wrong because it ignores instance modifiers.

Resolving on instances means one would need to register an instance and then reuse that instance in its views. I don't see a elegant way of doing this. We may end up with code like this:

```py
pet_load_schema = PetSchema(exclude=('id',))

[...]

PR #144, opened 2017-08-15
Added use_instances parameter to Marshmallow plugin

theirix:
It allows to support partial Marshmallow schemas because partial fields can be specified only for Marshmallow instance.

It is not user-facing change but useful for writing plugins.

PR #333, opened 2018-10-31
Nested schema auto resolution as default and respect modifiers in references

Bangertm:
Relates to #307 

This is a proof of concept on combining the logic in inspect_schema_for_auto_referencing with the nesting logic in field2property.  

1. Default schema_name_resolver now passed to MarshmallowPlugin
2. Logic for adding nested schmeas moved from
3. MarshmallowPlugin.inspect_schema_for_auto_referencing to
4. OpenAPIConverter.field2property
5. Removed using LazyDict in favor of regular Dict

All the tests are passing, but I presume that there are some minor differences in the specs that are not captured by the test.  Of potential concern would be cases like [this](https://github.com/marshmallow-code/apispec/blob/b0e1b214a0c295bcf2205cc0995c45a400f4f7e7/tests/test_ext_marshmallow.py#L576).  The second two definition adds would be unnecessary and if the schema name resolved to something other than what the user is putting in as the name there will be extra components in the model.  Not sure that is too big of a deal or not.

At this point the cyclomatic complexity of field2property has increased, but there should be some simplifying that can be done.  For example I'm not sure that the unbound self referencing cases are valid and the concept of embedding the reference string in field metadata feels like a relic of not having a ref dictionary.


apispec/ext/marshmallow/__init__.py
```diff
@@ -38,6 +38,10 @@ class UserSchema(Schema):
 from .openapi import OpenAPIConverter
 
 
+def resolver(schema):
+    return schema.__name__.replace('Schema', '')
+
```
lafrech, line 49: Or

```py
    return schema.__name__.rstrip('Schema') or schema.__name__
```

`PetSchemaSchema` -> `PetSchema`
`SchemaSchema` -> `Schema`
`Schema` -> `Schema`

There would still be a collision between `SchemaSchema` and `Schema` but admittedly it doesn't matter.

apispec/ext/marshmallow/__init__.py
```diff
@@ -50,7 +54,7 @@ class MarshmallowPlugin(BasePlugin):
             def schema_name_resolver(schema):
                 return schema.__name__
     """
-    def __init__(self, schema_name_resolver=None):
+    def __init__(self, schema_name_resolver=resolver):
         super(MarshmallowPlugin, self).__init__()
         self.schema_name_resolver = schema_name_resolver
```
lafrech: I think this would make it easier to subclass:

```py
    def __init__(self, schema_name_resolver=None):
        super(MarshmallowPlugin, self).__init__()
        self.schema_name_resolver = or resolver
```

No need to import resolver or change the signature.

apispec/ext/marshmallow/openapi.py
```diff
@@ -362,12 +364,19 @@ def field2property(self, field, use_refs=True, dump=True, load=True, name=None):
...
                 schema_dict = self.resolve_schema_dict(field.schema, dump=dump, load=load)
                 if ret and '$ref' in schema_dict:
                     ret.update({'allOf': [schema_dict]})
                 else:
                     ret.update(schema_dict)
         elif isinstance(field, marshmallow.fields.List):
+            if isinstance(field.container, marshmallow.fields.Nested):
+                nested_schema_class = resolve_schema_cls(field.container.schema)
+                if nested_schema_class and nested_schema_class not in self.refs:
+                    self.add_nested_schema_to_spec(nested_schema_class)
```
lafrech: The MA3 `Dict` case needs to be addressed, too.

lafrech (commented):
This has the same (mostly theoretical) limitation as the current implementation: the container fields (`Nested`, `List(Nested)`, `Dict(Nested)`) are hardcoded, it won't work with a custom container field.

But your proposal here reminds me that the whole `field2property` logic hardcodes the container fields. This is nothing new. In [this comment](https://github.com/marshmallow-code/apispec/issues/172#issuecomment-415065055) I propose a hook to manage those cases. Perhaps could it be extended to manage the auto-ref logic.

In any case, after a quick look, this implementation looks better than the existing IMHO. This looks like the right place for the job. And the hook I mention in above paragraph could cover the custom container field corner case for both documentation and auto-ref.

apispec/ext/marshmallow/__init__.py
lafrech: My first thought was `schema.__name__.replace('Schema', '', 1)` but this will replace by the left.

Generally, schemas are named `SomethingSchema` so it should be by the right.

Say you have a file for pet schemas, the schema for this file is `PetSchemaFileSchema`.

More likely, you want to serialize schematics, the schema is `SchematicsSchema` so we'd better strip right.

If we make this default, we should add a not in the docs to tell users to name their schemas `SomethingSchema` (or specify another resolver).

Bangertm: rstrip keeps on replacing characters until it finds on not in `chars` so SchemaSchema -> ''.  Changed to only replace 'Schema' once.

lafrech: Oh, right. I was wrong about `rstrip`.

But `schema.__name__.replace('Schema', '', 1) or schema.__name__` fails on the `SchematicsSchema` case.

    'SchematicsSchema'.replace('Schema', '', 1)
    'ticsSchema'

We might have to go with a longer form:

```py
def resolve(schema):
    return schema.__name__[:-6] or schema.__name__ if schema.__name__.endswitdh('Schema') else schema.__name__
```

lafrech:
I think your recent commits make my comment above irrelevant (unless it was irrelevant from the beginning), as the resolving happens in a single place (https://github.com/marshmallow-code/apispec/pull/333/files#diff-daf6a711ac369a8d051b9606e0a9eca6R367) and there is no need to add specific cases for each container field. This is great.

apispec/ext/marshmallow/openapi.py
```diff
@@ -603,13 +621,11 @@ class Meta:
...
-                return self.field2property(
-                    field_obj, use_refs=use_refs, dump=dump, load=load, name=name,
-                )
-
             observed_field_name = self._observed_name(field_obj, field_name)
-            jsonschema['properties'][observed_field_name] = prop_func
+            property = self.field2property(
+                field_obj, use_refs=use_refs, dump=dump, load=load, name=name,
+            )
+            jsonschema['properties'][observed_field_name] = property
```
lafrech, line 643: AFAIU, this change and the change above (`OrderedLazyDict()` -> `OrderedDict()`) are related to each other but unrelated to the issue, right?

Last time I checked, I didn't understand this part of the code, so I appreciate that you could make it simpler but I can't comment on the impact.

Bangertm: No these changes are necessary if we want to add a schema to the spec from within `field2property`.  The way it worked before is that for each field a `prop_func` was created with a default parameter of the `field_obj`.  Then that `prop_func` is loaded into a `LazyDict`.  A the end of `fields2jsonschema` the `LazyDict` is returned but _none of the `prop_func`s are called_.  They only get called when accessing the item in the dict.  

This strategy was added [here](https://github.com/marshmallow-code/apispec/commit/fa1ec525fcd1d840e5d1e955ab8cfdfc11367943).  With the plan being that if execution of `field2property` was delayed until all of the schemas are in the spec (and in the ref dictionary) then when `field2property` eventually gets called it will find the the nested schemas in the ref, insert a ref into the spec, and stop infinite recursion.

Now if we want to move the functionaity to add a nested schema to the spec in `field2property` there is a problem with this approach.  The first time iterating over the spec's definitions all of the all of the `prop_func`s in the spec get called for the first time and `field2property` tires to add schemas to the dictionary during iteration which throws an error.  For some reason things like `print spec.to_dict()` will work but `json.dumps(spec.to_dict())` will not.  

I think the original authors of the autoreferencing functionality were running into this, which is why they put they logic to iterate over the fields and add nested schemas to the spec where they did.  In `definition_helper` `inspect_schema_for_autoreferencing` was adding all of the nested schemas into the spec before any of the `field2property` functions are called.  

The plan in this pr is to use regular dictionaries and call `field2property` immediately from `fields2jsonschema`.  All nested schemas will get added to the spec along the way via a call to `spec.definition` from `field2property` - instead of manually in the original plan or via `inspect_schema_for_autoreferencing`.  

I hope this clarifies things, but I fear it might not...

Bangertm (commented):
> I think your recent commits make my comment above irrelevant (unless it was irrelevant from the beginning), as the resolving happens in a single place (https://github.com/marshmallow-code/apispec/pull/333/files#diff-daf6a711ac369a8d051b9606e0a9eca6R367) and there is no need to add specific cases for each container field. This is great.

I'm glad you pointed it out for the Dict case because I didn't realize that I had redundant logic in the List case.  Can you take a look a the two test cases added to make sure they cover the appropriate case.

apispec/ext/marshmallow/__init__.py
Bangertm: Should be good now.

lafrech:
Current implementation of `spec.components.schemas` does not detect if a name is used twice. The second schema overrides the first. We should probably raise an Exception when trying to register a second schema with an already registered name. While this is not critical when registration is manual, it can be useful when registration is automatic as the resolving can lead to tricky cases.

I think we both have the same nominal use case in mind. All schemas are called `EntitySchema` and are registered as `Entity` in the spec. But there can be collisions if user doesn't follow this naming scheme. Like `PetSchema` and `Pet`. And letting the user specify another resolver opens the door to all sorts of tricky issues.

I don't think this should be addressed here, but it pushes for a simple check-raise in spec.components.schema.

I opened #340 about this.

apispec/ext/marshmallow/openapi.py
```diff
@@ -392,6 +393,34 @@ def field2property(self, field, use_refs=True, dump=True, load=True, name=None):
...
+        schema_cls = resolve_schema_cls(schema)
+        name = self.schema_name_resolver(schema_cls)
+
+        if name and schema_cls not in self.refs:
+            self.spec.components.schema(
+                name,
+                schema=schema,
+                dump=dump,
+                load=load,
+            )
```
lafrech, line 429: I don't think we should pass dump/load here. Those are meant to remove some fields when registering a parameter schema or a response schema, but not when registering a definition.

When registering a schema definition, it contains all fields, and dump_only / load_only attributes are documented, which should be enough.

Assuming a schema is used both for parameter and response, it could be auto-registered in dump or load mode depending on who registers it first, which is wrong.

I'd remove `dump` and `load`, and revert the corresponding change in `schema_helper` above.

apispec/ext/marshmallow/__init__.py
```diff
@@ -168,12 +154,12 @@ def schema_helper(self, name, schema, **kwargs):
...
-
-        # Auto reference schema if schema_name_resolver
-        if self.schema_name_resolver:
-            self.inspect_schema_for_auto_referencing(schema_instance)
+        kwargs.pop('definition', None)
+        json_schema = self.openapi.schema2jsonschema(
+            schema_instance,
+            name=name,
+            **kwargs
+        )
```
lafrech: Not sure about passing `**kwargs`, here. See below.

lafrech:
On second thought, the auto-ref feature could be a reason not to raise an Exception, because it might auto-register a Schema that was already registered manually...

We could add a test to check if the already registered schema with the same name is the same schema class, but then it becomes a bit smelly.

lafrech:
Do we want to force the feature or just make it default?

In the current state of the PR, the only way to cancel the feature is to pass a no-op resolver to override the default. This is due to my change request of not passing the resolver default in the signature. I realized that afterwards. We could revert that change. It is not a friendly API.

We can go the other way and decide the feature is mandatory. This does not seem unreasonable. After all, when viewing the spec with ReDoc (and perhaps swagger-ui, I don't remember), it is done in the viewer frontend.

If we do that, we should provide a default name for when the schema name can't be resolved (when the resolver returns None). It can be `'Schema_{}'.format(some_incremented_counter)`.

Then we may write:

```py
        if schema_cls not in self.refs:
            self.spec.components.schema(
                self.thin_layer_above_schema_name_resolver_providing_default_name(schema_cls),
                schema=schema,
                dump=dump,
                load=load,
            )
```

Or we don't try to think too much and we leave it as it is.

I think this thought was triggered by the self-nesting or circular nesting case. I still don't totally grasp the change you made w.r.t. the delayed resolution (`LazyDict`). It seems too nice to be true. I was wondering if it would work if the resolver returned `None`. Wouldn't we fall back on the old issue (before #44)? Making sure all schemas are registered, even with a dummy name, sounds like a simple way around this. Sorry if I'm still missing something.

Don't miss the "good" in "too good to be true". I like the code simplification brought by this PR.

Bangertm:
I'm still trying to wrap my head around the implications of making this feature the default.  There are just so many assumptions that have been made along the way.  

Regarding collisions, my biggest concern is that a nested schema comes in once with `dump=False` and then gets referenced someplace where `dump=True`. That schema second reference will be missing fields.  

Part of the rational for excluding fields when `dump` or `load` is `False` is that the schema's parameters are listed in line, but if we have the general plan of using the `schema_name_resolver` to add the schema to the doc and insert a reference it might negate the need to have `dump` and `load` functionality at all.  

For example we could have a check early in `resolve_schema_dict` to use add schemas not currently in the spec using the resolved name and insert a ref instead of inlining the parameters.  

>  I was wondering if it would work if the resolver returned None. Wouldn't we fall back on the old issue (before #44)?

With the current implementation there is a check on whether `schema_name_resolver` returned a value and if not then the schema is not added to the spec.  For circular references, this will revert the the problem before #44.  But we could also remove the check.  But that would also remove a convenient way for users to opt out of this functionality.

lafrech:
> Regarding collisions, my biggest concern is that a nested schema comes in once with dump=False and then gets referenced someplace where dump=True. That schema second reference will be missing fields.

Exactly.

> Part of the rational for excluding fields when dump or load is False is that the schema's parameters are listed in line, but if we have the general plan of using the schema_name_resolver to add the schema to the doc and insert a reference it might negate the need to have dump and load functionality at all.

Yeah, what I thought, too. We should double-check that. I wouldn't mind dropping the dump/load logic since schema used as reference should have the dumpOnly/loadOnly attributes correctly set.

> For circular references, this will revert the the problem before #44.

OK, I thought so. We can't afford to break the circular reference use case so it looks like we're in the middle of the bridge: either we keep things as they are, either we make the auto-ref mandatory.

Keeping things as they are means back to https://github.com/marshmallow-code/apispec/issues/307. Either keep, remove or deprecate current implementation (which can be easily introduced by subclassing MarshmallowPlugin).

If we choose to make the feature mandatory, we need to sort out a few things:
- ensure the name defaults to something unique
- think of a nice way to manage name collisions (is an exception acceptable?)
- remove dump/load logic if it has become useless

I wrote:

> On second thought, the auto-ref feature could be a reason not to raise an Exception, because it might auto-register a Schema that was already registered manually...

This shouldn't be an issue since we check if the schema is already registered. The issue is when registering manually a schema that has been auto-registered before. We could skip that second registration, but that means the auto-name will be used instead of the user-specified name. Doing this silently sounds wrong, so an exception might be acceptable. If the user wants to specify the name manually, he should register the schema before using it.

This logic could be a blocker for users wanting to specify names manually in case of circular references. 

And if not a blocker, it could be an annoyance for users wanting to register manual names as they'd need to ensure they register the schemas in the right order. In practice, I'm not sure it is such an issue. I think I already do that in my code.

We may postpone this to 2.0 if we can't sort it out easily. We'd still need to decide what to do in 1.0.

Bangertm:
Given that this pr is somewhat experimental I added a couple of commits to move towards always adding the schema to the spec and inserting a reference whenever a Marshmallow schema is encountered in a request or response body.

I'm starting to think that this might generally be the most desirable out of the box behavior.  All of the schemas have names in the user's program.  Part of the reason for creating a openapi spec is to communicate the structure and the names of the objects to the user.  Auto-resolving the names makes this easier.  In fact we may want to remove the `name` parameter from `Components.schema()` and just resolve that too...

> OK, I thought so. We can't afford to break the circular reference use case so it looks like we're in the middle of the bridge: either we keep things as they are, either we make the auto-ref mandatory.

The only way to break the circular reference case is to override the default `schema_name__resolver` with a function that returned `None` with one of the users' schemas.  Given that the default resolver never returns `None` I'm not sure why anyone would want to create a function that didn't work.  If they really wanted to have their own custom names for circular referencing that were different from the names of the schemas in their program, it's easy enough to setup a function that returns a the custom name based on the schema class.  

I added code that warns the user if they attempt to add a schema that is already in the spec.  I'm not sure we need an exception, but that could easily be changed.  I went with a warning for now because if the users manually adds a schema that has already been added to the spec (automatically) the resulting spec will still be valid.  It just may not have the names that they were expecting in the ref and may have an "unused" schema.  

If we remove the dump/load logic from `schema2jsonschema` there should be no issues with fields being missing.  I'll look at tackling that next.

lafrech:
> In fact we may want to remove the `name` parameter from `Components.schema()` and just resolve that too...

This method can be used without `MarshmallowPlugin`.

> The only way to break the circular reference case is to override the default `schema_name__resolver` with a function that returned None with one of the users' schemas. Given that the default resolver never returns `None` I'm not sure why anyone would want to create a function that didn't work.

So should we specify that custom resolver MUST return a `string` and not handle the `None` case? Currently, the `None` case is silently handled (https://github.com/marshmallow-code/apispec/pull/333/files#diff-daf6a711ac369a8d051b9606e0a9eca6R409) but it will break in case of circular reference resulting in hard to debug errors.

Bangertm:
After thinking about this for a few days I think this is in a pretty good state.

Out of the box: 
- Whenever a nested schema is encountered it is added to the spec and referenced
- Whenever a schema is found in request body or in a response body it is added to the spec and referenced (this eliminates the need for `dump` and `load`
- Circular referencing schemas are automatically resolved - users who are upgrading will get a warning that they are entering a schema twice (once automatically and then again manually) and will likely want to make some changes to their code to remove the manual entry of the schema and to change the schema's name if they don't like the auto one

With a custom `schema_name_resolver` function:
- Users can provide whatever names they want - easiest would be with a dictionary mapping schema class to a name string
- If users really want to have nested schema properties listed inline instead of referenced they can have their resolver return `None` for a given schema
- If the custom name resolver returns `None` for a schema in a circular referencing chain they get a helpful exception telling them not to do that

I think this seems like a decent mix of desired out of the box behavior with enough customization for users who want that.  

The nice thing about this from my perspective is that out of the box every schema in the spec will have a name that matches what I call the schema in my code.  I don't have to make sure to add nested schemas to the spec before adding the schemas that nest them and I don't have to provide a name resolver.  It also makes the resolving schemas from a doc string much more convenient because those schemas get automatically named as well.  I can probably eliminate adding schemas manually altogether can just add the paths.

lafrech:
Great!

I still need to review the code changes more thoroughly but I agree on the principle.

> Users can provide whatever names they want - easiest would be with a dictionary mapping schema class to a name string

Yeah, or an access to a class attribute.

```py
class MySchema(Schema):
    schema_name = 'MySuperSchema'
    ...

def resolver(schema):
    return schema.schema_name
```

apispec/ext/marshmallow/__init__.py
```diff
@@ -38,12 +39,22 @@ class UserSchema(Schema):
...
+
+
 class MarshmallowPlugin(BasePlugin):
     """APISpec plugin handling marshmallow schemas
 
     :param callable schema_name_resolver: Callable to generate the schema definition name.
         Receives the `Schema` class and returns the name to be used in refs within
-        the generated spec.
+        the generated spec. When working with circular referencing this function
+        must must not return `None` for schemas in a cirular refeerence chain.
```
lafrech: s/refeerence/reference

apispec/ext/marshmallow/__init__.py
```diff
@@ -166,14 +152,13 @@ def schema_helper(self, name, schema, **kwargs):
...
+        self.warn_if_schema_already_in_spec(schema_cls)
         # Store registered refs, keyed by Schema class
         self.openapi.refs[schema_cls] = name
-
-        json_schema = self.openapi.schema2jsonschema(schema_instance, name=name)
-
-        # Auto reference schema if schema_name_resolver
-        if self.schema_name_resolver:
-            self.inspect_schema_for_auto_referencing(schema_instance)
+        json_schema = self.openapi.schema2jsonschema(
```
lafrech: Fits in one line? (Its a detail, but it'd simplify the diff.)

apispec/ext/marshmallow/openapi.py
```diff
@@ -392,6 +390,45 @@ def field2property(self, field, use_refs=True, dump=True, load=True, name=None):
...
+            try:
+                return self.schema2jsonschema(schema)
+            except RuntimeError:
+                raise APISpecError(
+                    '{resolver} returned None for {schema} which is part of a'
+                    ' chain of circular referencing schemas. Please ensure'
+                    ' that the schema_name_resolver passed to'
+                    ' MarshmallowPlugin returns a string for all circular'
+                    ' referencing schemas.'.format(
+                        resolver=self.schema_name_resolver,
```
lafrech: Do we need to print the function in the error message?

What about

> Name resolver returned None for schema {schema}...

tests/test_ext_marshmallow.py
```diff
@@ -122,30 +123,14 @@ def resolver(schema):
...
-        definitions = get_definitions(spec)
-        assert 1 == len(definitions)
-        assert 'analysis' in definitions
-        # Inspect/Read objects will not auto reference because resolver func
-        # return None
-        json.dumps(spec.to_dict())
-        # Other shema still not referenced
-        definitions = get_definitions(spec)
-        assert 1 == len(definitions)
+        with pytest.raises(APISpecError):
```
lafrech: Maybe match the error message

```py
    with pytest.raises(APISpecError, match='Name resolver returned None for schema'):
```

tests/test_ext_marshmallow.py
```diff
@@ -122,30 +123,14 @@ def resolver(schema):
...
+            spec.components.schema('analysis', schema=schema)
+
+    @pytest.mark.parametrize('schema', [AnalysisSchema, AnalysisSchema()])
+    def test_warning_when_schema_added_twice(self, spec, schema):
+        spec.components.schema('Analysis', schema=schema)
+        with pytest.warns(UserWarning):
```
lafrech: Likewise, you could check the warning message. No big deal.

tests/test_ext_marshmallow.py
```diff
@@ -160,14 +145,17 @@ def test_can_use_schema_in_parameter(self, spec, schema):
...
+            reference = parameter['content']['application/json']['schema']
+
+        assert reference == {'$ref': ref_path(spec) + 'Pet'}
 
-        assert schema['name']['type'] == 'string'
-        assert schema['password']['type'] == 'string'
+        resolved_schema = spec.components._schemas['Pet']
+        assert resolved_schema['properties']['name']['type'] == 'string'
+        assert resolved_schema['properties']['password']['type'] == 'string'
         # dump_only field is ignored
```
lafrech: Remove comment. And perhaps change test to

```py
assert resolved_schema['properties']['id'] == ...
```

tests/test_ext_marshmallow.py
```diff
@@ -181,14 +169,17 @@ def test_can_use_schema_in_response(self, spec, schema):
...
+            reference = response['content']['application/json']['schema']
 
-        assert schema['id']['type'] == 'integer'
-        assert schema['name']['type'] == 'string'
+        assert reference == {'$ref': ref_path(spec) + 'Pet'}
+
+        resolved_schema = spec.components._schemas['Pet']
+        assert resolved_schema['properties']['id']['type'] == 'integer'
+        assert resolved_schema['properties']['name']['type'] == 'string'
         # load_only field is ignored
```
lafrech: Same as dump_only above.

tests/test_openapi.py
```diff
@@ -8,15 +8,16 @@
...
-from apispec.ext.marshmallow.openapi import OpenAPIConverter, MARSHMALLOW_VERSION_INFO
+from apispec.ext.marshmallow.openapi import MARSHMALLOW_VERSION_INFO
 from apispec import exceptions, utils, APISpec
 
-from .utils import get_definitions
-
+from .utils import get_definitions, ref_path
+from .conftest import make_spec
 
 @pytest.fixture(params=('2.0', '3.0.0'))
```
lafrech: Maybe move this feature to conftest.py.

tests/test_openapi.py
```diff
@@ -693,47 +653,46 @@ def test_field2property_nested_self_ref_with_meta(self, openapi):
         res = openapi.field2property(self_nesting2, name='Foo')
         assert res == {'$ref': '#/definitions/Bar'}
 
-    def test_field2property_nested_dump_only(self, openapi):
-        category = fields.Nested(CategorySchema)
-        res = openapi.field2property(category, name='Foo', dump=False)
-        props = res['properties']
-        assert 'breed' not in props
+    def test_nested2(self, spec):
```
lafrech: I'm not sure about naming and purpose of the test.

lafrech:
I left a few comments.

Icing on the cake, it I may suggest: cleaning history (squashing a few commits together to only have a few meaningful commits) would help reviewing and digging history in the future.

Bangertm:
Thanks for the feedback.  All should be addressed.  I didn't know you could match exception and warning messages so easily in pytest - super convenient.  

Everything is squashed down to one commit.  I was trying for two - one for moving the schema name resolution functionality and one for using references for schemas found in request and response bodies but I couldn't resolve the conflicts.

Bangertm:
One more thing... Noticed that `use_instances` isn't used anymore so I removed it from the function signature. That diff should probably be squashed too, but wanted to give you a chance to review.

lafrech:
`use_instances` was added in https://github.com/marshmallow-code/apispec/pull/144 to support partial schemas. This PR breaks the partial schema use case.

There is a wider discussion about this in https://github.com/marshmallow-code/apispec/issues/169. All fields modifiers are concerned (`partial`, `exclude`,...). Basically, resolving on classes ignores modifiers, resolving on instances would make user code unrealistically cumbersome. A third way could be to resolve on instances and also compare modifiers, if that is feasible. Until this is done, user should refrain form using modifiers.

I don't really mind the regression on `partial` if it is documented. It would be nice to have in mind a long-term solution to this. Like modifying `resolve_nested_schema` to resolve not only on the class but also on modifiers. I don't see any blocker right now.

Also, some `partial` logic was added by @sloria in https://github.com/marshmallow-code/apispec/commit/09e3febc9997d27d90a892f94b06d29cc98de995 before #144. I guess it should be removed, shouldn't it?

lafrech:
> Also, some `partial` logic was added by @sloria in https://github.com/marshmallow-code/apispec/commit/09e3febc9997d27d90a892f94b06d29cc98de995 before #144. I guess it should be removed, shouldn't it?

Scratch that.

But still extensions relying on `use_instances` will be broken by the change. Copying @theirix.

It would be nice to have a feature in marshmallow to compare schema instances (compare schema class and modifiers).

Bangertm:
The way things are implemented now has the effect that is somewhat similar to `use_instances` being hard coded to `True`.  If `resolve_schema_dict` is called with a schema instance it is schema instance not the schema class that will be added to the spec.  The `partial` modifier will be applied to that that schema.  

The difference is that the previous implementation would always return an inline definition and never a reference when `use_instances` is `Tue` because the schema instance would never been in the `ref` dictionary.  Now most likely `resolve_schema_dict` will return a reference.  There wouldn't be a problem if only one version of the schema is in the spec, but there will be collisions if the schema is being used with different modifiers.  

I tend to think that using the schema class plus the schema's modifiers as the key in `ref` might be the way to go.

lafrech:
> I tend to think that using the schema class plus the schema's modifiers as the key in ref might be the way to go.

@Bangertm, here's a draft of Schema comparison feature: https://github.com/marshmallow-code/marshmallow/pull/1062.

lafrech:
Thinking of it, the `__eq__` function is not so helpful. Besides, it would only work for marshmallow 3. And we only want to compare the modifiers that affect the documentation. We're better off keeping the logic here.

We may ignore `'context'` and `'unknown'`.

```
MODIFIERS = {'only', 'exclude', 'many', 'load_only', 'dump_only', 'partial'}
```

And as you suggested, use class + modifiers as key.

And we need to add some more auto-naming logic to manage the case when multiple instances of the same class are provided. It doesn't have to be clever. People wanting nice names may pass them manually. All we need is to make sure each schema has a different name. So it can be an incremented counter. AFAIU, we already need this anyway for the pathological case of the resolver generating the same name for two different schema classes.

Bangertm:
> And we need to add some more auto-naming logic to manage the case when multiple instances of the same class are provided. It doesn't have to be clever. People wanting nice names may pass them manually. All we need is to make sure each schema has a different name. So it can be an incremented counter. AFAIU, we already need this anyway for the pathological case of the resolver generating the same name for two different schema classes.

We might want to just warn the user of the name collision instead.  I can't image that people would be happy with Pet, Pet1, Pet2, etc in their spec.  If they have to take some action in this case anyway we might just want to warn them and tell them to either add the schema with a unique name or provide a custom schema name resolver.

lafrech:
Alright. Should we raise an exception then? If we don't offer a viable fallback, a warning might not be enough.

Bangertm:
Yes I guess raising an exception is best.  I was thinking to use a warning so that they could complete the operation, inspect the spec, and figure out what to do.  But in this case, a name clash is overwriting an existing schema in the spec so we should stop them where that is happening because in all likelihood that is not what they want to happen.  

Where should we raise from?  Very specifically in `OpenAPIConverter.resolve_nested_schema` where the problem might happen via automatic name resolution or more generically in `MarshmallowPlugin.schema_helper` or even `Components.schema`?

Bangertm:
Have now added some commits to include the instance modifiers in the key.  I'm pretty sure that we have to make this change if we go forward with the auto name resolver functionality.  Previously a schema like:

```python
class MultiModifierSchema(Schema):
     pet_unmodified = Nested(PetSchema)
     pet_exclude = Nested(PetSchema, exclude=('name',))
```
would have resolved properly with appropriate inline definitions for each field.  Without taking modifiers into account both fields will reference the last schema added even if they are both in the spec.

Also based on this case I added an ability to generate a unique name.  It would have been really annoying to get an exception when trying to add this schema to the spec and just warning the user would have yielded an incorrect spec.  So now the spec will be correct and the user will get an indication of how to get better names.  

Any thoughts on a better algorithm for generating the unique name?  I'm not super happy with it but can't think of something better right now.

apispec/ext/marshmallow/common.py
```diff
@@ -60,3 +61,28 @@ def make_schema_key(schema):
...
+        return name
+    if not counter:  # first time time through recursion
+        warnings.warn(
+            'Multiple schemas resolved to the name {} - name has been modified '
+            'either manually add each of the schemas with a different name or '
+            'provide custom schema_name_resolver'.format(name),
+        )
+    else:  # subsequent recursions
+        counter_str = str(counter)
+        name = name[:-len(counter_str)]
```
lafrech: [Minor] No need to define `counter_str` as it is not reused.

```py
    name = name[:-len(str(counter))]
```

lafrech:
I agree on the principle.

I had a quick look at the code. I'll review more thoroughly later.

Edit: This is wrong. This case is covered. ~I think there is still the (unlikely) possibility of a collision if `Pet` schema is resolved twice, generating `Pet` and `Pet1` while `Pet1` already exists. This might be complicated to address cleanly. Or we could add a UUID (`Pet-81c3209c-8aff-45c8-b2cf-67f746b0e07`) instead of a counter value. The names would suck, but the collisions would be unlikely. After all, it is just a fallback.~

lafrech:
I think this is the way to go.

Do you think you could rebase and update docs?

Bangertm:
Got the branch up to date and added some information in the docs on working with nested fields and modifiers.

apispec/ext/marshmallow/__init__.py
```diff
@@ -208,4 +190,14 @@ def operation_helper(self, operations, **kwargs):
...
+
+    def warn_if_schema_already_in_spec(self, schema_key):
+        """Method to warn the user if the schema has already been added to the
+        spec.
+        """
+        if schema_key in self.openapi.refs:
+            warnings.warn(
+                '{} has already been added to the spec. Adding it twice may '
+                'cause references to not resove properly'.format(schema_key[0]),
+            )
```
sloria, line 204: Nit: explicitly pass `UserWarning` here.

apispec/ext/marshmallow/common.py
```diff
@@ -39,3 +44,44 @@ def get_fields(schema):
...
+    :param int counter: the counter of the number of recursions
+    :return: the unique name
+    """
+    if name not in components._schemas:
+        return name
+    if not counter:  # first time time through recursion
+        warnings.warn(
+            'Multiple schemas resolved to the name {} - name has been modified '
+            'either manually add each of the schemas with a different name or '
+            'provide custom schema_name_resolver'.format(name),
```
sloria, line 82: Same here--explicitly pass `UserWarning`.

docs/using_plugins.rst
```diff
@@ -197,6 +197,45 @@ If your API uses `method-based dispatching <http://flask.pocoo.org/docs/0.12/vie
...
+Marshmallow Plugin
+----------
+
+Nesting Schemas
+***************
+
+By default, Marshmallow `Nested` fields are represented by a `JSON Reference object
+<https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#referenceObject>`_.
+If the schema has been added to the spec via `spec.components.schema <apispec.core.Components.schema>`
+the user supplied name will be used in the reference. Otherwise apispec will
```
sloria: Typo: "user-supplied".

docs/using_plugins.rst
```diff
@@ -197,6 +197,45 @@ If your API uses `method-based dispatching <http://flask.pocoo.org/docs/0.12/vie
...
+
+Marshmallow Plugin
+----------
+
+Nesting Schemas
+***************
+
+By default, Marshmallow `Nested` fields are represented by a `JSON Reference object
+<https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#referenceObject>`_.
+If the schema has been added to the spec via `spec.components.schema <apispec.core.Components.schema>`
```
sloria: Nit: missing comma after `spec.components.schema`.

docs/using_plugins.rst
```diff
@@ -197,6 +197,45 @@ If your API uses `method-based dispatching <http://flask.pocoo.org/docs/0.12/vie
...
+the user supplied name will be used in the reference. Otherwise apispec will
+add the nested schema to the spec using an automatically resolved name for the
+nested schema. The default `schema_name_resolver <apispec.ext.marshmallow.resolver>`
+function will resolve a name based on the schema's class `__name__` dropping a
```
sloria: Nit: missing comma after `__name__`.

docs/using_plugins.rst
```diff
@@ -197,6 +197,45 @@ If your API uses `method-based dispatching <http://flask.pocoo.org/docs/0.12/vie
     #            'post': {}}}
     #
 
+
+Marshmallow Plugin
+----------
```
sloria: Nit: make sure the underline lengths are correct, else Sphinx will raise a warning.

docs/using_plugins.rst
```diff
@@ -197,6 +197,45 @@ If your API uses `method-based dispatching <http://flask.pocoo.org/docs/0.12/vie
...
+function will resolve a name based on the schema's class `__name__` dropping a
+trailing "Schema" so that `class PetSchema(Schema)` resolves to "Pet".
+
+To change the behavior of the name resolution simply pass an alternative
+function accepting a `Schema` class and returning a string to the plugin's
+constructor. If the `schema_name_resolver` function returns a value that
+evaluates to `False` in a boolean context the nested schema will not be added to
+the spec an instead defined in line.
+
+Note: Circular Referencing Schemas cannot be defined in line due to infinite
```
sloria: Typo: "Circular-referencing schemas cannot be defined inline...with cirular-referencing schemas."

docs/using_plugins.rst
```diff
@@ -197,6 +197,45 @@ If your API uses `method-based dispatching <http://flask.pocoo.org/docs/0.12/vie
...
+recursion so a `schema_name_resolver` function must return a string name when
+working with circular referencing schemas.
+
+Schema Modifiers
+***************
+
+Instances of a Marshmallow Schema can be initialized with modifier parameters
```
sloria: Typo: "...marshmallow `Schema`...".

or 

"`Schema` instances can be initialized..."

docs/using_plugins.rst
```diff
@@ -197,6 +197,45 @@ If your API uses `method-based dispatching <http://flask.pocoo.org/docs/0.12/vie
...
+to exclude fields or ignore the absence of required fields. apispec will respect
+schema modifiers in the generated schema definition. If a particular schema is
+initialized in an application with modifiers it may be added to the spec with
```
sloria: Nit: Missing comma after "modifiers".

docs/using_plugins.rst
```diff
@@ -197,6 +197,45 @@ If your API uses `method-based dispatching <http://flask.pocoo.org/docs/0.12/vie
...
+each set of modifiers and apispec will treat each unique set of modifiers -
```
sloria: Nit: Use `--` to render a dash.

sloria (commented):
Just did a quick once-over, mostly on the docs. I just had a few nits. Can you also update CHANGELOG.rst? Feel free to create a new entry for "1.0.0 (unreleased)". I didn't look at this in-depth, so I'm not the best person to describe these changes concisely.

No need to block on me for merging. @Bangertm @lafrech Feel free to merge when you're confident with this.

docs/using_plugins.rst
```diff
@@ -197,6 +197,45 @@ If your API uses `method-based dispatching <http://flask.pocoo.org/docs/0.12/vie
...
+add the nested schema to the spec using an automatically resolved name for the
+nested schema. The default `schema_name_resolver <apispec.ext.marshmallow.resolver>`
+function will resolve a name based on the schema's class `__name__`, dropping a
+trailing "Schema" so that `class PetSchema(Schema)` resolves to "Pet".
+
+To change the behavior of the name resolution simply pass an alternative
+function accepting a `Schema` class and returning a string to the plugin's
+constructor. If the `schema_name_resolver` function returns a value that
+evaluates to `False` in a boolean context the nested schema will not be added to
+the spec an instead defined in line.
```
sloria: Typo: "and instead defined inline."

docs/using_plugins.rst
```diff
@@ -197,6 +197,45 @@ If your API uses `method-based dispatching <http://flask.pocoo.org/docs/0.12/vie
...
+
+Note: Circular-referencing schemas cannot be defined in line due to infinite
```
sloria: Typo: "inline".

sloria (changes requested):
A few more drive-by nits.

@Bangertm Can you please also update the CHANGELOG?

Bangertm:
@sloria changelog is updated.  Can you take a look to make sure it's clear.

apispec/ext/marshmallow/openapi.py
```diff
@@ -392,6 +391,45 @@ def field2property(self, field, use_refs=True, dump=True, load=True, name=None):
 
         return ret
 
+    def resolve_nested_schema(self, schema):
+        """Method to resolve a dictinary from schemas
```
sloria: Nit: "Method to" is unnecessary. Also, typo in "dictionary. How about "Return the Open API representation of a marshmallow `Schema`."

apispec/ext/marshmallow/openapi.py
```diff
@@ -392,6 +391,45 @@ def field2property(self, field, use_refs=True, dump=True, load=True, name=None):
...
+
+        Tests to see if the schema is already in the spec and adds it if
```
sloria: "Adds the schema to the spec if it isn't already present."

apispec/ext/marshmallow/openapi.py
```diff
@@ -392,6 +391,45 @@ def field2property(self, field, use_refs=True, dump=True, load=True, name=None):
...
+        necessary
+
+        Typically will return a dictionary with the reference to the schema's
+        path in the spec unless the `schema_name_resolver` returns `None` in
```
sloria: Comma after `None`.

> None, in which case

CHANGELOG.rst
```diff
@@ -1,6 +1,18 @@
 Changelog
 ---------
 
+1.0.0 (unreleased)
+++++++++++++++++++
+
+Features:
+
+- References nested Schemas by adding the Schema to the spec with an automatically
```
sloria: How about

```
- Automatically generate references to nested schemas with a computed name, e.g.
  ``fields.Nested(PetSchema())`` -> ``#components/schemas/Pet``.
- Automatically generate references for ``requestBody`` using the above mechanism.
```

Make sure to add periods after each item for consistency with the rest of the changelog.

sloria (commented):
Just a bit more copy editing.

@lafrech Since you're more familiar with the substance of this, would you mind doing the final pass?

lafrech:
@Bangertm, sorry for the delay about this. I just reviewed it again and I think t is good to go once @sloria's comments are addressed.

When this is merged, we can release a RC version (maybe with other fixes from https://github.com/marshmallow-code/apispec/milestone/2 it you have something ready) and we're close to a final v1!

lafrech:
Strangely enough, I wasn't notified about the push. Perhaps because it was a force-push.

Thanks @Bangertm for going through all this.

Let's merge this!

Anything else you'd like to add before we release a new RC version?

buxx:
Thank's @lafrech and @Bangertm for your work !

lafrech:
I've been working on a similar caching mechanism and it was a bit more tricky because the key could include `dict`s (https://github.com/Scille/umongo/pull/165).

I'm tempted to add a comment to avoid potential issues in the future:

```py
    # This works because modifiers are either strings or lists/tuples.
    # If MODIFIERS ever include dicts, this code will need to be adapted.
    for modifier in MODIFIERS:
        attribute = getattr(schema, modifier)
        try:
            hash(attribute)
        except TypeError:
            attribute = tuple(attribute)
        modifiers.append(attribute)
```

Issue #331, opened 2018-10-30
TypeError when adding non-Marshmallow definition


Bangertm:
After upgrading apispec to 0.29 I'm seeing a TypeError when adding a non-Marshmallow definition to the spec.  

For example: 

```
spec = APISpec(title='Demo',
               version=1.0,
               plugins=[MarshmallowPlugin()])

spec.definition('Example', properties={'key1': {'type': 'integer'},
                                          'key2': {'type': 'string'}})
```

Raises: TypeError: definition_helper() takes exactly 3 arguments (2 given).

The issue here seems to be calling the `definition_helper` of `MarshmallowPlugin` without a `schema` parameter.  This was previously caught via by including `TypeError` in this [except clause](https://github.com/marshmallow-code/apispec/blob/56b12fbb85224fc93284032986a55c8fa3d1168c/apispec/core.py#L278).

Our use case is that we have Marshmallow schemas for most of our routes, but there are a couple of simple return values that we never need to validate and are easy enough add to the schema via properties like above.

lafrech:
What about

- we add the default value `schema=None` to `MarshmallowPlugin.definition_helper` and return `{}` if schema is `None`?

```py
    def definition_helper(self, name, schema=None, **kwargs):
        if schema is None:
            return None
        ...
```

- maybe we add a `or {}` conditional in `add_path` to allow `definition_helpers` to return `None` rather than `{}`?

```py
    ret.update(plugin.definition_helper(name, definition=ret, **kwargs) or {})
```

Bangertm:
Testing for a schema parameter and returning `{}` if it's not passed makes sense to me.  

One other thing that looks odd is passing the definition parameter to definition_helper and then not using it - at least in any of the builtin plugins.

lafrech:
> Testing for a schema parameter and returning {} if it's not passed makes sense to me.

Alright. I'd rather get https://github.com/marshmallow-code/apispec/pull/305 merged before to avoid merge conflicts.

> One other thing that looks odd is passing the definition parameter to definition_helper and then not using it - at least in any of the builtin plugins.

This was decided in https://github.com/marshmallow-code/apispec/pull/122.

PR #336, opened 2018-11-06
Allow helpers to return None

lafrech:
Fixes https://github.com/marshmallow-code/apispec/issues/331.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -30,11 +30,21 @@ Features:
     spec.path('/pets/', operations=...)
     spec.components.schema('Pet', properties=...)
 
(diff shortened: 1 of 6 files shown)


lafrech:
@sloria, @Bangertm, any comment before I merge this?

Bangertm (commented):
Looks good.  Tested with the api that was causing the problem and it works great.

lafrech:
Alright, let's merge this. It's never too late to rework the CHANGELOG if it is unclear.

sloria:
Thanks guys. Since this is a bugfix, looks like it's time to release! 🚢 🇮🇹 

I'll get #338 merged in then cut a release.

PR #345, opened 2018-11-20
Remove normalize_path from APISpec.path

lafrech:
See https://github.com/marshmallow-code/apispec/issues/78#issuecomment-431846368 and below.


sloria:
I agree with your analysis in https://github.com/marshmallow-code/apispec/issues/78#issuecomment-432574250 . Let's go with it.

PR #359, opened 2019-01-24
Fix: Use existing reference if name resolver returns None

lafrech:
This reverses the logic in `resolve_nested_schema` to give priority to existing reference.


lafrech:
Ping @Bangertm. I'm not sure you were notified.

This modifies the code you wrote in #333.

Bangertm (approved):
Yes this looks good.  I hadn't considered this case.

lafrech:
Yeah, that's really a corner case.

Anyway, it's more logical this way. No need to generate a name if the schema is already registered.

Let's merge this.

PR #360, opened 2019-01-25
Revise field2properties

Bangertm:
Fixes an issue with an invalid reference object being created by extracting properties from field metadata before handling nested schemas.  

Also extracts methods from `field2properties` to make the method smaller and add documentation about each step.


sloria (approved):
Good refactor. Thanks!

Issue #198, opened 2018-04-09
Minor issues with field2choices


lafrech:
I think that in marshmallow swagger, some pathological cases are not handled the best way.

This

```python
        field = fields.Str(validate=[
            validate.Equal('only choice'),
            validate.Equal('only choice_2')])
```
produces

```
    ['enum'] = ['only choice', 'only_choice_2']
```

which is wrong. Admittedly, the field declaration does not make much sense.

Also

```python
        field = fields.Str(validate=[
            validate.Equal('only choice'),
            validate.OneOf(['freddie', 'brian', 'john'])])
```

produces

```
    ['enum'] = ['only choice']
```

which is wrong too. This field declaration is a big dodgy as well.

Those are corner cases that might happen, perhaps, in some complex inheritance scenarios.

apispec should at least return the correct output. Maybe raising a warning is out of scope, as if we engage in consistency validation in apispec, this could end up being a lot of work.

No time and interest to address this right now but I thought I'd leave a note here.

PR #364, opened 2019-01-28
Add support for outputting field patterns

DStape:


apispec/ext/marshmallow/openapi.py
```diff
@@ -330,6 +330,26 @@ def field2length(self, field, **kwargs):
...
+        """
+        validators = [
+            validator for validator in field.validators
+            if hasattr(validator, 'regex')
+        ]
+
+        attributes = {}
+        for validator in validators:
+            # `SRE_Pattern` for < Python3.7
+            if type(validator.regex).__name__ in ('SRE_Pattern', 'Pattern'):
```
sloria: Minor: This can be written a little more robustly with

```python
# compat.py

import re

RegexType = type(re.compile(''))
```

```python
from apispec.compat import RegexType

if isinstance(validator.regex, RegexType):
    #...
```

tests/test_openapi.py
```diff
@@ -239,6 +240,32 @@ def test_field_with_allow_none(self, openapi):
...
+
+    def test_field_with_no_pattern(self, openapi):
+        field = fields.Str()
+        ret = openapi.field2property(field)
+        assert 'pattern' not in ret
+
+    def test_field_with_multiple_patterns(self, openapi):
+        regex_validators = [
+            validate.Regexp('loser'),
+            validate.Regexp('winner'),
```
sloria, line 258: I'm not sure about this "last one wins" behavior, but I can't think of a better alternative at the moment.

Perhaps we should raise a warning if there are multiple `Regexp` validators on a field.

sloria (changes requested):
Thanks for the PR!

Just a couple suggestions.

tests/test_openapi.py
lafrech: Yeah, apispec doesn't really enforce correctness of the spec. There are other similar instances of undefined or invaild output (https://github.com/marshmallow-code/apispec/issues/198).

AFAIU, only one pattern can be specified (I could imagine validating with several patterns but I don't think OpenAPI allows that.). Let's assume the user knows what he is doing and he only specified one regex and we can write

```py
        validator = next(
            (v for v in field.validators if hasattr(v, 'regex') and isinstance(v.regex, RegexType)),
            None
        )
        if validator:
            attributes['pattern'] = v.regex.pattern
        return attributes
```

Or we go further and we add a warning if several regex are specified.

I think both are better than the "last one wins" method.

apispec/ext/marshmallow/openapi.py
DStape: I like this much better... done.

tests/test_openapi.py
DStape: Sure, code updated to build a generator expression. If can be iterated more than once a UserWarning gets raised.

DStape:
Just rebased, which should fix the failing CI build.

sloria (approved):
Made a tiny improvement to the warning message and test.

Thanks @DStape for updating this. Good to merge after Travis is green.

DStape:
No problem @sloria, thanks for merging!

PR #367, opened 2019-01-29
Refactor schema2 fields2

Bangertm:
Redistributes logic between `schema2jsonschema` and `fields2jsonschema` and between `schema2parameters` and `fields2parameters`.  Now the `fields2` methods only use a `fields` parameter instead of both `fields` and `schema`.  

Factors logic for filtering fields and warning the user about fields defined in `Meta.fields` and `Meta.additional` into common functions.

I think this helps to understand what is going on in these functions.

### DIFF
--- a/apispec/ext/marshmallow/common.py
+++ b/apispec/ext/marshmallow/common.py
@@ -21,7 +21,13 @@ def resolve_schema_instance(schema):
         return schema()
     if isinstance(schema, marshmallow.Schema):
         return schema
-    return marshmallow.class_registry.get_class(schema)()
+    try:
+        return marshmallow.class_registry.get_class(schema)()
+    except marshmallow.exceptions.RegistryError:
+        raise ValueError(
+            '{!r} is not a marshmallow.Schema subclass or instance and has not'
+            ' been registered in the Marshmallow class registry.'.format(schema),
+        )
 
 
 def resolve_schema_cls(schema):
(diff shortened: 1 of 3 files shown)


apispec/ext/marshmallow/common.py
```diff
@@ -21,7 +21,13 @@ def resolve_schema_instance(schema):
         return schema()
     if isinstance(schema, marshmallow.Schema):
         return schema
-    return marshmallow.class_registry.get_class(schema)()
+    try:
+        return marshmallow.class_registry.get_class(schema)()
+    except marshmallow.exceptions.RegistryError:
+        raise ValueError(
+            '{} is not a Marshmallow Schema subclass or instance and has not'
```
sloria: "Marshmallow" can be removed. Alternatively, you could say


> ...is not a marshmallow.Schema subclass...


Also, we should also probably show the `__repr__`  instead of `__str__`:

```python
'{!r} is not a marshmallow.Schema subclass...
```

apispec/ext/marshmallow/common.py
```diff
@@ -34,16 +40,68 @@ def resolve_schema_cls(schema):
...
+    """
     if hasattr(schema, 'fields'):
-        return schema.fields
+        fields = schema.fields
     elif hasattr(schema, '_declared_fields'):
-        return copy.deepcopy(schema._declared_fields)
-    raise ValueError("{0!r} doesn't have either `fields` or `_declared_fields`".format(schema))
+        fields = copy.deepcopy(schema._declared_fields)
+    else:
+        raise ValueError("{0!r} doesn't have either `fields` or `_declared_fields`".format(schema))
```
sloria: Nit: I know this was here before, but the `{0!r}` can be simplified to `{!r}`.

lafrech:
Looks good.

Can be merged once rebased.

sloria (commented):
Just a few nits. I can fix those up later, or feel free to do them and merge this.

apispec/ext/marshmallow/common.py
```diff
@@ -21,7 +21,13 @@ def resolve_schema_instance(schema):
         return schema()
     if isinstance(schema, marshmallow.Schema):
         return schema
-    return marshmallow.class_registry.get_class(schema)()
+    try:
+        return marshmallow.class_registry.get_class(schema)()
+    except marshmallow.exceptions.RegistryError:
+        raise ValueError(
+            '{} is not a Marshmallow Schema subclass or instance and has not'
+            ' been registered in the Marshmallow class registry'.format(schema),
```
sloria: Nit: Remove "Marshallow" and add a period at the end.

apispec/ext/marshmallow/openapi.py
```diff
@@ -522,13 +498,33 @@ def fields2parameters(
...
             "Schemas with many=True are only supported for 'json' location (aka 'in: body')"
 
-        exclude_fields = getattr(getattr(schema, 'Meta', None), 'exclude', [])
-        dump_only_fields = getattr(getattr(schema, 'Meta', None), 'dump_only', [])
+        fields = get_fields(schema, exclude_dump_only=True)
+
+        return self.fields2parameters(fields, default_in=default_in)
+
+    def fields2parameters(
+            self, fields, use_refs=True,
```
sloria: This appears overindented.

PR #374, opened 2019-02-01
Define valid HTTP method depending on openapi version

lafrech:
See https://github.com/marshmallow-code/apispec/issues/165#issuecomment-388753808.

Here are two different implementations (see intermediate commit). Tell me which you prefer, or if you prefer a third approach.

Let's not overthink this too much.

Do you think we need to add a test?

### DIFF
--- a/apispec/core.py
+++ b/apispec/core.py
@@ -6,7 +6,7 @@ from apispec.compat import iterkeys, iteritems
 from .exceptions import APISpecError, PluginMethodNotImplementedError, DuplicateComponentNameError
 from .utils import OpenAPIVersion, deepupdate
 
-VALID_METHODS = [
+VALID_METHODS_OPENAPI_V2 = [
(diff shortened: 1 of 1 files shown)


lafrech:
On second thought, the first approach

```py
VALID_METHODS = {
    2: VALID_METHODS_OPENAPI_V2,
    3: VALID_METHODS_OPENAPI_V3,
}
```

is more in line with the rest of `clean_operation`

```py
        ref_paths = {
            'parameter': {
                2: 'parameters',
                3: 'components/parameters',
            },
            'response': {
                2: 'responses',
                3: 'components/responses',
            },
        }
```

I can force-reset the second commit.

sloria:
Yeah, let's go with the first approach. Go ahead and merge after you've done the reset.

sloria (approved):

PR #381, opened 2019-02-04
Rework components interface

lafrech:
This addresses my comment here: https://github.com/marshmallow-code/apispec/pull/375#issuecomment-459724696.

I find it awkward to pass dicts as kwargs, and I think it would be clearer and more future-proof to pass the component `dict` as positional parameter, keeping kwargs only for plugin-specific parameters.

This is a totally breaking change. Tell me what you think.

I can try to fix the black issues. It would be easier if the changes were printed in the Travis log, as I wouldn't have to run it locally.

Docs would need an update.


apispec/core.py
```diff
@@ -108,15 +108,7 @@ def to_dict(self):
...
-    def schema(
-        self,
-        name,
-        properties=None,
-        enum=None,
-        description=None,
-        extra_fields=None,
-        **kwargs
-    ):
+    def schema(self, name, definition=None, **kwargs):
```
sloria: Rather than use an OAS 2 term, "definition", what if we use a consistent parameter name for all the component methods? "component", perhaps? "obj" could also work, since OAS calls these "Objects", but that could be conflated with Python's "object".

sloria (commented):
I support this change. Just a minor suggestion re: naming.

For black formatting, I suggest using a Docker image, https://github.com/marshmallow-code/marshmallow/issues/1104#issuecomment-454599345.

I can also fix this up for you if you don't have time.

lafrech:
Good. Parameter renamed. Code blacked. And I did a bit of cleanup while I was at it.

apispec/core.py
```diff
@@ -138,82 +133,81 @@ def schema(
...
+                ret.update(plugin.schema_helper(name, component, **kwargs) or {})
             except PluginMethodNotImplementedError:
                 continue
-        if properties:
-            ret["properties"] = properties
-        if enum:
-            ret["enum"] = enum
-        if description:
-            ret["description"] = description
-        if extra_fields:
```
sloria, line 154: Ah, nice. This cleans things up quite a bit.

`special_topics.rst` will need to be updated--we can probably just get rid of the "Adding Additional Fields To Schema Objects" section altogether.

Update: ended up modifying the section to show how to add additional properties to auto-generated properties.

sloria (commented):
Looks good. I'll fix up the docs and update the changelog.

sloria (approved):

PR #386, opened 2019-02-07
Don't document many=True schemas as array

### LINKED ISSUE #383: Docstring-only schemas incorrectly referenced as arrays?
Tested with Marshmallow 3.0.0rc3 and APISpec 1.0.0rc1.

Took a copy of the APISpec quickstart code and removed the defintions, so that schemas were only being referenced from the docstrings.  I'm seeing that the nested schema is being picked up as an array (incorrect) but then is also referenced as an array from the parent schema (correct).

Example (as short as I can make it, ignoring imports):

```py
spec = APISpec(
    title="Swagger Petstore",
    version="1.0.0",
    openapi_version="3.0.2",
    plugins=[FlaskPlugin(), MarshmallowPlugin()],
)

app = Flask(__name__)

class CategorySchema(Schema):
    id = fields.Int()
    name = fields.Str(required=True)

class PetSchema(Schema):
    category = fields.Nested(CategorySchema, many=True)
    name = fields.Str()

@app.route("/pet")
def random_pet():
    """
    ---
    get:
      responses:
        200:
          content:
            application/json:
              schema: PetSchema
    """
    return jsonify({'foo':'bar'})

@app.route("/category")
def random_category():
    """
    ---
    get:
      responses:
        200:
          content:
            application/json:
              schema: CategorySchema
    """
    return jsonify({'foo':'bar'})

with app.test_request_context():
    spec.path(view=random_pet)
    spec.path(view=random_category)
    print(spec.to_dict()
```
Here's what gets output (note that both the nested reference and the child are both array types):

```py
"schemas": {                                                  
  "Category": {                                               
    "type": "array",                                          
    "items": {                                                
      "type": "object",                                       
      "properties": {                                         
        "id": {"type": "integer", "format": "int32"}                                                     
      }                                                       
    }                                                         
  },                                                          
  "Pet": {                                                    
    "type": "object",                                         
    "properties": {                                           
      "category": {                                           
        "type": "array",                                      
        "items": {                                            
          "$ref": "#/components/schemas/Category"             
        }                                                     
      }                                                       
    }                                                         
  }                                                           
} 
```

I think the issue stems from fields2jsonschema method [here](https://github.com/marshmallow-code/apispec/blob/dev/apispec/ext/marshmallow/openapi.py#L647-L648) where the schema attributes are checked for many=True.  I would expect the schemas themselves would always be referenced as singletons, then references to those schemas would be checked for many.  I certainly might be missing something though, but at least in my lab just commenting out those 2 lines fixed everything.

Alternately, the schemas could be registered as they are in the example, but we try to do everything in docstrings if possible.  So if this is a non-issue feel free to close.

Many thanks for some awesome libraries!

lafrech:
Fixes #383.

This fix is based on @Bangertm's analysis in https://github.com/marshmallow-code/apispec/issues/383#issuecomment-461091813.

I'm gonna get pawned by black. I'll try to fix those issues at home if I get the time.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,12 @@ Features:
   from ``Regexp`` validators (:pr:`364`).
   Thanks :user:`DStape` for the PR.
 
+Bug fixes:
+
+- [apispec.ext.marshmallow]: Fix automatic documentation of schemas when
+  using ``Nested(MySchema, many==True)`` (:issue:`383`). Thanks
+  :user:`whoiswes` for reporting.
+
 Other changes:
 
 - *Backwards-incompatible*: Components properties are now passed as dictionaries rather than keyword arguments (:pr:`381`).
--- a/apispec/ext/marshmallow/openapi.py
+++ b/apispec/ext/marshmallow/openapi.py
@@ -450,7 +450,7 @@ class OpenAPIConverter(object):
             name = self.schema_name_resolver(schema_cls)
             if not name:
                 try:
-                    return self.schema2jsonschema(schema)
+                    json_schema = self.schema2jsonschema(schema)
                 except RuntimeError:
                     raise APISpecError(
                         "Name resolver returned None for schema {schema} which is "
@@ -459,6 +459,9 @@ class OpenAPIConverter(object):
                         " MarshmallowPlugin returns a string for all circular"
                         " referencing schemas.".format(schema=schema)
                     )
+                if getattr(schema, "many", False):
+                    return {"type": "array", "items": json_schema}
+                return json_schema
             name = get_unique_schema_name(self.spec.components, name)
             self.spec.components.schema(name, schema=schema)
         return self.get_ref_dict(schema_instance)
@@ -648,9 +651,6 @@ class OpenAPIConverter(object):
             jsonschema["title"] = Meta.title
         if hasattr(Meta, "description"):
             jsonschema["description"] = Meta.description
-
-        if getattr(schema, "many", False):
-            jsonschema = {"type": "array", "items": jsonschema}
 
         return jsonschema
 
--- a/tests/test_ext_marshmallow.py
+++ b/tests/test_ext_marshmallow.py
@@ -102,7 +102,6 @@ class TestDefinitionHelper:
 
     @pytest.mark.parametrize("schema", [AnalysisSchema, AnalysisSchema()])
     def test_resolve_schema_dict_auto_reference_return_none(self, schema):
-        # this resolver return None
         def resolver(schema):
             return None
 
@@ -161,6 +160,27 @@ class TestDefinitionHelper:
 
         assert "Pet" in definitions
         assert "Pet1" in definitions
+
+    def test_resolve_nested_schema_many_true_resolver_return_none(self):
+        def resolver(schema):
+            return None
+
+        class PetFamilySchema(Schema):
+            pets_1 = Nested(PetSchema, many=True)
+            pets_2 = List(Nested(PetSchema))
+
+        spec = APISpec(
+            title="Test auto-reference",
+            version="0.1",
+            openapi_version="2.0",
+            plugins=(MarshmallowPlugin(schema_name_resolver=resolver),),
+        )
+
+        spec.components.schema("PetFamily", schema=PetFamilySchema)
+        props = get_schemas(spec)["PetFamily"]["properties"]
+        pets_1 = props["pets_1"]
+        pets_2 = props["pets_2"]
+        assert pets_1["type"] == pets_2["type"] == "array"
 
 
 class TestComponentParameterHelper:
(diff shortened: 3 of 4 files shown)


lafrech:
BTW, I didn't understand this part of the code. It sounded related but I left it unchanged.

```py
            # marshmallow>=2.7.0 compat
            field.metadata.pop("many", None)
```

Any idea what this is about?

sloria:
Thanks! I'm off to do errands but I'll hopefully have some time later today or tomorrow to review.

> I'm gonna get pawned by black.

I really hope https://github.com/pre-commit/pre-commit.github.io/issues/211 becomes a reality...

sloria:
> Any idea what this is about?

https://github.com/marshmallow-code/apispec/pull/64 is the PR that added that. I don't recall the rationale, though.. =/

lafrech:
I've been using your pre-commit docker container at home but AFAIU, either I need to add my user to the docker group which is not recommended because it means more or less give my user root privileges, or I need to run it as root, which is not practical (and creates root-owned files in the repo).

I didn't take the time to try pyenv.

lafrech:
I don't see anything related in the 2.7.0 entry of the changelog.
https://github.com/marshmallow-code/marshmallow/blob/dev/CHANGELOG.rst#270-2016-04-04

Removing the line doesn't break any test.

sloria:
My guess is it's tied to this change: https://github.com/marshmallow-code/marshmallow/pull/413 , which may result in `many` being added to `field.metadata` (?).

lafrech:
Yes, probably.

But I don't think it makes any difference as we only pull a limited list of items from metadata (`VALID_PROPERTIES` + `x-*`).

I'd be tempted to remove it...

sloria:
Sure, it's your call. If there's no way for `many` to make it incorrectly into the OAS doc, then it can be removed.

lafrech:
One one hand, I'm pretty confident. On the other hand, I don't see what changed since the change was accepted, so if there was a good reason by the time, it should still be valid. I'm still in favor of removing. Happy if @Bangertm has an opinion about this and this PR.

Bangertm:
Looks good to me.  Thanks.

sloria (approved):
Code looks good, and I tested it out on the code in the OP. Works as expected.

```python
from apispec import APISpec
from apispec_webframeworks.flask import FlaskPlugin
from flask import Flask, jsonify
from pprint import pprint

spec = APISpec(
    title="Swagger Petstore",
    version="1.0.0",
    openapi_version="3.0.2",
    plugins=[FlaskPlugin(), MarshmallowPlugin()],
)

app = Flask(__name__)


class CategorySchema(Schema):
    id = fields.Int()
    name = fields.Str(required=True)


class PetSchema(Schema):
    category = fields.Nested(CategorySchema, many=True)
    name = fields.Str()


@app.route("/pet")
def random_pet():
    """
    ---
    get:
      responses:
        200:
          content:
            application/json:
              schema: PetSchema
    """
    return jsonify({"foo": "bar"})


@app.route("/category")
def random_category():
    """
    ---
    get:
      responses:
        200:
          content:
            application/json:
              schema: CategorySchema
    """
    return jsonify({"foo": "bar"})


with app.test_request_context():
    spec.path(view=random_pet)
    spec.path(view=random_category)
    pprint(spec.to_dict())
# {'components': {'parameters': {},
#                 'responses': {},
#                 'schemas': {'Category': {'properties': {'id': {'format': 'int32',
#                                                                'type': 'integer'},
#                                                         'name': {'type': 'string'}},
#                                          'required': ['name'],
#                                          'type': 'object'},
#                             'Pet': {'properties': {'category': {'items': {'$ref': '#/components/schemas/Category'},
#                                                                 'type': 'array'},
#                                                    'name': {'type': 'string'}},
#                                     'type': 'object'}},
#                 'securitySchemes': {}},
#  'info': {'title': 'Swagger Petstore', 'version': '1.0.0'},
#  'openapi': '3.0.2',
#  'paths': OrderedDict([('/pet',
#                         OrderedDict([('get',
#                                       {'responses': {200: {'content': {'application/json': {'schema': {'$ref': '#/components/schemas/Pet'}}}}}})])),
#                        ('/category',
#                         OrderedDict([('get',
#                                       {'responses': {200: {'content': {'application/json': {'schema': {'$ref': '#/components/schemas/Category'}}}}}})]))]),
#  'tags': []}
```

sloria:
This closes the last issue in the 1.0 milestone. Are we a go to release? @lafrech @Bangertm

lafrech:
Thanks for finishing this.

There's still this

```py
            # marshmallow>=2.7.0 compat
            field.metadata.pop("many", None)
```

but it can wait. No big deal. AFAIU, it is useless but doesn't hurt.

I don't think any of the open issues justifies postponing 1.0.0, so let's go!

lafrech:
OK, I get it. When that change was introduced in #64, there was a line below in the code that would include all metadata unconditionally:

https://github.com/lucasrcosta/apispec/blob/d9697baebfbba67808d42ec5a90de0cebc0cae29/apispec/ext/marshmallow/swagger.py#L129

```py
ret.update(field.metadata)
```

This was before `VALID_PROPERTIES` were introduced.

It can now be safely removed. Sending a PR.

PR #403, opened 2019-03-07
Allow adding schema with an existing name if it is identical to the existing schema

### LINKED ISSUE #402: Adding schema with an existing name should be allowed if the schemas are identical
Currently adding the same (identical) schema with the same name raises a `DuplicateComponentNameError`.

Normally this makes sense, but it causes issues in combination with [apispec_oneofschema](https://github.com/timakro/apispec-oneofschema) when trying to register two `OneOfSchema` instances which multiplex the same underlying schema.

Here is a minimal example (where `SchemaA` is used in both `SchemaAorB` and `SchemaAorC`).
```python
from apispec import APISpec
from apispec_oneofschema import MarshmallowPlugin 
from marshmallow_oneofschema import OneOfSchema

class SchemaA(Schema):
    id = fields.Integer()

class SchemaB(Schema):
    id = fields.Integer()

class SchemaC(Schema):
    id = fields.Integer()

class SchemaAorB(OneOfSchema):
    type_schemas = {
        "ClassA": SchemaA,
        "ClassB": SchemaB,
    }

class SchemaAorC(OneOfSchema):
    type_schemas = {
        "ClassA": SchemaA,
        "ClassC": SchemaC,
    }

spec = APISpec( 
    title="MyAPI",
    version="1.0.0",
    openapi_version="3.0.2", 
) 

spec.components.schema("SchemaAorB", schema=SchemaAorB)

# The following line raises DuplicateComponentNameError because
# SchemaA has already been registered as part of SchemaAorB.
spec.components.schema("SchemaAorC", schema=SchemaAorC)
```
Error:
```
~/code/apispec/apispec/core.py in schema(self, name, component, **kwargs)
    132         if name in self._schemas:
    133             raise DuplicateComponentNameError(
--> 134                 'Another schema with name "{}" is already registered.'.format(name)
    135             )
    136         component = component or {}

DuplicateComponentNameError: Another schema with name "SchemaA" is already registered.
```

It would be great if `apispec` did not raise an error in this situation, provided that the schema to be added is identical to the existing one with the same name. It should of course continue to raise an error if the schemas are actually different.

I'll submit a PR with a suggested fix soon. (I hope I'm not missing any reason why it is a bad idea to allow this.)

maxalbert:
Fixes #402.

It simply moves the check for the duplicate name _after_ the new schema has been constructed and compares the existing one to the new one. The `DuplicateComponentNameError` is raised only if the new schema is different to the existing one.

I also tweaked the test so that it first adds the same schema again under the same name (which should succeed) and afterwards tries to add a different schema under the same name (which should fail).

Any comments are welcome.

### DIFF
--- a/apispec/core.py
+++ b/apispec/core.py
@@ -129,10 +129,6 @@ class Components(object):
 
         https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#schemaObject
         """
-        if name in self._schemas:
-            raise DuplicateComponentNameError(
-                'Another schema with name "{}" is already registered.'.format(name)
-            )
         component = component or {}
         ret = component.copy()
         # Execute all helpers from plugins
@@ -141,6 +137,12 @@ class Components(object):
                 ret.update(plugin.schema_helper(name, component, **kwargs) or {})
             except PluginMethodNotImplementedError:
                 continue
+
+        if name in self._schemas and ret != self._schemas[name]:
+            raise DuplicateComponentNameError(
+                'A different schema with name "{}" is already registered.'.format(name)
+            )
+
         self._schemas[name] = ret
         return self
 
--- a/apispec/exceptions.py
+++ b/apispec/exceptions.py
@@ -11,7 +11,7 @@ class PluginMethodNotImplementedError(APISpecError, NotImplementedError):
 
 
 class DuplicateComponentNameError(APISpecError):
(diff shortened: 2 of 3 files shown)


sloria:
Thanks for the PR. Consensus on #402 seems to be that this shuold be handled in apispec_oneofschema rather than in apispec, so closing.

PR #407, opened 2019-03-12
Fixed example app, added route to serve apispec.

jptreen:
Made the example application not return 500 when you try and use it (the 'get_random_pet function' did not exist).
Added flask routes to serve the generated apispec at /swagger.json and /swagger.yml.

This would have made me feel warm and fuzzy when I encountered this project half an hour ago, hopefully it will now do that for somebody else.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -41,13 +41,16 @@ Example Application
 
 .. code-block:: python
 
+    import json
+    import yaml
+    from collections import Counter
     from apispec import APISpec
     from apispec.ext.marshmallow import MarshmallowPlugin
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -41,38 +41,46 @@ Example Application
...
     class PetSchema(Schema):
         category = fields.Nested(CategorySchema, many=True)
         name = fields.Str()
-
-
+        hits = fields.Int()
+    
+    
     # Optional Flask support
+    # navigate to /swagger.json or /swagger.yml for genreated spec.
```
lafrech, line 75: s/genreated/generated

PR #408, opened 2019-03-13
WIP - Add Components.build_ref

lafrech:
Provides an abstraction to reference a component without having to deal with component paths that depend on OAS version.

```python
            # OAS 2
            spec.components.build_ref('schema', 'MySchema')
            # '#/definitions/MySchema'
            spec.components.build_ref('security_scheme', 'MySecurityScheme')
            # '#/securityDefinitions/MySecurityScheme')
            # OAS 3
            spec.components.build_ref('schema', 'MySchema')
            # '#/components/schemas/MySchema'
            spec.components.build_ref('security_scheme', 'MySecurityScheme')
            # '#/components/securitySchemes/MySecurityScheme')
```

### DIFF
--- a/src/apispec/core.py
+++ b/src/apispec/core.py
@@ -86,6 +86,21 @@ class Components(object):
     They became sub-fields of "components" top-level field in OAS v3.
     """
 
+    COMPONENT_SUBSECTIONS = {
+        2: {
+            'schema': 'definitions',
+            'parameter': 'parameters',
+            'response': 'responses',
+            'security_scheme': 'securityDefinitions',
+        },
+        3: {
+            'schema': 'schemas',
+            'parameter': 'parameters',
+            'response': 'responses',
+            'security_scheme': 'securitySchemes',
+        },
+    }
+
     def __init__(self, plugins, openapi_version):
         self._plugins = plugins
         self.openapi_version = openapi_version
@@ -95,17 +110,12 @@ class Components(object):
         self._security_schemes = {}
 
     def to_dict(self):
-        if self.openapi_version.major < 3:
-            schemas_key = "definitions"
-            security_key = "securityDefinitions"
-        else:
-            schemas_key = "schemas"
-            security_key = "securitySchemes"
+        component_subsections = self.COMPONENT_SUBSECTIONS[self.openapi_version.major]
         return {
-            "parameters": self._parameters,
-            "responses": self._responses,
-            schemas_key: self._schemas,
-            security_key: self._security_schemes,
+            component_subsections["schema"]: self._schemas,
+            component_subsections["parameter"]: self._parameters,
+            component_subsections["response"]: self._responses,
+            component_subsections["security_scheme"]: self._security_schemes,
         }
 
     def schema(self, name, component=None, **kwargs):
--- a/src/apispec/core.py
+++ b/src/apispec/core.py
@@ -220,6 +220,29 @@ class Components(object):
         self._security_schemes[component_id] = component
         return self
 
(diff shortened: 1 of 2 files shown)


sloria:
Code looks fine, but what is the use case for this?

lafrech:
The marshmallow plugin automatically registers and references schemas.

apispec also allows to register other kinds of components but referencing them is still a burden. It is a bit verbose, and it requires that the user builds the whole component path.

This methods abstracts the component path and even makes it possible to reference the component from client code without knowing the OAS version in advance.

Now, I'm realizing this is only useful for people passing their parameters/responses as Python structures in the code, which is the case when using flask-rest-api, for instance. But I don't think it can serve people using YAML docstrings, so it may be a corner case.

I'm working on a flask-rest-api feature that would use that. I'll provide it as an examples when it is more advanced. I could put an equivalent code in flask-rest-api instead. No big deal. Putting it here would avoid a little duplication and I thought it could benefit other users, but that's debatable.

I'm not settled on this, hence the "draft" PR status.

Bangertm:
It would be convenient if we could define these paths on one place and then us it in the places where we are building reference paths: 

https://github.com/marshmallow-code/apispec/blob/78caabe30dfc575b900e3a69c00b4c59dc72d82d/src/apispec/ext/marshmallow/openapi.py#L694-L701

https://github.com/marshmallow-code/apispec/blob/78caabe30dfc575b900e3a69c00b4c59dc72d82d/src/apispec/core.py#L41-L60

lafrech:
Good point.

I forgot about `clean_operations`, and in fact, I'm thinking it may nix my initial use case. Maybe I'm trying to achieve something (building refs) I don't have to because `clean_operations` will ultimately do it anyway. I shall check that when get back home, in the flask-rest-api branch I'm working on.

There would still be room for refactor and as @Bangertm pointed out. And maybe `clean_operations` doesn't really reflect what it does but oh well.

Putting this on hold for now.

lafrech:
Note to self: while refactoring, modify the logic in `to_dict` to avoid returning empty sections.

I just discovered that ReDoc adds an empty "Authentication" section on top of the page when there is an empty `securitySchemes` dict in the spec.

For now, I patched it in my application code:

```py
    compo_to_dict_orig = api.spec.components.to_dict

    def compo_to_dict_fix():
        ret = compo_to_dict_orig()
        for section in {
                'schemas', 'parameters', 'responses', 'securitySchemes'}:
            if ret.get(section) == {}:
                del ret[section]
        return ret

    api.spec.components.to_dict = compo_to_dict_fix
```

lafrech:
Oops, I marked this "ready for review" by mistake. Sorry about that.

Looks like I can't undo.

lafrech:
> I forgot about `clean_operations`, and in fact, I'm thinking it may nix my initial use case. Maybe I'm trying to achieve something (building refs) I don't have to because `clean_operations` will ultimately do it anyway. I shall check that when get back home, in the flask-rest-api branch I'm working on.

Exactly. I confirm this PR is useless. Sorry about the noise.

> Note to self: while refactoring, modify the logic in to_dict to avoid returning empty sections.

I've been working on this already. I shall submit a PR when I'm done.

Also, in `tests.utils`, `ref_path` assumes the reference is a schema reference, but the function could be extended to work for parameters, responses, etc. This is related to the refactor mentioned above as it could use the same mapping.

Keeping this open not to forget about the refactor.

lafrech:
Closing this as issues discussed in comments are addressed in #419 and #420.

PR #409, opened 2019-03-13
MarshmallowPlugin: Resolve Schema in response headers

lafrech:
TODO

- [x] Tests
- [x] Doc ?


lafrech:
Thanks @sloria for fixing black issues. I was busy this WE.

PR #414, opened 2019-03-29
 Fix reference parameter/response with MarshmallowPlugin 

lafrech:
`MarshmallowPlugin` doesn't play well with parameter references because it assumes a parameter is a `dict`, which is not the case when passing a reference by name (the plugin works before `clean_operations` transforms the name into a full reference `dict`: `{"$ref": name})`, and it tries a `get` on the string.

This PR fixes this.

It also addresses a similar issue with responses. There was no error in this case because the resolve function does not use `get` but only `in`, which doesn't crash because the string is iterable, but I figured this was luck and it would be better to eliminate `dict` instances first.

### DIFF
--- a/tests/test_ext_marshmallow.py
+++ b/tests/test_ext_marshmallow.py
@@ -571,7 +571,7 @@ class TestOperationHelper:
             },
         )
         get = get_paths(spec_fixture.spec)["/pet"]["get"]
-        len(get["parameters"]) == 1
+        assert len(get["parameters"]) == 1
         resolved_schema = {
             "type": "array",
             "items": {"$ref": ref_path(spec_fixture.spec) + "Pet"},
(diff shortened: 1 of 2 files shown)


tests/test_ext_marshmallow.py
```diff
@@ -684,6 +683,46 @@ def test_schema_partially_v3(self, spec_fixture):
             },
         }
 
+    def test_parameter_reference(self, spec_fixture):
+        if spec_fixture.spec.openapi_version.major < 3:
+            param = {"schema": PetSchema}
+            reference = '#/parameters/Pet'
```
lafrech: Note: I'd like to extend `tests.utils.ref_path` to parameter, responses and all kinds of components (https://github.com/marshmallow-code/apispec/pull/408#issuecomment-478176641). This can be done later.

PR #416, opened 2019-03-31
fix unique schema key issues

zedrdave:
Fixed 2 issues with current code:

```py
        try:
            hash(attribute)
        except TypeError:
            attribute = tuple(attribute)
```

1. If `attribute` is immutable (and can be hashed) nothing is actually done (the hash is not assigned).

2. For schema attributes that are `set` (such as `only`, `exclude` etc), an exception is triggered and a tuple is used, which randomly fails to match on itself (since `str(tuple(x))` is non-deterministic when `x` is a `set`)…


lafrech:
Thanks for submitting this. I think it is on the right track.

We could discuss what needs to be covered by a `try`/`except` and by a `if` condition, but on the principle, it looks good to me.

Do you think you could add tests?

lafrech:
```py
MODIFIERS = ["only", "exclude", "load_only", "dump_only", "partial"]
```

Modifiers can be strings or iterables (hashable or not) such as list, tuple, set.

Would the code be simpler and the intent clearer if we just wrote it like this?

```py
        try:
            # Hashable (string, tuple)
            attribute = hash(attribute)
        except TypeError:
            # Unhashable iterable (list, set)
            attribute = hash(frozenset(attribute))
```

Bangertm:
Is there an example of a case where classes with the same modifiers are not matching properly?

The original intent of `make_schema_key` was to create a key that can be used in the `ref` dictionary.  We were originally using just the schema class for the key but when we wanted to take modifiers into account we need a way to include them in that key.  My thought process was to create a NamedTuple that included the class and each of the modifiers, but since some of the modifiers are not hashable (sets) we needed to convert them to something that is hashable.

The intent behind this code block is to test to see if the attribute is hashable and if not make it hashable.  I wanted to avoid storing the hash in the NamedTuple for debug reasons i.e. if someone wanted to inspect the `ref` dictionary it would be more convenient to see something that looked like the original modifier than a random number.  

Maybe instead of changing unhasable attributes to tuples we just change them to frozensets?

lafrech:
> Maybe instead of changing unhasable attributes to tuples we just change them to frozensets?

Good point.

```python
        try:
            # Hashable (string, tuple)
            hash(attribute)
        except TypeError:
            # Unhashable iterable (list, set)
            attribute = frozenset(attribute)
```

lafrech:
I think my last proposal is in line with @Bangertm's intent and fixes the bug.

@zedrdave, would you like to

- update your PR 
- add a test case
- add yourself to AUTHORS.rst

?

zedrdave:
@lafrech Hi, I finally have a bit of spare time. Will try and take care of this today.

tests/test_ext_marshmallow_common.py
```diff
@@ -19,13 +19,19 @@ def test_raise_if_schema_class_passed(self):
     def test_same_schemas_instances_equal(self):
         assert make_schema_key(PetSchema()) == make_schema_key(PetSchema())
 
+    def test_same_schemas_instances_with_sets_equal(self):
+        # Note: this test relies on non-deterministic hashing behaviour:
+        for i in range(0, 50):
+            assert make_schema_key(PetSchema(load_only=('1', '2', '3', '4'))) == make_schema_key(PetSchema(load_only=('4', '3', '2', '1')))
+
```
lafrech, line 26: Those are not sets but tuples, so I don't think this test tests what it means to.

I understand the reason for the loop, but I wouldn't do that. I'd just test with a set and a loop.

```py
assert make_schema_key(PetSchema(load_only=['1', '2', '3', '4'])) == make_schema_key(PetSchema(load_only=['4', '3', '2', '1']))
assert make_schema_key(PetSchema(load_only=set(['1', '2', '3', '4']))) == make_schema_key(PetSchema(load_only=set(['4', '3', '2', '1'])))
```

lafrech:
Do you think you could rebase on dev (rather than merge dev in your branch)? It would make the diff and the history more readable. At this point, you're better off starting a new branch from dev and cherry-picking your fix. You can push force in this branch or start a new PR if that's more convenient to you.

I that's too much trouble, I'll try to find the time to pick this up.

lafrech:
I did the rework in https://github.com/marshmallow-code/apispec/pull/439.

zedrdave:
@lafrech Thanks for doing the rewrite! (sorry for the lack of reactivity… haven't had lots of time to work on side-projects lately)

PR #421, opened 2019-04-02
Strip empty component sections from doc

lafrech:
This PR removes empty component subsections (parameters, responses, etc.) from the docs.

I'm tempted not to consider that a breaking change.

### DIFF
--- a/tests/test_ext_marshmallow.py
+++ b/tests/test_ext_marshmallow.py
@@ -51,7 +51,8 @@ class TestDefinitionHelper:
             openapi_version="2.0",
             plugins=(MarshmallowPlugin(schema_name_resolver=resolver),),
         )
-        assert get_schemas(spec) is None
+        with pytest.raises(KeyError):
+            get_schemas(spec)
 
         spec.components.schema("analysis", schema=schema)
         spec.path(
@@ -84,7 +85,8 @@ class TestDefinitionHelper:
             openapi_version="2.0",
             plugins=(MarshmallowPlugin(schema_name_resolver=resolver),),
         )
-        assert get_schemas(spec) is None
+        with pytest.raises(KeyError):
+            get_schemas(spec)
 
         spec.components.schema("analysis", schema=schema)
         spec.path(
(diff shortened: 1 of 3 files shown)


tests/test_ext_marshmallow.py
```diff
@@ -51,7 +51,7 @@ def resolver(schema):
             openapi_version="2.0",
             plugins=(MarshmallowPlugin(schema_name_resolver=resolver),),
         )
-        assert {} == get_schemas(spec)
+        assert get_schemas(spec) is None
```
sloria: Nit: It might be good for the `get_` functions to default to a sentinel value, like `marshmallow.missing`.

sloria (approved):
Just a little nit. Go ahead and address and merge, or we can address it later.

I wouldn't consider this a breaking change because it's not an API change. This warrants a minor version bump.

tests/test_ext_marshmallow.py
lafrech: I'd rather avoid `missing` as this could pass a wrong change leaking the `missing` object into the generated spec.

I could create a dedicated `MISSING_FROM_SPEC` singleton.

Overall, it looks just as simple to catch `KeyError`. Also, when using OAS3, `KeyError` covers the case where the whole `components` section is missing. See new commit.

sloria: Works for me.

sloria:
Is there a corresponding issue for this PR? I thought I remember seeing one but I can't find it.

lafrech:
Just a comment: https://github.com/marshmallow-code/apispec/pull/408#issuecomment-475682734.

lafrech:
Merged. Thanks for reviewing.

sloria:
No prob. Will you take care of the release, or shall I do that?

lafrech:
I can. Never did. Do I just need to push a tag?

sloria:
Yeah, pretty much. The process is the same as marshmallow's: https://github.com/marshmallow-code/marshmallow/blob/dev/RELEASING.md

lafrech:
Thanks. I never noticed this file.

lafrech:
Done.

I added release notes on Tidelift for 1.1.0 and 1.1.1. 1.1.2 and 1.2.0 are not there yet.

PR #423, opened 2019-04-05
Coerce response status codes to strings

### LINKED ISSUE #422: Consider coercing response codes to strings
The OpenAPI specification [allows responses of the form 2XX, 4XX, etc](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#patterned-fields-1). Using this library, it's possible in a docstring to write something like this:

 ```yaml
responses:
    200:
        content:
            application/json:
                schema: GoodResponseSchema
    4XX:
        content:
            application/json:
                schema: RequestErrorSchema
```

This causes the following error:

```
File "/Users/zach/.pyenv/versions/3.6.5/lib/python3.6/json/encoder.py", line 353, in _iterencode_dict
    items = sorted(dct.items(), key=lambda kv: kv[0])
TypeError: '<' not supported between instances of 'str' and 'int'
```

Which happens because the `200` key is stored as an `int`, and the `4XX` key is stored as a string. A preprocessing step to make them all strings would fix that problem.

zachmullen:
Fixes #422

### DIFF
--- a/src/apispec/core.py
+++ b/src/apispec/core.py
@@ -68,7 +68,7 @@ def clean_operations(operations, openapi_major_version):
             ]
         if "responses" in operation:
             for code, response in iteritems(operation["responses"]):
-                operation["responses"][code] = get_ref(
+                operation["responses"][str(code)] = get_ref(
                     "response", response, openapi_major_version
                 )
 


sloria:
@zachmullen Are you still working on this?

zachmullen:
No, I won't have time to work on this for the immediate future unfortunately.

lafrech:
I sent a similar PR: https://github.com/marshmallow-code/apispec/pull/424.

PR #427, opened 2019-04-09
Marshmallow exclude field allows for tuple

blagasz:
Correct unnecessary error when Marshmallow exclude field is defined as a tuple.

### DIFF
--- a/src/apispec/ext/marshmallow/common.py
+++ b/src/apispec/ext/marshmallow/common.py
@@ -95,7 +95,7 @@ def filter_excluded_fields(fields, Meta, exclude_dump_only):
     :param Meta: the schema's Meta class
     :param bool exclude_dump_only: whether to filter fields in Meta.dump_only
     """
-    exclude = getattr(Meta, "exclude", [])
+    exclude = list(getattr(Meta, "exclude", []))
     if exclude_dump_only:
         exclude += getattr(Meta, "dump_only", [])
 
--- a/src/apispec/ext/marshmallow/common.py
+++ b/src/apispec/ext/marshmallow/common.py
@@ -97,7 +97,7 @@ def filter_excluded_fields(fields, Meta, exclude_dump_only):
     """
     exclude = list(getattr(Meta, "exclude", []))
     if exclude_dump_only:
-        exclude += getattr(Meta, "dump_only", [])
+        exclude.extend(getattr(Meta, "dump_only", []))
 
     filtered_fields = OrderedDict(
         (key, value) for key, value in fields.items() if key not in exclude


lafrech:
Thanks.

Do you have a test case showing the error?

blagasz:
Hi!

I am not sure what you mean by a test case, but I recived the following
error message in a project after upgrading to the new version of apispec:

Traceback (most recent call last):
  File "/usr/lib/python3.7/runpy.py", line 183, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "/usr/lib/python3.7/runpy.py", line 142, in _get_module_details
    return _get_module_details(pkg_main_name, error)
  File "/usr/lib/python3.7/runpy.py", line 109, in _get_module_details
    __import__(pkg_name)
  File
"/home/blagasz/Nextcloud/repos/heartbit-backend/heartbit/__init__.py", line
395, in <module>
    spec.path(view=piece_item_event_list)
  File
"/home/blagasz/.venv/heartbit/lib/python3.7/site-packages/apispec/core.py",
line 301, in path
    plugin.operation_helper(path=path, operations=operations, **kwargs)
  File
"/home/blagasz/.venv/heartbit/lib/python3.7/site-packages/apispec/ext/marshmallow/__init__.py",
line 202, in operation_helper
    operation["parameters"]
  File
"/home/blagasz/.venv/heartbit/lib/python3.7/site-packages/apispec/ext/marshmallow/__init__.py",
line 93, in resolve_parameters
    schema_instance, default_in=parameter.pop("in"), **parameter
  File
"/home/blagasz/.venv/heartbit/lib/python3.7/site-packages/apispec/ext/marshmallow/openapi.py",
line 496, in schema2parameters
    fields = get_fields(schema, exclude_dump_only=True)
  File
"/home/blagasz/.venv/heartbit/lib/python3.7/site-packages/apispec/ext/marshmallow/common.py",
line 69, in get_fields
    return filter_excluded_fields(fields, Meta, exclude_dump_only)
  File
"/home/blagasz/.venv/heartbit/lib/python3.7/site-packages/apispec/ext/marshmallow/common.py",
line 100, in filter_excluded_fields
    exclude += getattr(Meta, "dump_only", [])
TypeError: can only concatenate tuple (not "list") to tuple

Do you mean to create a proper test case for the repo from that?

Regards, Sz'

src/apispec/ext/marshmallow/common.py
```diff
@@ -95,7 +95,7 @@ def filter_excluded_fields(fields, Meta, exclude_dump_only):
     :param Meta: the schema's Meta class
     :param bool exclude_dump_only: whether to filter fields in Meta.dump_only
     """
-    exclude = getattr(Meta, "exclude", [])
+    exclude = list(getattr(Meta, "exclude", []))
     if exclude_dump_only:
         exclude += getattr(Meta, "dump_only", [])
```
lafrech: I think `dump_only` could be a tuple as well.

blagasz: Yes. I tested this a little more:
tuple + tuple is OK
list + list is also OK
That's why it only throws error, when one of them is set to a tuple and the other receives its default value as list.

So it should always be the same type before concatenation, not necessarily list, could be tuple.
The list's extend method might result in a more elegant solution.

lafrech:
I think this is the right fix.

I scanned the marshmallow ext files and I didn't find any other place where a similar issue occurs.

Do you think you could

- add a test case demonstrating the fix works
- fix the typo in the docstring while you're at it ("A dictionary of of fields")
- add yourself to AUTHORS.rst

?

blagasz:
Yes, I'll we do all that, maybe in a new PR, as there is one unnecessary
commit in the current PR.

Regards, Sz'

blagasz:
Adding a new pull request.

PR #431, opened 2019-04-11
Fix Marshmallow Schema's Meta exclude field error

blagasz:
Mixing tuple and list when collecting fields in the Marshmallow extension may result in error.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -54,3 +54,4 @@ Contributors (chronological)
 - Marcin Lulek `@ergo <https://github.com/ergo>`_
 - Jonathan Beezley `@jbeezley <https://github.com/jbeezley>`_
 - David Stapleton `@dstape <https://github.com/DStape>`_
+- Szabolcs Blága `@blagasz <https://github.com/blagasz>`_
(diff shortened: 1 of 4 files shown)


lafrech:
Good.

Do you think you could test all combinations ([dump_only; exclude] x [tuple; list])?

Also, there are CI failures.

You should use a set, not a tuple, in the test, as the order doesn't matter (or perhaps better add `ordered=True` to the schema Meta).

(Note: no need to close/recreate a PR, you can amend/rebase and then `push -f`.)

blagasz:
Ok, I check that.

(Yes, I know, but I'm not a big git magician and accidentally added a
second commit and it would have been very complicated for me to remove.
From now on I will do the rebase and push -f combo.)

blagasz:
`pyupgrade` and `black` fails but I don't really know from the log what their problem is.

lafrech:
Yes, the log is not verbose enough.

If you're using Python 3.6+ you can install pre-commit to do that locally before pushing.
https://github.com/marshmallow-code/apispec/blob/dev/CONTRIBUTING.rst

lafrech:
Sorry, I was not notified on the force-push.

I'd be tempted to factorize the tests using some obfuscating pytest parametrization trickery, but that's just personal taste/obsession. I can do that later on if I have spare time to waste.

I think this is good to merge.

Thanks.

blagasz:
Ok, as you wish :)

Thanks for your support in this!

lafrech:
I did the rework and released in 1.2.1.

Issue #429, opened 2019-04-09
Support for Related and RelatedList


dpgaspar:
Hi,

Currently I'm using apispec to create components/schemas based on Marshmallow Schemas.
Some of my Schemas are generated by marshmallow-sqlalchemy so they are ModelSchemas.
The problem arises when I have Related or RelatedList "fields" this type identified on OpenApi Schema is "string". Can you give me some advise, or improve this?

Thank you!

dpgaspar:
Any news on this?

andrjohn:
@dpgaspar I have a PR out that may help. At the moment, any field that is not explicitly in a field mapping in openapi.py will default to be a string. You can work around this by using an API to add your custom fields to the mapping. My PR improves the handling when the field isn't in the mapping, for example, if your RelatedList type inherits from List then it will be given the `array` type.

dpgaspar:
oh great! Can you please make the PR reference this issue, maybe I can help further down the road.

dpgaspar:
The problem here is that I'm using marshmallow_sqlalchemy if I add:

``` python
DEFAULT_FIELD_MAPPING = {
    ....
    marshmallow_sqlalchemy.fields.Related: ("number", None),
}
```

"solves" my problem, but this seems wrong since the extension is for marshmallow not marshmallow_sqlalchemy, does this force a new extension?

Also the `Related` type mapped to `number` is not always correct, since primary keys can be of any type.

Can you please help:
- Are there any already made extensions for this?
- A PR with a new ext for marshmallow_sqlalchemy seems ok for you?

andrjohn:
I see that the class `RelatedList` inherits from `fields.List` and `Related` inherits from` fields.Field`. With PR #435, then these will be represented as "array" and "string" respectively. This seems correct for `RelatedList`. For Related fields, as you say, the type will vary.

You can probably work around the issue by creating your own custom field:

    @ma_plugin.map_to_openapi_type(mmfields.Number)
    class RelatedNumber(sqlfields.Related):
        pass

or after #435:

    class RelatedNumber(sqlfields.Related, mmfields.Number):
        pass

I'm not familiar with marshmallow-sqlalchemy, but it might be possible to change the `Related` class to mess with it's own type hierarchy in a way that could work.

dpgaspar:
Nice, thank you @andrjohn, I'll explore these alternatives. I'll close this one, if I need any further help I'll open a new one

Issue #250, opened 2018-07-17
Field that is subclass of `fields.Int` gets `{type: string}` in the generated yaml.


TheGrandmother:
Rolling Python: 3.6.5, marshmallow: 3.0.0b11 and apispec: 0.39.0

I have created a custom field like this:
```python
#custom_fields.py

class Id(fields.Integer):
    def __init__(self, **kw_args):
        if 'validate' not in kw_args or kw_args['validate'] is None:
            kw_args['validate'] = [lambda v: v > 0]
        else:
            kw_args['validate'] = kw_args['validate'].copy().append(
                lambda v: v > 0)
        super(Id, self).__init__(**kw_args)
```

And i then use it like this:
```python
class Datapoint(Schema):
    uuid = custom_fields.Id(required=True)
```

And I then add the definition and generate the specs with `to_yaml()` but then it says:
```yaml
uuid: {type: string}
```

This confuses me since the field works as expected otherwise and it will fail the validation will fail if i pass anything but an integer > 0 to the `uuid` field.

**NOTE**: I know that there is `field.UUID` but I, for some unknown reason, won't use it.

lafrech:
You should register your custom `Field` first.

https://apispec.readthedocs.io/en/stable/api_ext.html#apispec.ext.marshmallow.MarshmallowPlugin.map_to_openapi_type

```python
@ma_plugin.map_to_openapi_type(Integer)  # will map to ('integer', 'int32')
class Id(fields.Integer):
    ...

@ma_plugin.map_to_openapi_type('integer', 'ID')
class Id(fields.Integer):
    ...
```

TheGrandmother:
Thanks!
I had no idea :3

TheGrandmother:
Hmmm....
this does not seem to be working actually.
Using, as per the documentation:

```python
from apispec.ext.marshmallow import MarshmallowPlugin

@MarshmallowPlugin.map_to_openapi_type('integer', 'i32')
class Id(fields.Integer):
    ...
```
I get a: `AttributeError: 'str' object has no attribute 'openapi'` exception.
 With the following partial stack trace:
```
  File "/home/grandmother/git/matomatic3/models/custom_fields.py", line 5, in <module>
    @MarshmallowPlugin.map_to_openapi_type('integer', 'i32')
  File "/home/grandmother/.local/lib/python3.6/site-packages/apispec/ext/marshmallow/__init__.py", line 146, in map_to_openapi_type
[...]

PR #435, opened 2019-04-18
Uses class hierarchy to find type and format properties

### LINKED ISSUE #433: Subclassed Lists marked as string type
I am using a tool to generate a Marshmallow schema that I then present as a RESTful API. In order to override some of the methods of marshmallow.fields.List, the generated schema can have generated classes that subclass List. The resulting APISpec is _almost_ perfect because openapi.py mostly uses isinstance to see if the field is a list. For example:

        is_array = isinstance(
            field, (marshmallow.fields.Nested, marshmallow.fields.List)
        )

However, when working out the field type the mapping is looked up using a concrete type:

        self.field_mapping.get(type(field), ("string", None))

It isn't straightforward for me to add the new types using map_to_openapi_type because of the generated nature of my marshmallow and the way the presentation layer is kept separate. However, instead of falling back to `("string", None)`, we can either: fall back to `("array", None)` for instances of a List (and Nested?) either by explicitly checking for those types in `field2type_and_format`; or walking through the field_mapping (in order), checking for instances of each type.

andrjohn:
When sub-classing a marshmallow type then the type and format properties are not found because those properties are looked up in a mapping using concrete types. Other properties are added using `isinstance` to check if the type is a form or `List` or `Nested`.

This change uses the class hierarchy (using the MRO) to do a better job of finding a match in the field mapping. This will reduce the need for custom classes to be added to the field mapping, although that always remains an option if the user wants to control it.

This resolves #433 and #250
This also seems related to #429

### DIFF
--- a/src/apispec/ext/marshmallow/openapi.py
+++ b/src/apispec/ext/marshmallow/openapi.py
@@ -163,7 +163,20 @@ class OpenAPIConverter(object):
         :param Field field: A marshmallow field.
         :rtype: dict
         """
-        type_, fmt = self.field_mapping.get(type(field), ("string", None))
+        # If this type isn't directly in the field mapping then check the
+        # hierarchy until we find something that does.
+        for field_class in type(field).__mro__:
+            if field_class in self.field_mapping:
+                type_, fmt = self.field_mapping[field_class]
+                break
+        else:
+            warnings.warn(
+                "Field of type {} does not inherit from marshmallow.Field.".format(
+                    type(field)
+                ),
+                UserWarning,
+            )
+            type_, fmt = "string", None
 
         ret = {"type": type_}
 
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -55,3 +55,4 @@ Contributors (chronological)
 - Jonathan Beezley `@jbeezley <https://github.com/jbeezley>`_
 - David Stapleton `@dstape <https://github.com/DStape>`_
 - Szabolcs Blága `@blagasz <https://github.com/blagasz>`_
+- Andrew Johnson `@andrjohn <https://github.com/andrjohn>`_
--- a/tests/test_ext_marshmallow_openapi.py
+++ b/tests/test_ext_marshmallow_openapi.py
@@ -4,7 +4,6 @@ import re
 import warnings
 
 import pytest
-from pytest import mark
 
 from marshmallow import fields, Schema, validate
 
@@ -21,7 +20,7 @@ class TestMarshmallowFieldToOpenAPI:
         field = fields.String(validate=validate.OneOf(choices))
         assert openapi.field2choices(field) == {"enum": choices}
 
-    @mark.parametrize(
+    @pytest.mark.parametrize(
         ("FieldClass", "jsontype"),
         [
             (fields.Integer, "integer"),
@@ -53,7 +52,7 @@ class TestMarshmallowFieldToOpenAPI:
         assert res["type"] == "array"
         assert res["items"] == openapi.field2property(fields.String())
 
-    @mark.parametrize(
+    @pytest.mark.parametrize(
         ("FieldClass", "expected_format"),
         [
             (fields.Integer, "int32"),
--- a/tests/test_ext_marshmallow_openapi.py
+++ b/tests/test_ext_marshmallow_openapi.py
@@ -826,33 +826,31 @@ def test_openapi_tools_validate_v3():
         pytest.fail(str(error))
 
 
-class ValidationSchema(Schema):
-    id = fields.Int(dump_only=True)
-    range = fields.Int(validate=validate.Range(min=1, max=10))
-    multiple_ranges = fields.Int(
-        validate=[
-            validate.Range(min=1),
-            validate.Range(min=3),
-            validate.Range(max=10),
-            validate.Range(max=7),
-        ]
-    )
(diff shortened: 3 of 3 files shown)


andrjohn:
This PR is made up of a number of commits. I tidy up the tests a bit before adding tests for the new behaviour. It is probably easier to review per commit than all in one go and I can drop commits the maintainers aren't happy with.

sloria:
Thanks for the PR! I'll try to get review this within the next few days.

lafrech:
I didn't get the time to review the main modification in this PR, but I approve the 3 first commits already (the tests rework). Thanks for cleaning this up.

dpgaspar:
Will check if it solves #429 on monday

src/apispec/ext/marshmallow/openapi.py
```diff
@@ -163,7 +163,23 @@ def field2type_and_format(self, field):
         :param Field field: A marshmallow field.
         :rtype: dict
         """
-        type_, fmt = self.field_mapping.get(type(field), ("string", None))
+
+        #
+        # If this type isn't directly in the field mapping then check the
+        # heirachy until we find something that does.
```
lafrech: s/heirarchy/hierarchy

lafrech:
Reviewed. Looks good to me !

src/apispec/ext/marshmallow/openapi.py
```diff
@@ -163,7 +163,23 @@ def field2type_and_format(self, field):
         :param Field field: A marshmallow field.
         :rtype: dict
         """
-        type_, fmt = self.field_mapping.get(type(field), ("string", None))
+
+        #
```
sloria:
```suggestion
```

src/apispec/ext/marshmallow/openapi.py
```diff
@@ -163,7 +163,23 @@ def field2type_and_format(self, field):
         :param Field field: A marshmallow field.
         :rtype: dict
         """
-        type_, fmt = self.field_mapping.get(type(field), ("string", None))
+
```
sloria:
```suggestion
```

src/apispec/ext/marshmallow/openapi.py
sloria:
```suggestion
        # hierarchy until we find something that does.
```

src/apispec/ext/marshmallow/openapi.py
```diff
@@ -163,7 +163,23 @@ def field2type_and_format(self, field):
...
+        #
+        # If this type isn't directly in the field mapping then check the
+        # heirachy until we find something that does.
+        #
```
sloria:
```suggestion
```

src/apispec/ext/marshmallow/openapi.py
```diff
@@ -163,7 +163,23 @@ def field2type_and_format(self, field):
...
+        for field_class in type.mro(type(field)):
```
sloria: Not exactly sure of the difference, but perhaps `inspect.getmro` is preferred?

sloria (approved):
Just added a few nits. Since @lafrech has reviewed the substantive changes, this is good to go after the nits are addressed.

src/apispec/ext/marshmallow/openapi.py
lafrech: Why not

```py
    type(field).__mro__
```

Or even 

```py
    field.__class__.__mro__
```

Is it discouraged to access those attributes?

We do it already:

https://github.com/marshmallow-code/apispec/blob/0845c66b67998f8e47b624c10efebe7bd5dbd532/src/apispec/ext/marshmallow/common.py#L120

lafrech: OK, apparently, using `type` is preferred over accessing `__class__` (https://stackoverflow.com/questions/9610993/python-type-or-class-or-is). Unless we're using old style PY2 classes, which shouldn't be the case if inheriting `Field`.

I didn't see any advice against accessing `__mro__` (https://stackoverflow.com/questions/2010692/what-does-mro-do), and I think it is more efficient than calling `mro()` (which does not matter in the context).

sloria: Thanks for doing the research @lafrech. So it sounds like ` type(field).__mro__` is preferred. Let's go with it.

lafrech: Yeah, no big deal, all of these are functional anyway.

I can change `schema.__class__` into `type(schema)` in common.py someday.

andrjohn:
Thanks @lafrech and @sloria for the quick review. I've pushed a new version with the nits fixed.

dpgaspar:
No it does not solve #429

sloria:
This is released in 1.3.0.

@andrjohn I've invited you to have commit bits. No pressure to do any more work than you already have. Just want to open the door to further collaboration. Feel free to manage issues and review PRs as you fit.

lafrech:
Thanks @andrjohn for your work on this.

I'm realizing perhaps the doc about Custom Fields should be updated (https://github.com/marshmallow-code/apispec/blob/8ed68fb305a302a4691cd04742eb2412b9603718/docs/using_plugins.rst#custom-fields):

- Mention the fact the type of the parent class is used.
- Maybe change the 2nd example as it demonstrates a case that is now covered by default. Just don't let the class inherit from `Integer`.

sloria:
Good catch @lafrech. Issue: https://github.com/marshmallow-code/apispec/issues/437

Issue #426, opened 2019-04-08
Stringification of response status codes leads to invalid swagger


DStape:
Recent merge of https://github.com/marshmallow-code/apispec/pull/424 results in our swagger documentation being generated with the following response definitions:

```
"responses": {
          "HTTPStatus.OK": {
```

Previously:
```
"responses": {
          "200": {
```
i.e. usage of https://docs.python.org/3/library/http.html#http-status-codes results in the generation of invalid swagger.

Potential fix (```get_ref()``` in ```/src/apispec/core.py```):
```
for code, response in iteritems(operation["responses"]):
    try:
        status_code = int(code)
    except TypeError, ValueError:
        # log error and skip
     responses[str(status_code)] = get_ref("response", response, openapi_major_version)
```

I'm happy to open a PR for this, but would appreciate any feedback on the approach laid out above (cast the code to an int before casting to str).

Thanks,
David

sloria:
I think the try-except approach makes sense. @lafrech ?

lafrech:
I'm not a big fan of the try-except approach where we try to catch all kinds of errors, potentially hiding input errors. And I wouldn't log an error since this code raises on `'4XX'`, for instance, which is valid as per #422.

We only ever tested/supported status codes as string or integers. AFAIU, `HTTPStatus` worked because it's a subclass of `IntEnum` and it would serialize as integer when json dumped.

Since `HTTPStatus` is part of Python core and is totally relevant here, it would be nice to support it and I think we can do it explicitly:

```py
            for code, response in iteritems(operation["responses"]):
                if not PY2:
                    from http import HTTPStatus
                    if isinstance(code, HTTPStatus):
                        code = code.value
                responses[str(code)] = get_ref(
                    "response", response, openapi_major_version
                )
            operation["responses"] = responses
```

DStape:
Will take your comments on board and will have a PR out soon.

DStape:
PR #436

sloria:
The fix is released in 1.3.1.

PR #436, opened 2019-04-23
Fix stringification of response codes

DStape:
OpenAPI 2 states that the available status codes are described by RFC
7231. The RFC states that the, "status-code element is a three-digit integer
code", therefore we interpret that as OpenAPI 2 does not allow
non-integer characters to be present in status codes.

OpenAPI 3 on the other hand is a lot less ambiguous and states that the
status code field "MAY contain the uppercase wild character X",
therefore meaning non-integer characters are valid (e.g. '2XX').

Also add support for parsing Python 3's in-built http.HTTPStatus.

Addresses issue #426

### DIFF
--- a/src/apispec/core.py
+++ b/src/apispec/core.py
@@ -1,14 +1,18 @@
 # -*- coding: utf-8 -*-
 """Core apispec classes and functions."""
 from collections import OrderedDict
+import warnings
 
-from apispec.compat import iterkeys, iteritems
+from apispec.compat import iterkeys, iteritems, PY2
 from .exceptions import (
     APISpecError,
     PluginMethodNotImplementedError,
     DuplicateComponentNameError,
 )
 from .utils import OpenAPIVersion, deepupdate, COMPONENT_SUBSECTIONS, build_reference
+
+if not PY2:
+    from http import HTTPStatus
 
 VALID_METHODS_OPENAPI_V2 = ["get", "post", "put", "patch", "delete", "head", "options"]
 
@@ -69,6 +73,15 @@ def clean_operations(operations, openapi_major_version):
         if "responses" in operation:
             responses = OrderedDict()
             for code, response in iteritems(operation["responses"]):
+                if openapi_major_version < 3:
+                    try:
+                        code = int(code)
+                    except (TypeError, ValueError):
+                        warnings.warn("Non-integer code not allowed in OpenAPI < 3")
+                elif not PY2:
+                    if isinstance(code, HTTPStatus):
+                        code = code.value
+
                 responses[str(code)] = get_ref(
                     "response", response, openapi_major_version
                 )
--- a/src/apispec/core.py
+++ b/src/apispec/core.py
@@ -3,16 +3,13 @@
 from collections import OrderedDict
 import warnings
 
-from apispec.compat import iterkeys, iteritems, PY2
+from apispec.compat import iterkeys, iteritems
 from .exceptions import (
     APISpecError,
     PluginMethodNotImplementedError,
     DuplicateComponentNameError,
 )
 from .utils import OpenAPIVersion, deepupdate, COMPONENT_SUBSECTIONS, build_reference
-
-if not PY2:
-    from http import HTTPStatus
 
 VALID_METHODS_OPENAPI_V2 = ["get", "post", "put", "patch", "delete", "head", "options"]
 
@@ -73,14 +70,11 @@ def clean_operations(operations, openapi_major_version):
         if "responses" in operation:
             responses = OrderedDict()
             for code, response in iteritems(operation["responses"]):
-                if openapi_major_version < 3:
-                    try:
-                        code = int(code)
-                    except (TypeError, ValueError):
+                try:
+                    code = int(code)  # handles IntEnums like http.HTTPStatus
+                except (TypeError, ValueError):
+                    if openapi_major_version < 3:
                         warnings.warn("Non-integer code not allowed in OpenAPI < 3")
-                elif not PY2:
-                    if isinstance(code, HTTPStatus):
-                        code = code.value
 
                 responses[str(code)] = get_ref(
                     "response", response, openapi_major_version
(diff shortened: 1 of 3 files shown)


lafrech:
Thanks for the PR. Comments inline.

src/apispec/core.py
```diff
@@ -69,6 +73,17 @@ def get_ref(obj_type, obj, openapi_major_version):
         if "responses" in operation:
             responses = OrderedDict()
             for code, response in iteritems(operation["responses"]):
+                if openapi_major_version < 3:
```
lafrech: apispec generally does not try to enforce compliance to OAS when it comes to OAS2 vs. OAS3 differences. In other words, I wouldn't have bothered raising a warning if passing 4XX when using OAS2.

I think the rationale is that it would be too much work and it would add unnecessary complexity to check everything. We only try our best to ensure the user can express what he needs, but we don't put all safeguards to ensure he doesn't mess up.

Now that you did the work, I don't have a strong feeling against keeping your warning, though. Just saying.

tests/test_core.py
```diff
@@ -422,6 +424,69 @@ def test_response(self, spec):
                 == metadata["components"]["responses"]["test_response"]
             )
 
+
+    def test_response_with_HTTPStatus_code(self, spec):
+        if int(sys.version[0]) == 2:
```
lafrech: You should use [`@pytest.mark.skip`](https://docs.pytest.org/en/latest/skipping.html) for that.

tests/test_core.py
```diff
@@ -422,6 +424,69 @@ def test_response(self, spec):
...
+
+        elif python_openapi_combo() == (3, 3):
+            uut()
+
+            assert assertStatusCodeInResponseList() == True
+
+        else:
+            pytest.fail("Did not expect Python{} and OpenAPI {} combination"
+                        .format(sys.version[0], spec.openapi_version.major))
+
```
lafrech, line 453: I don't see what differs between Python versions in this test. If there is no difference, this can be greatly simplified.

src/apispec/core.py
DStape: OK, point taken. If you don't have a strong feeling then I won't change this, but I'd still be happy to do so if your opinion changed (or if anyone else objects the current approach). Removing the check would not cause any problems for my use-case anyway, when I wrote this code I was simply in "follow-the-spec to -a-tee" mode :)

tests/test_core.py
DStape: Agreed. Will fix.

tests/test_core.py
DStape: Slight tweak - using ```@pytest.mark.skipif``` instead.

DStape:
Latest push made 3 changes:

1) now using ```@pytest.mark.skipif```
2) removed test code duplication
3) bool comparisons using ```is``` rather than ```==```

DStape:
Ran through code formatter (```black```). Now expecting CI job to pass...

src/apispec/core.py
lafrech: Yes. Let's wait for @sloria's take on the warning.

Also, if we keep the warning, I think we should still let the invalid code pass through. The user is warned but his intent is respected.

tests/test_core.py
lafrech: I mean I don't think you need to test Python version, do you? Couldn't you just check `spec.openapi_version.major`?

Besides, I prefer to check `spec.openapi_version.major < 3` so that we don't have to change that is it doesn't change in OAS4.

Also, and admittedly this is a matter of taste, I don't think you really need to define `uut` and `assertStatusCodeInResponseList`.

For instance, `assertStatusCodeInResponseList` is called only twice and it would be just as readable to do

```py
is spec.openapi_version.major <3:
    assert status_code not in get_paths(spec)["/pet/{petId}"]["get"]["responses"]
else:
    assert status_code not in get_paths(spec)["/pet/{petId}"]["get"]["responses"]
```

We like DRY but if you look at the other tests, this is how we generally do.

In another context, I'd write

```py
assert status_code in get_paths(spec)["/pet/{petId}"]["get"]["responses"] == spec.openapi_version.major >= 3
```

but I tend to prefer readability and use similar patterns across tests.

Notes:

- If we implement the change I suggest above, the status code is always in and the test can be factored out.

- If we kept the warning, you could also inline `uut` without repetition by always catching the warnings and assert it is emitted only on PY2.

Enough nitpicking. Thanks for your work on this and your reactivity.

src/apispec/core.py
DStape: Updated - the warning remains but the ```continue``` statement has gone, so the user's status code will always be documented.

tests/test_core.py
DStape: My last refactoring removed a bunch of stuff, but I should have tidied up further. The test should be a lot cleaner now.

And no problem, thank you for the quick feedback and review comments :).

DStape:
Latest push:

In OAPI2 case, documenting non-integer code is no longer skipped (but warning message will still be emitted).
```test_response_with_status_code_range()``` updated to account for above change, and it has also been further simplified.

lafrech:
Thanks for the rework. Looks good to me.

About the warning (https://github.com/marshmallow-code/apispec/pull/436#discussion_r277737350), I don't really mind either way. I don't think we should try to enforce compliance as that would be a daunting task, but now that it's already here...

I'm courageously deferring the decision to @sloria.

sloria:
Re: the warning--it's fine in this context. I simplified the code so that `HTTPStatus` objects get cast to an `int` anyway, so the warning doesn't add much more complexity.

sloria (approved):

lafrech:
The try-except approach is what @DStape proposed at first (https://github.com/marshmallow-code/apispec/issues/426). I liked the explicit approach but I don't really mind. Maybe the try-except will cover more cases. Thanks for reworking and merging.

DStape:
Thank you for your time on this @lafrech and @sloria .

PR #438, opened 2019-04-25
tox/pre-commit: show-diff-on-failure

lafrech:
This changes displays the diff in travis log. I picked it from werkzeug tox config: https://github.com/pallets/werkzeug/blob/master/tox.ini#L24.

Useful for people who didn't install pre-commit (occasional contributors, Python 3.5 users,...).

I checked it works by adding a dummy commit with intentional errors.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -15,7 +15,7 @@ commands = pytest {posargs}
 [testenv:lint]
 deps = pre-commit~=1.14
 skip_install = true
-commands = pre-commit run --all-files
+commands = pre-commit run --all-files --show-diff-on-failure
 
 [testenv:docs]
 deps = -rdocs/requirements.txt


sloria:
We used to do this but it was removed in https://github.com/marshmallow-code/apispec/commit/836f294c2260ffb0a9d5771c352da83b38b2822c for some reason ¯\\_(ツ)_/¯  . Happy to add it back if it's useful, though.

lafrech:
Weird. It didn't work on my last build: https://travis-ci.org/marshmallow-code/apispec/jobs/525844418.

I did check it worked before submitting the PR.

(I sometimes find that black, while reformatting, makes choices that sort of hide the intent of the developer, by putting the line breaks at places that don't express the meaning of the code. Perhaps I should capture examples and post them on the bugtracker, but it could be too subjective.)

PR #447, opened 2019-05-08
In resolve_nested_schema(), add newly found nested schema to refs dict

flayman:
In OpenAPIConverter.resolve_nested_schema() of apispec.ext.marshallow.openapi, in the case where the schema key is not found in self.refs, a name is resolved for the nested schema but that name and the schema key are never added to self.refs, causing self.get_ref_dict() to fail on previous line 701 with "KeyError: SchemaKey(SchemaClass=<class '[some.class.Name]'>, ...'. This simple patch correctly adds the dictionary entry for the nested schema to self.refs.

### DIFF
--- a/src/apispec/ext/marshmallow/openapi.py
+++ b/src/apispec/ext/marshmallow/openapi.py
@@ -473,6 +473,8 @@ class OpenAPIConverter(object):
                 return json_schema
             name = get_unique_schema_name(self.spec.components, name)
             self.spec.components.schema(name, schema=schema)
+            # add nested schema name to the refs dict, otherwise get_ref_dict() fails
+            self.refs[schema_key] = name
         return self.get_ref_dict(schema_instance)
 
     def schema2parameters(


Bangertm:
Do you have a test case where this is failing?  The nested schema is added to the spec in the previous line: 
```python
self.spec.components.schema(name, schema=schema)
```

That has the effect of adding the schema to the references dictionary in `MarshmallowPlugin.schema_helper` here: https://github.com/marshmallow-code/apispec/blob/72e4b298f030f00f68e40e27f60a0c3e0c998b8d/src/apispec/ext/marshmallow/__init__.py#L167

Several tests cases are working as expected so I'd like to see what is causing the issue you are seeing.

flayman:
Hi,

I don't have a simple isolated test case I can give you at the moment, but
I do have a failure condition as I described. I instantiate the spec
instance like this:

spec = APISpec(
    title="My lovely title",
    version=api.version,
    openapi_version="2.0",
    info=dict(description="My lovely description")
)

That instantiation is based on an example I found somewhere, which is
probably for an older version of the api. There doesn't seem to be an info
parameter anymore. I do not provide the plugins parameter, which then
defaults to an empty tuple. Because there are no plugins, no helpers are
called in Components.schema(). I am creating the spec object from within
flask-restplus so that I can convert a Marshmallow schema into a JSON Schema
as I've seen described here
<https://github.com/noirbizarre/flask-restplus/issues/438#issuecomment-460350182>.
The method described is old and the OpenAPIConverter class requires some
additional parameters now including spec, so I made a rudimentary spec for
that purpose. APISpec is probably not meant to be instantiated in this way,
but nothing prevents me from doing it. I'd love to know of a more correct
method of achieving the result I'm looking for.

Having said all that, it seems to me that this action is too fundamental to
leave to a plugin helper. There's already a test for whether the key is
found on self.refs, so why not just add it once the schema is resolved?

Bangertm:
In your use case you should initialize `OpenAPIConverter` with a `schema_name_resolver` function that always returns `None`.  

In the typical use case using the Marshmallow Plugin, by default `OpenAPIConverter` will add the nested schema to the spec and the insert a JSON reference object into the spec when it encounters a nested field.  In your case this is not going to make sense.  The reference object is not going to resolve to anything because the call to `self.spec.components.schema` is a no-op.  

By making the `schema_name_resolver` always return `None` you will define your nested schemas in line in the same JSON object, which I think is what you want.

lafrech:
IIUC, you're using `OpenAPIConverter` without using the `MarshmallowPlugin`, which is not really a suported use case.

I'd rather not try to support this.

If you intend to keep the flask-restplus / apispec hybrid, you can indeed use a `schema_name_resolver` as suggested by @Bangertm.

flayman:
Hi all,

Thank you @Bangertm. I'm gaining a better understanding of the architecture. I thought I needed to use a non-trivial resolver, but I'd failed to read the documentation carefully enough. There it is: "_If the schema_name_resolver function returns a value that evaluates to False in a boolean context the nested schema will not be added to the spec and instead defined in-line._" In fact that is exactly what I want. I didn't like having to show nested schema models in the flask-restplus Swagger UI as defined models. It turns out the nicest way to do this using the MarshmallowPlugin and preserving the reference to schema2jsonschema  is as follows:

```
def resolver(schema):
    return None

spec = APISpec(
    title=api.title,
    version=api.version,
    openapi_version="2.0",
    plugins=[MarshmallowPlugin(schema_name_resolver=resolver)],
    info=dict(description=api.description)
)

schema2jsonschema = spec.plugins.pop().openapi.schema2jsonschema
```

Works well. I also devised a way to do it with a non-trivial resolver, but this is not worth the effort. I have some observations about the interplay going on with the various classes. The use of a resolver returning a string prevents the early return in OpenAPIConverter.resolve_nested_schema() on lines 472 or 473, giving rise to the call to self.get_ref_dict(). It is then delegated to the plugin helpers of the apispec.core.Components class through self.spec.components.schema() to add the key to self.refs in order to avoid a fatal KeyError in get_ref_dict(). There is a lot of indirection going on. Is that beneficial?  At least one, and hopefully only one, provided plugin in its implementation of schema_helper() will need to add the key to the refs dict and so there needs to be a plugin provided when name is not None, but there's no guarantee the plugin will do that. On the other hand, it is something that always needs to be done in exactly the same way unless I'm missing something. This is not a concern of Marshmallow and so I don't understand why it is a concern of the MarshmallowPlugin. It seems to be a concern of the OpenAPIConverter class. If it were up to me, then either the OpenAPIConverter class would add the nested schema to its own refs dict or the fetching done by get_ref_dict() would also be delegated to the Components class and handled gracefully. Does that make sense?

Anyway, thanks very much for the help. My own requirements are satisfied. If you want to close this request then I don't have a problem with that.

lafrech:
Closing this. Thanks for the comments. Could be useful when refactoring.

Issue #453, opened 2019-06-10
Allow to pass parameters argument to path


karec:
The OpenAPI specifications allow to set parameters for a path item object https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#path-item-object to make them available for all operations.

AFAIK, [`APISpec.path()`](https://github.com/marshmallow-code/apispec/blob/dev/src/apispec/core.py#L279) method doesn't take a `parameters` argument, resulting in duplicating documentation on every operation of the path for REST APIs.

If I've not missed a something here, I would be happy to work on a PR to add that feature

EDIT: details and spelling

lafrech:
Makes total sense to me.

PR welcome!

karec:
On it !

lafrech:
Closed by #455.

Thank you so much @karec for your work on this.

lafrech:
Should `MarshmallowPlugin` get a `path_helper` to allow passing schemas as path parameters?

Just wondering out loud. I don't see a practical use case.

karec:
Good question, I don't see a practical use case either, since we can still reference schema in our parameter object. Also, `Parameter Object` will require more than just a schema (name and in at least)

PR #455, opened 2019-06-12
Add support for path level parameters

karec:
For #453 

* Create `clean_parameters` function to avoid duplicate code
* Extracted `get_ref` from `clean_operations` to reuse it in
  `clean_parameters`
* Added `parameters` argument to `APISpec.path()`
* Sending `parameters` argument to `plugin.path_helper` as a kwargs,
  allowing plugins to update it
* Added tests for global parameters feature
* Added tests for duplicate parameters in operations
* Added name to existing parameters in tests as required by OpenAPI
  specification
* Now raising an `APISpecError` in case of duplicate parameters in
  path / operations

### DIFF
--- a/docs/writing_plugins.rst
+++ b/docs/writing_plugins.rst
@@ -29,12 +29,62 @@ A plugin with a path helper function may look something like this:
 
 
     class MyPlugin(BasePlugin):
-        def path_helper(self, path, func, **kwargs):
+        def path_helper(self, path, func, parameters, **kwargs):
             """Path helper that parses docstrings for operations. Adds a
             ``func`` parameter to `apispec.APISpec.path`.
             """
             operations = load_operations_from_docstring(func.__doc__)
             return Path(path=path, operations=operations)
+
+
+All helpers take `**kwargs` arguments that are passed to `apispec.APISpec` methods by users and forwarded to plugins methods,
+allowing them to take extra arguments if required.
+
+A plugin with an operations helper that add `deprecated` flag may look like this
+
+.. code-block:: python
+
+   # deprecated_plugin.py
+
+   from apispec import BasePlugin
+   from apispec.compat import iteritems
+   from apispec.yaml_utils import load_operations_from_docstring
+
+
+   class DeprecatedPlugin(BasePlugin):
+       def operation_helper(self, path, operations, **kwargs):
+           """Operation helper that add `deprecated` flag if in `kwargs`
+           """
+           if kwargs.pop("deprecated", False) is True:
+               for key, value in iteritems(operations):
+                   value["deprecated"] = True
+
+
+Using this plugin
+
+
+.. code-block:: python
+
+   import json
+   from apispec import APISpec
+   from deprecated_plugin import DeprecatedPlugin
+
+   spec = APISpec(
+       title="Gisty",
+       version="1.0.0",
+       openapi_version="3.0.2",
+       plugins=[DeprecatedPlugin()],
+   )
+
+   # path will call operation_helper on his operations
+   spec.path(
+       path="/gists/{gist_id}",
+       operations={"get": {"responses": {"200": {"description": "standard response"}}}},
+       deprecated=True,
+   )
+   print(json.dumps(spec.to_dict()["paths"]))
+   # {"/gists/{gist_id}": {"get": {"responses": {"200": {"description": "standard response"}}, "deprecated": true}}}
+
 
 
 The ``init_spec`` Method
@@ -111,5 +161,5 @@ Next Steps
 To learn more about how to write plugins:
 
 * Consult the :doc:`Core API docs <api_core>` for `BasePlugin <apispec.BasePlugin>`
-* View the source for an existing apispec plugin, e.g. `FlaskPlugin <https://github.com/marshmallow-code/apispec-webframeworks/blob/master/apispec_webframeworks/flask.py>`_.
+* View the source for an existing apispec plugin, e.g. `FlaskPlugin <https://github.com/marshmallow-code/apispec-webframeworks/blob/master/src/apispec_webframeworks/flask.py>`_.
 * Check out some projects using apispec: https://github.com/marshmallow-code/apispec/wiki/Ecosystem
--- a/docs/writing_plugins.rst
+++ b/docs/writing_plugins.rst
@@ -29,7 +29,7 @@ A plugin with a path helper function may look something like this:
 
 
     class MyPlugin(BasePlugin):
-        def path_helper(self, path, func, parameters, **kwargs):
+        def path_helper(self, path, func, **kwargs):
             """Path helper that parses docstrings for operations. Adds a
             ``func`` parameter to `apispec.APISpec.path`.
             """
@@ -37,10 +37,9 @@ A plugin with a path helper function may look something like this:
             return Path(path=path, operations=operations)
 
 
-All helpers take `**kwargs` arguments that are passed to `apispec.APISpec` methods by users and forwarded to plugins methods,
-allowing them to take extra arguments if required.
+All plugin helpers must accept extra `**kwargs`, allowing custom plugins to define new arguments if required.
 
-A plugin with an operations helper that add `deprecated` flag may look like this
+A plugin with an operation helper that adds `deprecated` flag may look like this
 
 .. code-block:: python
 
@@ -61,7 +60,6 @@ A plugin with an operations helper that add `deprecated` flag may look like this
 
 
 Using this plugin
-
 
 .. code-block:: python
 
--- a/src/apispec/core.py
+++ b/src/apispec/core.py
@@ -21,7 +21,7 @@ VALID_METHODS = {2: VALID_METHODS_OPENAPI_V2, 3: VALID_METHODS_OPENAPI_V3}
 
 
 def get_ref(obj_type, obj, openapi_major_version):
-    """Return object or rererence
+    """Return object or reference
 
     If obj is a dict, it is assumed to be a complete description and it is returned as is.
     Otherwise, it is assumed to be a reference name as string and the corresponding $ref
@@ -46,24 +46,30 @@ def clean_parameters(parameters, openapi_major_version):
     :param list parameters: List of parameters mapping
     :param int openapi_major_version: The major version of the OpenAPI standard
     """
-    for parameter in parameters:
-        if isinstance(parameter, dict) and parameter.get("in") == "path":
-            parameter["required"] = True
-
-    # OpenAPI Spec 3 and 2 don't allow for duplicated parameters
-    # A unique parameter is defined by a combination of a name and location
-    try:
-        param_objs = [(p["name"], p["in"]) for p in parameters if isinstance(p, dict)]
-    except KeyError as e:
-        raise InvalidParameterError("Missing key {} for parameter".format(e))
-
     seen = set()
-    for p in param_objs:
-        if p in seen:
-            raise DuplicateParameterError(
-                "Duplicate parameter with name {} and location {}".format(p[0], p[1])
+    for parameter in [p for p in parameters if isinstance(p, dict)]:
+
+        # check missing name / location
+        missing_attrs = [attr for attr in ("name", "in") if attr not in parameter]
+        if missing_attrs:
+            raise InvalidParameterError(
+                "Missing keys {} for parameter".format(missing_attrs)
             )
-        seen.add(p)
+
+        # OpenAPI Spec 3 and 2 don't allow for duplicated parameters
+        # A unique parameter is defined by a combination of a name and location
+        unique_key = (parameter["name"], parameter["in"])
+        if unique_key in seen:
+            raise DuplicateParameterError(
+                "Duplicate parameter with name {} and location {}".format(
+                    parameter["name"], parameter["in"]
+                )
+            )
+        seen.add(unique_key)
+
+        # Add "required" attribute to path parameters
+        if parameter["in"] == "path":
+            parameter["required"] = True
 
     return [get_ref("parameter", p, openapi_major_version) for p in parameters]
 
--- a/src/apispec/core.py
+++ b/src/apispec/core.py
@@ -9,6 +9,7 @@ from .exceptions import (
     PluginMethodNotImplementedError,
     DuplicateComponentNameError,
     DuplicateParameterError,
+    InvalidParameterError,
 )
 from .utils import OpenAPIVersion, deepupdate, COMPONENT_SUBSECTIONS, build_reference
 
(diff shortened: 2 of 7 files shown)


lafrech:
Looks pretty good, thanks.

Nitpick: Splitting the changes in two commits (one for the refactor and one for the new feature) would make it easier to review, but not big deal. It is a small changeset.

src/apispec/core.py
```diff
@@ -18,6 +18,55 @@
...
+def clean_parameters(parameters, openapi_major_version):
+    """Ensure that all parameters with "in" equal to "path" are also required
+    as required by the OpenAPI specification, as well as normalizing any
+    references to global parameters and checking for duplicates parameters
+
+    See https ://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#parameterObject.
+
+    :param list parameters: List of parameters mapping
+    :param int openapi_major_version: The major version of the OpenAPI standard
+    """
```
lafrech, line 48: AFAIK, at this stage, a parameter can be either a string (reference) or a dictionary.

Since `name` and `in` are required in a parameter, shouldn't we assume that if it's a `dict`, it contains `name` and `in`?

Just wondering, I didn't check. But this would allow us to simplify this function.

(We could even raise a custom exception if `name` or `in` is missing, rather than let the code crash. That would make debugging easier. This is out of scope and could be done in another PR.)

src/apispec/core.py
```diff
@@ -18,6 +18,55 @@
...
+    # A unique parameter is defined by a combination of a name and location
+    param_objs = [
+        (p["name"], p["in"])
+        for p in parameters
+        if isinstance(p, dict) and all(uk in p for uk in unique_keys)
+    ]
+    seen = set()
+    for p in param_objs:
+        if p in seen:
+            raise APISpecError(
```
lafrech: Please create a custom exception, like `DuplicateParameterError`.

tests/test_core.py
```diff
@@ -399,6 +399,84 @@ def test_parameter(self, spec):
                 == metadata["components"]["parameters"]["test_parameter"]
             )
 
+        spec.path(
```
lafrech, line 419: You could add a comment to explain the purpose of this test, and/or put this in another test method, like

```py
def test_parameter_duplicate(self, spec):
```

tests/test_core.py
```diff
@@ -399,6 +399,84 @@ def test_parameter(self, spec):
...
+        route_spec = self.paths["/pet/{petId}"]["get"]
+
+        spec.components.parameter("test_parameter", "path", route_spec["parameters"][0])
+        spec.path(
+            path=path,
+            operations=dict(put={}, get={}),
+            parameters=[{"name": "petId", "in": "path"}, "test_parameter"],
+        )
+
+        if spec.openapi_version.major < 3:
```
lafrech: I think you can avoid the condition by using `tests.utils.build_ref`.

tests/test_core.py
```diff
@@ -399,6 +399,84 @@ def test_parameter(self, spec):
...
+                {"name": "petId", "in": "path", "required": True},
+                {"$ref": "#/parameters/test_parameter"},
+            ]
+        else:
+            assert get_paths(spec)[path]["parameters"] == [
+                {"name": "petId", "in": "path", "required": True},
+                {"$ref": "#/components/parameters/test_parameter"},
+            ]
+
+        spec.path(
```
lafrech, line 462: Same remark here about comment and separate test, and about `build_ref`.

lafrech (changes requested):

tests/test_ext_marshmallow.py
```diff
@@ -598,7 +602,8 @@ def test_schema_array_uses_ref_if_available_v3(self, spec_fixture):
                 "get": {
                     "parameters": [
                         {
-                            "in": "body",
+                            "name": "Pet",
+                            "in": "query",
```
lafrech, line 606: Why this change and the one above? Did the test fail without them?

src/apispec/core.py
karec: You're right, OpenAPI requires `name` and `in` for a parameter object, so I've created a separate commit for this: d53eb9b but let me know if you prefer to move that commit into a new PR

tests/test_core.py
karec: Fixed in 3c01ee2e, moved tests in another method

tests/test_core.py
karec: Fixed in 3c01ee2e, moved to another method

tests/test_core.py
```diff
@@ -399,6 +404,99 @@ def test_parameter(self, spec):
                 == metadata["components"]["parameters"]["test_parameter"]
             )
 
+    def test_invalid_parameter(self, spec):
```
lafrech: Do you think you could parametrize the test?

```py

@pytest.mark.parametrize('parameters', (
    [{"name": "petId"}],  # Missing "in"
    [{"in": "path"}],  # Missing "name"
))
def test_invalid_parameter(self, spec, parameters):
    with pytest.raises(InvalidParameterError):
        spec.path(
            path=path,
            operations=dict(put={}, get={}),
            parameters=parameters,
        )
```

src/apispec/core.py
lafrech: Same PR is fine. I just didn't want you to feel you had to do it to get this PR merged. Thanks for doing it.

src/apispec/core.py
```diff
@@ -18,6 +20,54 @@
...
+        raise InvalidParameterError("Missing key {} for parameter".format(e))
+
+    seen = set()
+    for p in param_objs:
+        if p in seen:
+            raise DuplicateParameterError(
+                "Duplicate parameter with name {} and location {}".format(p[0], p[1])
+            )
+        seen.add(p)
+
```
lafrech, line 73: I'm pretty sure we could rework this function to merge the loops.

Let's give it a try.

```py
seen = set()
for parameter in [p for p in parameters if isinstance(p, dict)]:

    # Check missing name/location
    missing_attrs = [attr for attr in ("name", "in") if attr not in parameter]
    if missing_attrs:
        raise InvalidParameterError("Missing keys {} for parameter".format(missing_attrs))

    # OpenAPI spec doesn't allow for duplicated parameters
    # A unique parameter is defined by a combination of a name and location
    unique_key = (parameter["name"], parameter["in"])
    if unique_key in seen:
        raise DuplicateParameterError(
            "Duplicate parameter with name {} and location {}".format(parameter["name"], parameter["in"])
        )
    seen.add(unique_key)

    # Add "required" attribute to path parameters
    if parameter["in"] == path:
        parameter["required"] = True
```

Do you like this implementation better?

BTW, we should probably also force `required = True` for path parameters in `Components.parameter` so that it is enforced on parameter references as well. Don't feel like you have to do it in this PR, but if you want to tackle this, no problem.

src/apispec/core.py
```diff
@@ -18,6 +20,54 @@
 VALID_METHODS = {2: VALID_METHODS_OPENAPI_V2, 3: VALID_METHODS_OPENAPI_V3}
 
 
+def get_ref(obj_type, obj, openapi_major_version):
+    """Return object or rererence
```
lafrech: Typo (not yours, copy-pasted).

s/rererence/reference

lafrech (changes requested):

tests/test_ext_marshmallow.py
karec: Yes, since we now require "name" key in parameters test fail and raises an `InvalidParameterError`

lafrech: I think I wrote this comment before you introduced `InvalidParameterError` because I didn't see where the failure would actually happen. Now, the reason for the change is clear.

lafrech:
This PR adds an `InvalidParameterError` that is raised when a parameter misses `name` or `in` attribute.

This change could break applications that don't provide a name or location. Those are applications with broken documentation since name and location are required. But broken doc is not broken app.

Should we consider this a breaking change? Major releases are cheap, webargs is already 5.x, bumping apispec to 2.0 wouldn't be an issue.

@sloria, what do you think?

src/apispec/core.py
karec: Good catch, I've didn't look to the whole picture after latest update but I agree with this implementation, will make the update

karec: done in a2069bd296cef28aa4f51fe169a11ce06be26c44

karec: Enforced `required` in `Components.parameter` in a2069bd296cef28aa4f51fe169a11ce06be26c44

tests/test_core.py
```diff
@@ -399,6 +405,88 @@ def test_parameter(self, spec):
...
+        "parameters",
+        ([{"name": "petId"}], [{"in": "path"}]),  # missing "in"  # missing "name"
+    )
+    def test_invalid_parameter(self, spec, parameters):
+        path = "/pet/{petId}"
+
+        with pytest.raises(InvalidParameterError):
+            spec.path(path=path, operations=dict(put={}, get={}), parameters=parameters)
+
+        with pytest.raises(InvalidParameterError):
```
lafrech: This is a duplicate of lines above, leftover from before parametrization.

src/apispec/core.py
```diff
@@ -287,14 +331,18 @@ def path(
...
         operations = operations or OrderedDict()
+        parameters = parameters or []
 
         # Execute path helpers
         for plugin in self.plugins:
             try:
-                ret = plugin.path_helper(path=path, operations=operations, **kwargs)
+                ret = plugin.path_helper(
+                    path=path, operations=operations, parameters=parameters, **kwargs
+                )
```
lafrech, line 345: `BasePlugin.path_helper` signature and docstring should be updated in apispec/src/apispec/plugin.py

And the fact that `parameters` is passed should be tested in `tests.test_core.TestPlugins.test_plugin_path_helper_is_used`.

lafrech:
Also, the signature of `BasePlugin.path_helper` has changed. Since the former signature includes `**kwargs`, it may not qualify as a breaking change.

lafrech:
@karec, looks like you committed with two different email addresses, one of which is not connected to your GH account. I absolutely don't mind but if you wish to change that, better now than never.

sloria:
> Should we consider this a breaking change? Major releases are cheap, webargs is already 5.x, bumping apispec to 2.0 wouldn't be an issue.

Yes, I've no issue bumping to 2.0. I won't have time to review this PR for at least a week or so, so go ahead and handle this as you please @lafrech

src/apispec/core.py
karec: On this point I think we should:

* Keeping current unit tests (to ensure that plugins doesn't break with this new signature)
* Test that we have access to `parameters` in kwargs with current signature
* Test new signature

What do you think ?

lafrech: I wouldn't bother. I'd just test the new signature.

The point is about breaking custom plugins that wouldn't accept `**kwargs`. Modifying the signature when inheriting is a serious smell, so one shouldn't be doing this and the change this PR introduces is arguably not a breaking one.

Yet, we can be extra-conservative and make it a major release. Before doing so, we should make sure to document the `**kwargs` in the docs about writing plugins and in `BasePlugin` docstring so that we're totally covered for next time.

karec: Sorry for not responding earlier I'm writing doc on that topic before updating this PR

lafrech: No problem. No rush. Been busy as well.

The doc issue is not directly related. Don't feel like you *have* to take care of it. But if you do, that's great. Thanks.

karec: Don't worry, I prefer this PR to be as complete as possible and it's part of the feature after all. I updated tests and documentation

docs/writing_plugins.rst
```diff
@@ -29,14 +29,64 @@ A plugin with a path helper function may look something like this:
 
 
     class MyPlugin(BasePlugin):
-        def path_helper(self, path, func, **kwargs):
+        def path_helper(self, path, func, parameters, **kwargs):
```
lafrech: I wouldn't add `parameters` here.

docs/writing_plugins.rst
```diff
@@ -29,14 +29,64 @@ A plugin with a path helper function may look something like this:
...
             """Path helper that parses docstrings for operations. Adds a
             ``func`` parameter to `apispec.APISpec.path`.
             """
             operations = load_operations_from_docstring(func.__doc__)
             return Path(path=path, operations=operations)
 
 
+All helpers take `**kwargs` arguments that are passed to `apispec.APISpec` methods by users and forwarded to plugins methods,
```
lafrech: Suggestion:

All plugin helpers must accept extra `**kwargs`, allowing custom plugins to define new arguments if required.

docs/writing_plugins.rst
```diff
@@ -29,14 +29,64 @@ A plugin with a path helper function may look something like this:
...
+allowing them to take extra arguments if required.
+
+A plugin with an operations helper that add `deprecated` flag may look like this
```
lafrech: s/operations/operation
s/add/adds

src/apispec/plugin.py
```diff
@@ -45,7 +62,10 @@ def path_helper(self, path=None, operations=None, **kwargs):
     def operation_helper(self, path=None, operations=None, **kwargs):
         """May mutate operations.
 
+
```
lafrech: Extra newline to remove.

src/apispec/plugin.py
```diff
@@ -15,23 +15,40 @@ def init_spec(self, spec):
         """
 
     def schema_helper(self, name, definition, **kwargs):
-        """May return definition as a dict."""
+        """May return definition as a dict.
+
+        :param str name: identifier by which schema may be referenced
```
lafrech: Nitpick: we generally capitalize the first word of the param docstring

```
- :param str name: identifier by which schema may be referenced
+ :param str name: Identifier by which schema may be referenced
```

src/apispec/core.py
lafrech: Thanks @karec for working on this. Just a few minor comments. We're almost done.

docs/writing_plugins.rst
```diff
@@ -29,18 +29,17 @@ A plugin with a path helper function may look something like this:
...
             """Path helper that parses docstrings for operations. Adds a
             ``func`` parameter to `apispec.APISpec.path`.
             """
             operations = load_operations_from_docstring(func.__doc__)
             return Path(path=path, operations=operations)
 
 
-All helpers take `**kwargs` arguments that are passed to `apispec.APISpec` methods by users and forwarded to plugins methods,
-allowing them to take extra arguments if required.
+All plugin helpers must accept extra `**kwargs`, allowing allowing custom plugins to define new arguments if required.
```
lafrech: allowing allowing

PR #456, opened 2019-06-12
Rework MarshmallowPlugin.resolve_parameters

lafrech:
I'm pretty sure we could also get rid of the `and "in" in parameter` as I don't think a parameter without `in` is valid anyway (see https://github.com/marshmallow-code/apispec/pull/455#discussion_r293091574).


lafrech:
Let's keep the `"in"` condition. Missing "in" and "name"  will trigger an exception at the end of the process (`clean_parameters`) once https://github.com/marshmallow-code/apispec/pull/455 is merged but it is too early to raise at this stage, as those could (very theoretically) be added by another plugin working after `MarshmallowPlugin`.

Issue #463, opened 2019-06-24
parameter name and in required for referenced parameter


runtothesun:
Hi, I have just updated to version 2.0.0 of apispec and I've seen a whole load of new error messages `'Missing keys ['name', 'in'] for parameter'` the check I think was added in PR #455. In my code I create a top level parameter using:

`apispec.components.parameter('ParameterSchemaName','path',{
 'name': 'parameterName',
 'schema': {
   'type': 'string',
 },
 'description': '...',
 'required': True
})`

and then reference it in the doc string:

```
parameters:
-   $ref: '#/components/parameters/ParameterSchemaName
```

Now I need to add 'in' and 'name' into the docstring for this parameter but I have already specified the location and name in the top level definition. I see that there has been recent activity in this area so apologies if this is something that is in hand - I just wanted to raise it in case it was overlooked.

lafrech:
What if you pass only the ref name?

```yaml
post:
    requestBody:
        'ParameterSchemaName'
```

or

```yaml
post:
    requestBody:
        ParameterSchemaName
```

(I don't use YAML, I'm not sure about the quotes.)

lafrech:
This is probably something we overlooked. Sorry about that and thanks for reporting.

On second thought, I don't think my suggestion above will make it work. I'm pretty sure, however, that apispec expects only a ref name, not the whole ref component path, as it handles the path itself (including OASv2 / OASv3 differences) but that is another story.

lafrech:
My suggestion might solve your case, I can't tell as I'm not sure what YAML parser does, but there's definitely an issue.

`clean_parameters` can be called several times on the same operation (see `test_path_merges_paths`), so it should be idempotent (or we should change the logic to ensure it is called only once, if that is feasible).

On the first call, string references are changed into dictionaries, and on the second call the function assumes those are not references but parameter descriptions and complains about missing fields.

Changing

```py
    for parameter in [p for p in parameters if isinstance(p, dict)]:
```

into

```py
    for parameter in [p for p in parameters if isinstance(p, dict) and '$ref' not in p]:
```

solves it. Maybe not the most elegant way.

@karec, any thoughts on this?

We might also have to update tests/docs about how to pass references because I think only the ref ID should be passed. Passing a complete ref won't trigger an exception but the reference is broken in the spec IIUC.

lafrech:
[...]

PR #465, opened 2019-06-24
Update CHANGELOG (component refs by ID), docs (upgrading to 2.0.0)

lafrech:
Fix docs. See https://github.com/marshmallow-code/apispec/issues/463.

Maybe we could also add a YAML example in the "upgrading to 2.0.0" section.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -269,10 +269,7 @@ class TestPath:
                     }
                 ],
                 "responses": {
-                    "200": {
-                        "schema": {"$ref": "#/definitions/Pet"},
-                        "description": "successful operation",
-                    },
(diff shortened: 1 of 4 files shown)


tests/test_core.py
```diff
@@ -270,7 +270,7 @@ class TestPath:
                 ],
                 "responses": {
                     "200": {
-                        "schema": {"$ref": "#/definitions/Pet"},
+                        "schema": {"Pet"},
```
sloria:
```suggestion
                        "schema": "Pet",
```

sloria (approved):
Looks good

tests/test_core.py
lafrech: Good catch. Fixed.

PR #468, opened 2019-07-04
Switch to Azure Pipelines

sloria:
NOTE: This removes builds against apispec-webframeworks, as there's no way in AP to allow failures yet (https://github.com/Microsoft/azure-pipelines-tasks/issues/9302)


sloria:
The `allow_failures` in TravisCI was always a half-baked solution for us anyway, since it wasn't apparent when the `webframeworks` build failed unless you happen to look at the build details.

I think we can get most of the way by running the `apispec-webframeworks` pipeline on a regular schedule (e.g. nightly).

@lafrech Let me know if you have any issues with this.

sloria:
Current build failures are expected. I'll fix the marshmallow 3 compatibility in a separate PR.

PR #486, opened 2019-08-23
Drop official support for 3.0 pre-releases

### LINKED ISSUE #485: Drop marshmallow 3.0.0 RC releases compatibility
https://github.com/marshmallow-code/apispec/blob/78b14803dc966a2c41413f7e08aadcecc193b0b3/src/apispec/ext/marshmallow/openapi.py#L432-L442

MA3 stable is out, we don't need to support older RC versions.

We could rework this in next release.

Leaving it for now to avoid unnecessary conflicts with #478.

sloria:
Close #485

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,11 @@ Features:
   ``Schema`` class. This allows a custom resolver to generate different names
   depending on schema modifiers used in a ``Schema`` instance. (:pr:`476`)
 
+Other changes:
+
+- Drop support for marshmallow pre-releases. Only stable 2.x and 3.x versions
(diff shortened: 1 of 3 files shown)


lafrech:
I'd also add a 

```py
if MARSHMALLOW_VERSION_INFO[0] >= 3:
```

condition just above, rather than

```py
if hasattr(field, "inner")
```

I'd rather wait until https://github.com/marshmallow-code/apispec/pull/478 is merged to avoid conflicts there. I hope to review and merge it soon.

sloria:
Sounds good; I'll update this after you merge that.

PR #488, opened 2019-08-31
Change OpenAPIConverter into OpenAPIConverterMixin

lafrech:
What I suggested in https://github.com/marshmallow-code/apispec/pull/478#issuecomment-526732974.

This makes it easier to subclass `MarshmallowPlugin`.

### DIFF
--- a/src/apispec/ext/marshmallow/__init__.py
+++ b/src/apispec/ext/marshmallow/__init__.py
@@ -50,7 +50,7 @@ import warnings
 
 from apispec import BasePlugin
 from .common import resolve_schema_instance, make_schema_key, resolve_schema_cls
-from .openapi import OpenAPIConverter
+from .openapi import OpenAPIConverterMixin
 
 
 def resolver(schema):
@@ -63,7 +63,7 @@ def resolver(schema):
     return name
 
 
-class MarshmallowPlugin(BasePlugin):
+class MarshmallowPlugin(OpenAPIConverterMixin, BasePlugin):
     """APISpec plugin handling marshmallow schemas
 
     :param callable schema_name_resolver: Callable to generate the schema definition name.
@@ -85,17 +85,14 @@ class MarshmallowPlugin(BasePlugin):
         self.schema_name_resolver = schema_name_resolver or resolver
         self.spec = None
         self.openapi_version = None
-        self.openapi = None
+        self.init_attribute_functions()
+        # Schema references
+        self.refs = {}
 
     def init_spec(self, spec):
         super().init_spec(spec)
         self.spec = spec
         self.openapi_version = spec.openapi_version
-        self.openapi = OpenAPIConverter(
-            openapi_version=spec.openapi_version,
-            schema_name_resolver=self.schema_name_resolver,
-            spec=spec,
-        )
 
     def resolve_parameters(self, parameters):
         resolved = []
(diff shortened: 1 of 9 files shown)


sloria (approved):
I'm good with this. It makes `MarshmallowPlugin` quite fat, but I think it's worth it for the ease of extensibility.

sloria:
An alternative approach would be to set the converter as a class member on `MarshmallowPlugin`.

```python
    Converter = OpenAPIConverter
```

The plugin can be extended like so:

```python
    class Converter(MarshmallowPlugin.Converter):
        def property2parameter(self, ...):
            ...
```

wdyt?

lafrech:
Thanks for the doc update.

Mixin vs. class member is a matter of taste. Both work.

With class member, subclassing just takes one more line. Neglectable.

I have no issue with the plugin class being fat since the mixin is in another file, like if it was another class.

(I don't know if you ever looked at flask-rest-api. I separated the features using mixins. I'm not sure those are real mixins like mixins are meant for, it's hard because the features are not strictly independent. But it makes for a nice split of the code.)

Perhaps the mixin here makes sense because it avoids passing the schema_name_resolver from the plugin to the converter, or fetching the refs in the converter from the plugin, and so on. The separation is not clear enough to do things nicely with a separate class.

sloria:
I have no strong preference. Things can get confusing when inheritance chains get long with lots of mixins (at work we're considering [running with the nested class approach in our REST framework](https://github.com/4Catalyzer/flask-resty/issues/238)), but that's not a problem here.

> (I don't know if you ever looked at flask-rest-api. I separated the features using mixins.

werkzeug organizes things this way, too: https://github.com/pallets/werkzeug/blob/9844e0447e63366273c8b9f9f2400e19b45532d1/src/werkzeug/wrappers/request.py#L9-L16 . It's a nice way to split up the code.

So long as you're OK with publicizing all these methods, let's go ahead with this.

src/apispec/ext/marshmallow/__init__.py
```diff
@@ -53,21 +53,20 @@ class UserSchema(Schema):
...
     schema_cls = resolve_schema_cls(schema)
     name = schema_cls.__name__
     if name.endswith("Schema"):
         return name[:-6] or name
     return name
 
 
-class MarshmallowPlugin(BasePlugin):
-    """APISpec plugin handling marshmallow schemas
+class MarshmallowPlugin(BasePlugin, OpenAPIConverterMixin):
```
sloria: By convention, we put mixins before the base class because we usually want their methods to take precedence.

```suggestion
class MarshmallowPlugin(OpenAPIConverterMixin, BasePlugin):
```

lafrech: Right. Fixed.

lafrech:
> So long as you're OK with publicizing all these methods, let's go ahead with this.

The point is to make subclassing easier. I suppose allowing users to subclass means we're considering it public, either way (be it mixin or class member).

Should we remove the docstrings in openapi.py and field_converter.py saying the module is private API?

sloria:
I guess the class member way makes the methods "semi-private", as in you can still override them if you know what you're doing, but the converter class itself can remain documented as private/subject to change.

> Should we remove the docstrings in openapi.py and field_converter.py saying the module is private API?

We could just remove `:inherited-members:` from the `apispec.ext.marshmallow` autodoc in `api_ext.py`. Your call on this.

lafrech:
Closing in favor of #493.

PR #489, opened 2019-08-31
Make clean_parameters and clean_operations APISpec methods

lafrech:
Alternative to https://github.com/marshmallow-code/apispec/pull/466.

### DIFF
--- a/src/apispec/core.py
+++ b/src/apispec/core.py
@@ -21,101 +21,6 @@ VALID_METHODS_OPENAPI_V3 = VALID_METHODS_OPENAPI_V2 + ["trace"]
 VALID_METHODS = {2: VALID_METHODS_OPENAPI_V2, 3: VALID_METHODS_OPENAPI_V3}
 
 
-def get_ref(obj_type, obj, openapi_major_version):
-    """Return object or reference
-
-    If obj is a dict, it is assumed to be a complete description and it is returned as is.
-    Otherwise, it is assumed to be a reference name as string and the corresponding $ref
-    string is returned.
-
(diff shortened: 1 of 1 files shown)


src/apispec/core.py
```diff
@@ -360,14 +265,108 @@ def path(
...
         if description is not None:
             self._paths[path]["description"] = description
         if parameters:
-            parameters = clean_parameters(parameters, self.openapi_version.major)
+            parameters = self.clean_parameters(parameters, self.openapi_version.major)
             self._paths[path]["parameters"] = parameters
         return self
+
+    @staticmethod
+    def get_ref(obj_type, obj, openapi_major_version):
```
sloria: Can you make this an instance method and get ride of the `openapi_major_version` arg?

src/apispec/core.py
```diff
@@ -360,14 +265,108 @@ def path(
...
+
+        :param str obj_type: "parameter" or "response"
+        :param dict|str obj: parameter or response in dict form or as ref_id string
+        :param int openapi_major_version: The major version of the OpenAPI standard
+        """
+        if isinstance(obj, dict):
+            return obj
+        return build_reference(obj_type, openapi_major_version, obj)
+
+    def clean_parameters(self, parameters, openapi_major_version):
```
sloria: Same here: get ride of `openapi_major_version`?

sloria (approved):

src/apispec/core.py
lafrech: Good catch. Done.

Issue #342, opened 2018-11-12
default/missing value are passed in deserialized form with marshmallow 3


lafrech:
This will break automatic default documentation:

```py
    def field2property(self, field, use_refs=True, dump=True, load=True, name=None):
        [...]
        if 'doc_default' in field.metadata:
            ret['default'] = field.metadata['doc_default']
        else:
            default = field.missing
            if default is not marshmallow.missing and not callable(default):
                ret['default'] = default
```

Since we have the field object, maybe we could use it to dump the value for the doc?

Could that fail in some corner case?

lafrech:
Dumping the `missing` value could be wrong if the field does not round-trip.

In the general case, it's fine, but could there be example of fields where dump(load(value)) != value?

I can't think of a simple example, but it could happen in a custom field, or in a `Nested` field where the schema has complicated logic with pre/post load decorators.

It could also be wrong if a field behaves differently depending on the context. Again, `Nested` is the typical example for this as a schema can embed any sort of processing.

Also, a `Nested` field with a `missing` value of `{}` and a schema with `default` values in fields would generate a wrong documentation if we call nested_field.dump(`{}`).

As soon as we call `dump`, we're more or less in the callable case, and the safest thing to do would be to do nothing.

OTOH, in most cases, dumping `missing` is perfectly fine, so removing automatic default documentation would be a regression. Expecting the user to specify `doc_default` negates the benefits of the change in MA3 allowing to pass values in object form.

For now, we probably need to at least inhibit the feature in MA3 to avoid values in object form (potentially not json-serializable) being injected in the spec dict without being serialized.

lafrech:
Fixed by #490.

PR #490, opened 2019-08-31
Fix documentation of default from missing with marshmallow 3

lafrech:
Addresses #342.

It sucks not being able to document default values automatically, especially since it would probably work most of the times, but I don't see a safe way to do it.

Without this feature removal, we may pass to the spec dict objects that are not json serializable, which will trigger an exception when serializing the spec file.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,12 @@ Features:
   ``Schema`` class. This allows a custom resolver to generate different names
   depending on schema modifiers used in a ``Schema`` instance. (:pr:`476`)
 
+Bug fixes:
+
+- MarshmallowPlugin: With marshmallow 3, the default value of a field in the
+  documentation is the dump of the ``missing`` attribute, not ``missing``
+  itself (:pr:`490`).
+
 Other changes:
 
 - Drop support for Python 2 (:issue:`491`). Thanks :user:`hugovk` for the PR.
--- a/src/apispec/ext/marshmallow/__init__.py
+++ b/src/apispec/ext/marshmallow/__init__.py
@@ -1,4 +1,4 @@
-"""Marshmallow plugin for apispec. Allows passing a marshmallow
+"""marshmallow plugin for apispec. Allows passing a marshmallow
 `Schema` to `spec.components.schema <apispec.core.Components.schema>`,
 `spec.components.parameter <apispec.core.Components.parameter>`,
 `spec.components.response <apispec.core.Components.response>`
@@ -10,11 +10,10 @@ Requires marshmallow>=2.15.2.
 ``MarshmallowPlugin`` maps marshmallow ``Field`` classes with OpenAPI types and
 formats.
 
-It inspects field attributes to create corresponding OpenAPI documentation,
-allowing to automatically document properties such as read/write-only, range or
-length constraints, etc.
+It inspects field attributes to automatically document properties
+such as read/write-only, range and length constraints, etc.
 
-OpenAPI properties can also be passed as metadata in the ``Field`` instance
+OpenAPI properties can also be passed as metadata to the ``Field`` instance
 if they can't be inferred from the field attributes (`description`,...), or to
 override automatic documentation (`readOnly`,...). A metadata attribute is used
 in the documentation either if it is a valid OpenAPI property, or if it starts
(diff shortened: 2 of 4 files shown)


sloria:
> OTOH, in most cases, dumping missing is perfectly fine, so removing automatic default documentation would be a regression.

This. As a user, I want apispec to work mostly transparently, i.e. without having to modify my app code. 

Maybe we could handle errors better to lead the user toward using `doc_default` when it's necessary, but I don't think we should degrade the 90% use case.

lafrech:
Well, we could try to dump the `missing` value with the `Field`. Let's give it a try.

I don't know if we should document the `missing` value of `Nested` fields or container fields with `Nested` inside. The examples I found for default values are "scalar" values like integers or strings but I'm not sure a dict is allowed. Anyway, that's nothing new. Same applies to MA2.

lafrech:
I just pushed a new commit to serialize `missing` with the field.

It has a few smells:

- Accesses underscored `_serialize` method (no big deal).
- Can't pass `attr` and `obj` parameters of `_serialize`. This shouldn't be an issue in most cases as those are almost never used. In fact, the only uses I know of are `Function` and `Method` (I already commented about this in https://github.com/marshmallow-code/marshmallow/issues/1046) and those are dump_only.

We should probably only document missing for fields that are not dump_only. It doesn't make sense anyway, does it?

Also, I'd need to add proper tests to show it works with a field where this matters, such as `DateTime`.

lafrech:
Not quite.

Indeed we shouldn't document default on a `dump_only` field.

However, `Function` and `Method` may not be `dump_only` and in this case, the code is likely to crash. But maybe not, if the method can take `None` as a parameter.

> Maybe we could handle errors better to lead the user toward using `doc_default` when it's necessary

So shall we call `_serialize`, catch `Exception` and return a message telling the user to use `doc_default`? The result may even be wrong without an exception being raised, so this wouldn't be 100% sure. But admittedly, those would be corner cases. Perhaps a note in the docs would be sufficient.

Or rather than passing `None`, could we pass some sort of proxy object as `attr` and `obj` that would raise an exception when used? For `obj`, we could override getter methods, etc. But for `attr`...

sloria:
For now, I think documentation may be the best course.

lafrech:
I added a piece of doc about `MarshmallowPlugin`.

Feedback welcome.

> Indeed we shouldn't document default on a `dump_only` field.

Maybe not but a field with `dump_only` and `missing` is wrong in marshmallow anyway, and we generally don't try to validate user inputs in apispec, so I removed the corresponding case from this PR.

Note: `default` in `_VALID_PROPERTIES` is useless because it can't be in `metadata` as it is a reserved marshmallow attribute. I thought of removing it from the list but things might change in https://github.com/marshmallow-code/marshmallow/issues/1350 (MA 4) so I left it there for now, it is harmless.

sloria (approved):

sloria:
Go ahead and merge when you think this is ready.

lafrech:
@sloria, @bangertm, anything else you'd like to add to 3.0 ?

sloria:
nothing from me

Bangertm:
Been checked out the past few days, but I like the recent changes. The only other thing I have been thinking about is extracting the logic for finding Schemas in OpenAPI objects from `MarshmallowPlugin` and/or improving the documentation of those methods. 

`resolve_parameters`, `resolve_schema_in_request_body`, `resolve_schema` and `OpenAPIConverter.resolve_schema_dict` all have to do with searching for schemas within various OpenAPI objects and then converting them via `OpenAPIConverter`. It might be convenient if all this logic was implemented in a mixin class so that `MarshmallowPlugin` was left with just the plugin methods and a couple of helpers. 

It would also be convenient if the documentation for these methods was a little more clear about the expected structure of the input and where schemas could go. Part of the rational for pulling all the logic out would be to improve documentation and understandably, but we could also just improve the documentation of the methods as they are now.

sloria:
> resolve_parameters, resolve_schema_in_request_body, resolve_schema and OpenAPIConverter.resolve_schema_dict all have to do with searching for schemas within various OpenAPI objects and then converting them via OpenAPIConverter. It might be convenient if all this logic was implemented in a mixin class so that MarshmallowPlugin was left with just the plugin methods and a couple of helpers.

To follow the pattern in https://github.com/marshmallow-code/apispec/pull/493, we could do

```python
    SchemaResolver = SchemaResolver

    class SchemaResolver(MarshmallowPlugin.SchemaResolver):
        def resolve_parameters(...):
            ...
```

PR #493, opened 2019-09-02
Make OpenAPIConverter class a class member of MarshmallowPlugin

lafrech:
Discussed in https://github.com/marshmallow-code/apispec/pull/488.

@sloria, I cherry-picked your docs revisions.

openapi and field_converter are private API, but I left their section in the docs when rebasing your doc revisions commit. In other words, this leaves the docs in the same state as on `dev` branch.

### DIFF
--- a/src/apispec/ext/marshmallow/__init__.py
+++ b/src/apispec/ext/marshmallow/__init__.py
@@ -54,8 +54,7 @@ from .openapi import OpenAPIConverter
 
 
 def resolver(schema):
-    """Default implementation of a schema name resolver function
-    """
+    """Default schema name resolver function that strips 'Schema' from the end of the class name."""
     schema_cls = resolve_schema_cls(schema)
     name = schema_cls.__name__
     if name.endswith("Schema"):
@@ -64,7 +63,7 @@ def resolver(schema):
 
 
 class MarshmallowPlugin(BasePlugin):
-    """APISpec plugin handling marshmallow schemas
+    """APISpec plugin for translating marshmallow schemas to OpenAPI/JSONSchema format.
 
     :param callable schema_name_resolver: Callable to generate the schema definition name.
         Receives the `Schema` class and returns the name to be used in refs within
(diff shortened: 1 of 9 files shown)


src/apispec/ext/marshmallow/__init__.py
```diff
@@ -91,7 +93,7 @@ def init_spec(self, spec):
         super().init_spec(spec)
         self.spec = spec
         self.openapi_version = spec.openapi_version
-        self.openapi = OpenAPIConverter(
+        self.openapi = self.Converter(
```
sloria: Maybe rename this to `self.converter`? No big deal, though.

sloria (approved):

src/apispec/ext/marshmallow/__init__.py
lafrech: I picked it from your comment here: https://github.com/marshmallow-code/apispec/pull/488#issuecomment-526842419

I figured capitalized could make sense since it is a class.

Otherwise, you'd subclass it like this:

```py
    class converter(MarshmallowPlugin.converter):  # <-- lowercase
        def property2parameter(self, ...):
            ...
```

src/apispec/ext/marshmallow/__init__.py
```diff
@@ -91,7 +92,7 @@ def init_spec(self, spec):
...
```
sloria: I meant the instance attribute:

```suggestion
        self.converter = self.Converter(
```

lafrech: Oh, sure...

lafrech: I don't care either way.

I pushed a new commit with this change.

sloria: Looks good. Go ahead and merge if you’re happy with this

lafrech:
Apparently I had screwed up with the openapi and field_converter docs when rebasing. I put them back. The section still has the "Private API" warning.

sloria:
I had removed `:members:` from the openapi and field_converter docs in your previous PR because it made more sense to document the methods on `MarshmallowPlugin` (which is why I added `:inherited-members:`. With the class member approach, it makes sense to keep `:members:` (as you did).

lafrech:
Yeah I noticed that, undid your changes purposely, then screwed up and removed them again by error.

PR #495, opened 2019-09-04
 Any Type Fields

### LINKED ISSUE #395: Representing openapi's "Any Type"
I am trying to use `marshmallow.fields.Raw` to represent a "arbitrary json" field. Currently I do not see a way to represent a field in marshmallow that will produce a openapi property that is empty when passed through apispec.

The code in `apispec.marshmallow.openapi.OpenAPIConverter.field2type_and_format` (see https://github.com/marshmallow-code/apispec/blob/dev/apispec/ext/marshmallow/openapi.py#L159-L173) unconditionally sets the key `type` in the property object. However it seems that it is legal in openapi to have an empty schema (without the `type` key) which permits any type (see the "Any Type" section at the bottom of https://swagger.io/docs/specification/data-models/data-types/)

Please advise if this is possible with the current codebase or if I will need to implement a custom solution. Thanks.

Bangertm:
changes Marshmallow fields.Raw and fields.Field to OpenAPI Any Type fields
closes #395

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,9 @@ Changelog
 
 Features:
 
+- *Backwards-incompatible* Marshmallow ``fields.Raw`` and ``fields.Field`` are
+  now represented by OpenAPI `Any Type <https://swagger.io/docs/specification/data-models/data-types/#any>`_.
+  (:pr:`495`)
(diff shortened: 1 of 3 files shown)


lafrech (approved):

lafrech:
Thanks!

PR #497, opened 2019-09-06
FieldConverterMixin: inject field as kwarg in field2property

lafrech:
This change allow to write a custom field with `self` as first argument.

Because `field2property` injects `self` as `kwarg`, this is only possible if `field` is a `kwarg` too.

IOT, without the change, one would have to write:

```py
    def custom_string2properties(field, self, **kwargs):  # <- hurts the eye
        ret = {}
        if isinstance(field, CustomStringField):
            if self.openapi_version.major == 2:
                ret["x-customString"] = True
            else:
                ret["x-customString"] = False
        return ret
```


lafrech:
Replaced by #498.

PR #499, opened 2019-09-11
Remove unused defaults in OpenAPIConverter methods

lafrech:
Those default values are never used because the methods are always called with a value.

Note: We're reworking webargs to remove the `location` field metadata. Ultimately, it won't be parsed in apispec either, and `default_in` will be a mandatory location.

### DIFF
--- a/src/apispec/ext/marshmallow/openapi.py
+++ b/src/apispec/ext/marshmallow/openapi.py
@@ -142,7 +142,7 @@ class OpenAPIConverter(FieldConverterMixin):
 
         return self.fields2parameters(fields, default_in=default_in)
 
-    def fields2parameters(self, fields, default_in):
+    def fields2parameters(self, fields, *, default_in):
         """Return an array of OpenAPI parameters given a mapping between field names and
         :class:`Field <marshmallow.Field>` objects. If `default_in` is "body", then return an array
         of a single parameter; else return an array of a parameter for each included field in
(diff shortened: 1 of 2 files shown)


src/apispec/ext/marshmallow/openapi.py
```diff
@@ -142,7 +142,7 @@ def schema2parameters(
 
         return self.fields2parameters(fields, default_in=default_in)
 
-    def fields2parameters(self, fields, default_in="body"):
+    def fields2parameters(self, fields, default_in):
```
sloria:
```suggestion
    def fields2parameters(self, fields, *, default_in):
```

src/apispec/ext/marshmallow/openapi.py
```diff
@@ -183,7 +183,7 @@ def fields2parameters(self, fields, default_in="body"):
                 parameters.append(param)
         return parameters
 
-    def field2parameter(self, field, name="body", default_in="body"):
+    def field2parameter(self, field, name, default_in):
```
sloria:
```suggestion
    def field2parameter(self, field, *, name, default_in):
```

src/apispec/ext/marshmallow/openapi.py
```diff
@@ -200,15 +200,7 @@ def field2parameter(self, field, name="body", default_in="body"):
...
-    def property2parameter(
-        self,
-        prop,
-        name="body",
-        required=False,
-        multiple=False,
-        location=None,
-        default_in="body",
-    ):
+    def property2parameter(self, prop, name, required, multiple, location, default_in):
```
sloria:
```suggestion
    def property2parameter(self, prop, *, name, required, multiple, location, default_in):
```

sloria (approved):
With the keyword-only tweaks, this is good to go

sloria:
@lafrech We probably should do a general kw-only arguments audit before cutting 6.0. Do you want to do that?

lafrech:
That's 3.0 (apispec, not webargs).

Yeah, we should. I'll see when I can do that.

sloria:
Ah, 3.0 right. OK, thanks

PR #501, opened 2019-09-12
Use kw-only arguments where sensible

lafrech:


sloria (approved):

src/apispec/ext/marshmallow/openapi.py
```diff
@@ -109,7 +109,13 @@ def resolve_nested_schema(self, schema):
         return self.get_ref_dict(schema_instance)
 
     def schema2parameters(
-        self, schema, default_in="body", name="body", required=False, description=None
```
lafrech, line 112: Good catch. I wonder how I missed this one.

PR #504, opened 2019-09-22
Docs: rework example

lafrech:
- schema registration is optional
- what's the need for "test_request_context"?

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -65,9 +65,7 @@ Example Application
         return PetSchema().dump(pet)
 
 
-    # Register entities and paths
-    spec.components.schema("Category", schema=CategorySchema)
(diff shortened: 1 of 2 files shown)


sloria:
Does this work? I thought the app context was necessary for the flask plugin’s path helper.

lafrech:
I just checked and you're right.

Reworked to just remove the useless schema registrations.

sloria (approved):
The same change should also be made in the README. Otherwise gtg

docs/index.rst
```diff
@@ -66,8 +66,6 @@ Example Application
 
 
     # Register entities and paths
```
sloria:
```suggestion
    # Register the path and the entities within it
```

PR #510, opened 2019-10-21
Test against Python 3.8

lafrech:
That's a lot of combinations and wasted computer cycles.

On flask-smorest, I test MA2 and MA3 on oldest and latest PY and only MA3 on the others. I'd like to do the same here:

```
      - lint

      - py35-marshmallow2
      - py35-marshmallow3

      - py36-marshmallow3

      - py37-marshmallow3

      - py38-marshmallow2
      - py38-marshmallow3

      - py38-marshmallowdev
```

@sloria?

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -26,16 +26,18 @@ jobs:
   parameters:
     toxenvs:
       - lint
+
       - py35-marshmallow2
       - py35-marshmallow3
 
-      - py36-marshmallow2
       - py36-marshmallow3
 
-      - py37-marshmallow2
       - py37-marshmallow3
 
-      - py37-marshmallowdev
+      - py38-marshmallow2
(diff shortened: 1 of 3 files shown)


sloria:
>  I test MA2 and MA3 on oldest and latest PY and only MA3 on the others. I'd like to do the same here:

Fine with me. fwiw, Azure Pipelines gets us 10 concurrent builds, so as long as we don't go over 10 tox envs, the build times shouldn't get much slower.

sloria (approved):

lafrech:
Yeah, I don't mind the build time that much.

It's more about having computers in the cloud wasting energy for useless tests, just because it was easy and free to test on all possible combinations. I try to stick to a reasonable number of combinations.

sloria:
Sounds reasonable 👍

PR #513, opened 2019-11-01
Update main example so that it can run smoothly

bagerard:
I recently looked into this project and was interested to play around with flask so I copy paste the main example but got an error since `get_random_pet` isn't defined. Perhaps it was done on purpose to avoid bloating the example with unnecessary details but I believe it would be worth providing an example that works out of the box

Next thing I updated is the endpoint spec, the response object requires a `description` to be valid (https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#response-object)

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -19,8 +19,7 @@ Example Application
 
 .. code-block:: python
 
-    import random
-    import string
+    import uuid
 
     from apispec import APISpec
     from apispec.ext.marshmallow import MarshmallowPlugin
@@ -44,18 +43,8 @@ Example Application
 
 
     class PetSchema(Schema):
-        category = fields.List(fields.Nested(CategorySchema))
+        categories = fields.List(fields.Nested(CategorySchema))
         name = fields.Str()
-
-
-    def get_random_pet():
-        def _random_str():
-            return "".join(random.choices(string.ascii_uppercase, k=10))
-
-        return {
-            "name": _random_str(),
-            "category": [{"id": random.randint(1, 10000), "name": _random_str()}],
-        }
 
 
     # Optional Flask support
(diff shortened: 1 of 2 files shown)


lafrech:
> Perhaps it was done on purpose to avoid bloating the example with unnecessary details

Indeed, I'm pretty sure the `get_random_pet` was mean to be seen as a placeholder.

>  I believe it would be worth providing an example that works out of the box

I don't have a strong opinion but I'm not convinced about that. We could also make things more obvious by adding something like

    # Import model class
    from .model import Pet

and in the view func

    pet = Pet.random()

But current wording makes things obvious enough to me.

If we need an example that works out of the box, we may want to add an examples directory like in webargs.

I approve the other changes. Thanks for that.

We should also rename category into categories.

bagerard:
Hi @lafrech , thanks for the review.
I've renamed "category" into "categories" and I also got rid of the `get_random_pet` function, I replaced it with a simple dict in the controller, IMO it doesn't complicate the code that much and makes it fully runnable. Let me know what you think of it, if you'r still not convinced, I'll switch it to 
```
from .model import Pet
...
pet = Pet.random()
```

lafrech:
I'm fine with this change.

Nice reordering of the spec output. Perhaps you could use the same order on both examples (components before paths).

Also, please add yourself to AUTHORS.rst.

docs/index.rst
```diff
@@ -57,12 +59,16 @@ Example Application
...
                 application/json:
                   schema: PetSchema
         """
-        pet = get_random_pet()
-        return PetSchema().dump(pet)
+        pet_data = {
+            "name": "sample_pet_" + str(uuid.uuid1()),
+            "categories": [{"id": 1, "name": "sample_category"}],
+        }
+        return PetSchema().dump(hardcoded_pet)
```
lafrech: Oops. Inconsistency.

Proposal:

```py
        # Hardcoded example data
        pet_data = {
            "name": "sample_pet_" + str(uuid.uuid1()),
            "categories": [{"id": 1, "name": "sample_category"}],
        }
        return PetSchema().dump(pet_data)
```

bagerard: damn... sorry for that. It should be fine now, I've re-tested it to make sure its running file

lafrech:
Thanks.

PR #515, opened 2019-11-03
Add component.example

codeasashu:
Added tests
Completes a #245 Todo

### DIFF
--- a/src/apispec/core.py
+++ b/src/apispec/core.py
@@ -30,12 +30,14 @@ class Components:
         self._plugins = plugins
         self.openapi_version = openapi_version
         self._schemas = {}
+        self._examples = {}
         self._parameters = {}
         self._responses = {}
         self._security_schemes = {}
 
     def to_dict(self):
         subsections = {
+            "example": self._examples,
             "schema": self._schemas,
             "parameter": self._parameters,
             "response": self._responses,
@@ -46,6 +48,21 @@ class Components:
             for k, v in subsections.items()
             if v != {}
         }
+
+    def example(self, name, component, **kwargs):
+        """Add an example which can be referenced
+
+        :param str name: identifier by which example may be referenced.
+        :param dict component: example fields.
+
+        https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#exampleObject
+        """
+        if name in self._examples:
+            raise DuplicateComponentNameError(
+                'Another example with name "{}" is already registered.'.format(name)
+            )
+        self._examples[name] = component
+        return self
 
     def schema(self, name, component=None, **kwargs):
         """Add a new schema to the spec.
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -14,6 +14,7 @@ from apispec.exceptions import (
 
 from .utils import (
     get_schemas,
+    get_examples,
     get_paths,
     get_parameters,
     get_responses,
@@ -136,6 +137,13 @@ class TestComponents:
         "id": {"type": "integer", "format": "int64"},
         "name": {"type": "string", "example": "doggie"},
     }
+
+    # Referenced examples are only supported in OAS 3.x
+    @pytest.mark.parametrize("spec", ("3.0.0",), indirect=True)
+    def test_example(self, spec):
+        spec.components.example("PetExample", {"value": {"a": "b"}})
+        defs = get_examples(spec)
+        assert defs["PetExample"]["value"] == {"a": "b"}
 
     def test_schema(self, spec):
         spec.components.schema("Pet", {"properties": self.properties})
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -3,6 +3,10 @@
 from apispec.utils import build_reference
 
 
+def get_examples(spec):
+    return spec.to_dict()["components"]["examples"]
+
+
 def get_schemas(spec):
     if spec.openapi_version.major < 3:
         return spec.to_dict()["definitions"]
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -59,3 +59,4 @@ Contributors (chronological)
 - Dave `@zedrdave <https://github.com/zedrdave>`_
 - Emmanuel Valette `@karec <https://github.com/karec/>`_
 - Hugo van Kemenade `@hugovk <https://github.com/hugovk>`_
+- Ashutosh Chaudhary `@codeasashu <https://github.com/codeasashu>`_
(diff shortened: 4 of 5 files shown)


tests/utils.py
```diff
@@ -3,6 +3,12 @@
 from apispec.utils import build_reference
 
 
+def get_examples(spec):
+    if spec.openapi_version.major < 3:
```
lafrech: This condition should be removed.

tests/test_core.py
```diff
@@ -137,6 +138,14 @@ class TestComponents:
         "name": {"type": "string", "example": "doggie"},
     }
 
+    @pytest.mark.parametrize("spec", ("3.0.0",), indirect=True)
+    def test_example(self, spec):
+        # Referenced examples are only supported in OAS 3.x
```
lafrech: [Nitpick] I'd put this comment above `@pytest.mark.parametrize("spec", ("3.0.0",), indirect=True)`

src/apispec/core.py
```diff
@@ -47,6 +49,22 @@ def to_dict(self):
...
+        :param str name: identifier by which example may be referenced.
+        :param dict component: schema definition.
+
+        https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#exampleObject
+        """
+        if name in self._examples:
+            raise DuplicateComponentNameError(
+                'Another example with name "{}" is already registered.'.format(name)
+            )
+        # Skipping plugin helper
```
lafrech: I'd remove this comment.

lafrech (changes requested):
Thanks for this complete PR.

Just a few nitpicks and it's good to go.

I've been wondering if it would make sense to add `example_helper` to plugins. I guess a user could pass a schema and we could grab the `example` metadata from each field to build the complete example. Currently, when a user passes `example` as field metaclass, the value appears in the spec close to the field but not in a global response. Let's not complicate things. If someone ever needs to parse schemas to build examples, it can be done in another PR.

src/apispec/core.py
codeasashu: Yes, I had this added so I can add the `example_helper` support in this same PR, however, this may extend to several discussions, hence it can just be added who are using this package without plugin support.

Since this PR makes much more sense to document *response schemas* (normally you may only want to add example per schema field, which is good for plugin like marshmallow), marshmallow fields may be overkill features. `FlaskPlugin` makes much more sense going this route, which I will try to add in a separate PR soon (this weekend).

tests/utils.py
codeasashu: Fixed in 0776441f5060f10767cdd773fc5961c252a024c0

tests/test_core.py
codeasashu: Fixed in 0776441f5060f10767cdd773fc5961c252a024c0

src/apispec/core.py
codeasashu: Fixed in 0776441f5060f10767cdd773fc5961c252a024c0 .Marking this resolved

lafrech: I don't think I understand your answer but I'm fine with the rework.

Regarding your next PR, I don't see the link with `FlaskPlugin`. Don't bother explaining it to me if you find it easier to write the PR showing what you meant.

lafrech (approved):

src/apispec/core.py
```diff
@@ -47,6 +49,21 @@ def to_dict(self):
             if v != {}
         }
 
+    def example(self, name, component=None, **kwargs):
+        """Add a new example to the spec
+
+        :param str name: identifier by which example may be referenced.
+        :param dict component: schema definition.
```
lafrech: Uh-oh, looks like we let this slip through.

Should be "example fields".

lafrech (changes requested):

src/apispec/core.py
codeasashu: @lafrech Sure, adding PR with tests will make much more sense.

src/apispec/core.py
codeasashu: So, it should be "Add example fields to the spec", right?

codeasashu: or maybe: "Add an example which can be referenced", just like we have it in `parameter`'s docstring?

Also, I forgot to mention, `:param dict component: schema definition` is actually wrong, and it should be `:param dict component: example object` since examples are not actually definitions of anything but contains raw data?

lafrech: Yeah. Consistency first.

codeasashu: I've added the changes. Thanks for the reviews

tests/test_core.py
```diff
@@ -137,6 +138,14 @@ class TestComponents:
         "name": {"type": "string", "example": "doggie"},
     }
 
+    # Referenced examples are only supported in OAS 3.x
+    @pytest.mark.parametrize("spec", ("3.0.0",), indirect=True)
+    def test_example(self, spec):
+        spec.components.example("PetExample", {"value": {"a": "b"}})
+        defs = get_examples(spec)
+        assert "PetExample" in defs
```
lafrech: Nitpick ++

I'd remove this assert since it is covered by the line below.

I can do that later if you don't have time for that.

codeasashu: Done

src/apispec/core.py
```diff
@@ -47,6 +49,21 @@ def to_dict(self):
...
+        :param str name: identifier by which example may be referenced.
+        :param dict component: example fields.
+
+        https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#exampleObject
+        """
+        if name in self._examples:
+            raise DuplicateComponentNameError(
+                'Another example with name "{}" is already registered.'.format(name)
+            )
+        self._examples[name] = component or {}
```
lafrech: Since there is no helper, it makes no sense not to pass component.

This should be 

    self._examples[name] = component

and the signature should be

    def example(self, name, component, **kwargs):

like `security_scheme`.

codeasashu: Ah yes!! I forgot about it. Sorry for still being a noob

codeasashu: If you want, I can sqaush everything in a single commit again and force push, so that we can group these refactoring under a single commit, else feel free to merge this once the discussion is resolved

lafrech: No pb. Your first shot was pretty good already.

squash + push force would be great.

If you don't, I'll probably do it while merging.

Thanks again.

codeasashu: Thanks, I think it looks good

PR #517, opened 2019-11-28
Add match_info -> path to __location_map__

fedorfo:
### Before

If we add `@match_info_schema(SomeSchema)` to our method and do `setup_aiohttp_apispec`, we will have duplicated parameters in swagger.json. For example this method:

```
@docs(tags=["Tag"])
@match_info_schema(MatchInfoSchema)
@routes.get(r"/entity/{alias}")
async def get_entity(self) -> Tuple[Any, int]:
    ...
```

will add to swagger:

```
"parameters": [          
          {
            "in": "match_info",
            "name": "alias",
            "required": true,
            "type": "string"
          },
          {
            "in": "path",
            "name": "alias",
            "required": true,
            "type": "string"
          }
        ],
```

### After fix

All fields from match_info schema will be put in parameters with `in == "path"`

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -61,3 +61,4 @@ Contributors (chronological)
 - Hugo van Kemenade `@hugovk <https://github.com/hugovk>`_
 - Bastien Gerard `@bagerard <https://github.com/bagerard>`_
 - Ashutosh Chaudhary `@codeasashu <https://github.com/codeasashu>`_
+- Fedor Fominykh `@fedorfo <https://github.com/fedorfo>`_
--- a/src/apispec/ext/marshmallow/openapi.py
+++ b/src/apispec/ext/marshmallow/openapi.py
@@ -28,6 +28,7 @@ MARSHMALLOW_VERSION_INFO = tuple(
 
 
 __location_map__ = {
+    "match_info": "path",
     "query": "query",
     "querystring": "query",
     "json": "body",


lafrech:
From a quick search, match_info is used only in aiohttp. Is this correct?

Should this be done in a dedicated aiohttp support plugin?

(Forgive me if I'm missing the obvious. I'm not familiar with async frameworks.)

fedorfo:
>From a quick search, match_info is used only in aiohttp. Is this correct?

Seems that, the same property is used by [sanic](https://github.com/huge-success/sanic/blob/master/sanic/request.py#L494)

>Should this be done in a dedicated aiohttp support plugin?

At this moment we don't have any good option for changing this mapping outside (cause all we have is private dictionary). But yes, if we add any option for it, it will be okay for overriding in a dedicated plugin. And at this moment, I use the next code on my side:
```
    if with_swagger:
        from apispec.ext.marshmallow.openapi import __location_map__
        __location_map__["match_info"] = "path"
        setup_aiohttp_apispec(app, title="API", version="1.0", swagger_path="/api/doc")
```

But, anyway, aiohttp is one of popular http frameworks at this moment. And this `__location_map__` seems to be infected by different keywords from other frameworks. As I understand, `query` and `querystring` were added for different frameworks. So I don't see anything wrong in adding `match_info` key.

Pliner:
@lafrech I'm not sure but it seems there is no way for plugins based on `apispec` to let parameters be stored in `path` because there is no such value in `__location_map__`.

lafrech:
Sorry for the delay. Let's go ahead with this.

@fedorfo, do you mind adding yourself to AUTHORS file?

fedorfo:
> @fedorfo, do you mind adding yourself to AUTHORS file?

@lafrech Okay, why not? I did it.

lafrech:
All plugins I know use a dedicated path_helper to manage path parameters. They don't need `path` in the mapping.

https://github.com/marshmallow-code/apispec-webframeworks/blob/b5415316a672ed874d8e7a6c27c0bf6077e07f06/src/apispec_webframeworks/bottle.py#L54
https://github.com/marshmallow-code/apispec-webframeworks/blob/b5415316a672ed874d8e7a6c27c0bf6077e07f06/src/apispec_webframeworks/flask.py#L110https://github.com/marshmallow-code/apispec-webframeworks/blob/b5415316a672ed874d8e7a6c27c0bf6077e07f06/src/apispec_webframeworks/tornado.py#L97
https://github.com/marshmallow-code/flask-smorest/blob/b10c52319b731876b759505c9b05535bce7cf4e8/flask_smorest/spec/plugins.py#L80

Maybe this should be addressed somehow in aiohttp-apispec.

But apparently it is not that simple. Or maybe it could be done but adding this to the mapping here allows aiohttp-apispec to delegate more work to apispec, which is nice.

This, plus the fact that it is more or less standard (sanic uses it too).

Overall, I agree with @fedorfo in https://github.com/marshmallow-code/apispec/pull/517#issuecomment-559496613.

Adding this here comes with no cost, so if it makes user code much simpler, let's do it.

PR #521, opened 2019-12-11
Don't warn when using string "default" as status code in OASv2

lafrech:
This warning was introduced in #436.

According to https://swagger.io/docs/specification/2-0/describing-responses/ (section "Default Response"), `"default"` is a valid code.

Since the warning is not tested, I didn't modify any test.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,14 @@
 Changelog
 ---------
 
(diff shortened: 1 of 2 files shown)


lafrech:
Pinging @DStape who authored #436.

sloria (approved):

Issue #500, opened 2019-09-11
explode and style are hardcoded in OpenAPIConverter.property2parameter


lafrech:
I've been trying to document webargs's `DelimitedList` (child class of `List`) and I'm stuck because I'd need to override hardcoded stuff.

```py
    def field2parameter(self, field, name, default_in):
        location = field.metadata.get("location", None)
        prop = self.field2property(field)
        return self.property2parameter(
            prop,
            name=name,
            required=field.required,
            multiple=isinstance(field, marshmallow.fields.List),   # <--- List
            location=location,
            default_in=default_in,
        )

    def property2parameter(self, prop, name, required, multiple, location, default_in):
        openapi_default_in = __location_map__.get(default_in, default_in)
        openapi_location = __location_map__.get(location, openapi_default_in)
        ret = {"in": openapi_location, "name": name}

        if openapi_location == "body":
            ret["required"] = False
            ret["name"] = "body"
            ret["schema"] = {"type": "object", "properties": {name: prop}}
            if required:
                ret["schema"]["required"] = [name]
        else:
            ret["required"] = required
            if self.openapi_version.major < 3:
                if multiple:
                    ret["collectionFormat"] = "multi"
                ret.update(prop)
            else:
                if multiple:                    # <--- when List, set explode and style
                    ret["explode"] = True
                    ret["style"] = "form"
                if prop.get("description", None):
                    ret["description"] = prop.pop("description")
                ret["schema"] = prop
        return ret
```

To document `DelimitedList`, we'd need to set `explode` to False. That's assuming we use default `,` as delimiter. Using space or pipe would require to also modify `style`.

To make this more generic, we could remove `multiple` and create an extensible mechanism to allow custom post-processings after `property2parameter`. And add a post-processor for `List` to do what's currently done with `multiple`.

Kinda like what's been done with attribute_functions in `FieldConverter`.

(Maybe we could separate the property2bodyparameter case as right now it does not involve a specific `List` case.)

Bangertm:
Separating this out like with the attribute_functions makes sense to me. 

Note that based on the comment in #499 about not parsing location from metadata the path for `openapi_location == "body"` won't be used going forward because that logic is captured in `schema2parameters` here:  https://github.com/marshmallow-code/apispec/blob/15a06ebcc346d4ed1aa47e41b7a939393d7b9445/src/apispec/ext/marshmallow/openapi.py#L122-L135.

lafrech:
[...]

PR #526, opened 2020-01-01
Ignore "location" field metadata

lafrech:
Same rationale as https://github.com/marshmallow-code/webargs/pull/420. Ignore "location" in field metadata.

TODO:

- [x] Rename `default_in` as `location` or `in`. It is not a default anymore.
- [x] ~Remove dead code as per https://github.com/marshmallow-code/apispec/issues/500.~ Addressed in #581.

### DIFF
--- a/src/apispec/ext/marshmallow/openapi.py
+++ b/src/apispec/ext/marshmallow/openapi.py
@@ -110,27 +110,21 @@ class OpenAPIConverter(FieldConverterMixin):
         return self.get_ref_dict(schema_instance)
 
     def schema2parameters(
-        self,
-        schema,
-        *,
-        default_in="body",
-        name="body",
-        required=False,
-        description=None
+        self, schema, *, location="body", name="body", required=False, description=None
     ):
         """Return an array of OpenAPI parameters given a given marshmallow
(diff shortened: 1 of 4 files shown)


Bangertm:
Should also consider making the ‘in’ parameter required. Now that fields do not contain location apispec always has to know what to put in that parameter and defaulting to ‘body’ doesn’t make much sense.

lafrech:
Removal of potential dead code (see #500) can be done in another PR. But ideally before 4.0 if we decide to make some methods private.

Bangertm:
Just looking at this again. Once all the unneeded code is removed from `fields2parameters` the whole function can be simplified to a single list comprehension. 

```python
[
    self.field2parameter(
        field_obj,
        name=self._observed_name(field_obj, field_name),
        location=location,
    ) for field_name, field_obj in fields.item()
]
```
The conditional `if field_obj.dump_only: continue` isn't needed because that is already captured in the `get_fields` call above (not sure how there was so much duplication of logic).

Given this, instead of making the method private we could just remove it altogether and just do the list comprehension in `schema2parameters`

lafrech:
The refactor is done in #581.

I think this is ready to merge.

PR #528, opened 2020-01-03
Fix self-referencing schema in tests

lafrech:
`fields.Nested("self")` produces a warning.

    DeprecationWarning: Passing 'self' to `Nested` is deprecated. Use `Nested(lambda: MySchema(...))` instead.


sloria (approved):

lafrech:
Just noticed this PR broke compatibility with MA >=3.0.0,<3.3.0.

I don't know how to express this since

- marshmallow is optional dep
- marshmallow 2 is still supported

I guess we don't really care.

PR #554, opened 2020-04-21
Resolve components in operations

### LINKED ISSUE #550: Sample does not reference components
When I run the sample from https://github.com/marshmallow-code/apispec/blob/dev/docs/quickstart.rst

the expected output

```
# {'components': {'parameters': {},
#                 'responses': {},
#                 'schemas': {'Gist': {'properties': {'id': {'format': 'int64',
#                                                            'type': 'integer'},
#                                                     'name': {'type': 'string'}}}}},
#  'info': {'description': 'A minimal gist API',
#           'title': 'Gisty',
#           'version': '1.0.0'},
#  'openapi': '3.0.2',
#  'paths': OrderedDict([('/gist/{gist_id}',
#                         {'get': {'responses': {'200': {'content': {'application/json': {'schema': {'$ref': '#/definitions/Gist'}}}}}}})]),
#  'tags': []}
```

includes an additional `OrderedDict` (see `responses`) and does not set `$ref` to the schema:

```
{'components': {'schemas': {'Gist': {'properties': {'id': {'format': 'int64',
                                                           'type': 'integer'},
                                                    'name': {'type': 'string'}}}}},
 'info': {'description': 'A minimal gist API',
          'title': 'Gisty',
          'version': '1.0.0'},
 'openapi': '3.0.2',
 'paths': OrderedDict([('/gist/{gist_id}',
                        {'get': {'responses': OrderedDict([('200',
                                                            {'content': {'application/json': {'schema': 'Gist'}}})])}})])}
```
(version 3.3.0)

lafrech:
Fixes #550.

TODO:

- [x] Also manage requetBody in OAS3
- [x] Factorize what can be
- [x] Add tests

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -307,7 +307,7 @@ class TestPath:
                     }
                 ],
                 "responses": {
-                    "200": {"schema": "Pet", "description": "successful operation"},
+                    "200": {"description": "successful operation"},
                     "400": {"description": "Invalid ID supplied"},
                     "404": {"description": "Pet not found"},
                 },
@@ -617,6 +617,18 @@ class TestPath:
         message = "One or more HTTP methods are invalid"
         with pytest.raises(APISpecError, match=message):
             spec.path("/pet/{petId}", operations={"dummy": {}})
+
+    def test_path_resolve_response_schema(self, spec):
+        schema = {"schema": "PetSchema"}
+        if spec.openapi_version.major >= 3:
+            schema = {"content": {"application/json": schema}}
+        spec.path("/pet/{petId}", operations={"get": {"responses": {"200": schema}}})
+        resp = get_paths(spec)["/pet/{petId}"]["get"]["responses"]["200"]
+        if spec.openapi_version.major < 3:
+            schema = resp["schema"]
+        else:
+            schema = resp["content"]["application/json"]["schema"]
+        assert schema == build_ref(spec, "schema", "PetSchema")
 
 
 class TestPlugins:
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -630,6 +630,23 @@ class TestPath:
             schema = resp["content"]["application/json"]["schema"]
         assert schema == build_ref(spec, "schema", "PetSchema")
 
+    # requestBody only exists in OAS 3
+    @pytest.mark.parametrize("spec", ("3.0.0",), indirect=True)
+    def test_path_resolve_request_body(self, spec):
+        spec.path(
+            "/pet/{petId}",
+            operations={
+                "get": {
+                    "requestBody": {
+                        "content": {"application/json": {"schema": "PetSchema"}}
+                    }
+                }
+            },
+        )
+        assert get_paths(spec)["/pet/{petId}"]["get"]["requestBody"]["content"][
+            "application/json"
+        ]["schema"] == build_ref(spec, "schema", "PetSchema")
+
 
 class TestPlugins:
     @staticmethod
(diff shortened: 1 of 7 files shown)


lafrech:
I had to modify the `paths` used in the tests because it uses a schema passed as string and that would mess up with existing tests. Besides, it is not valid OAS 3.

lafrech:
I don't know if this should be considered a feature or a bugfix.

It fixes the code to match the doc, but it's basically adding a feature that wasn't there at all, although shown in one doc example.

I'd make this a minor release, not bugfix release.

tests/test_core.py
```diff
@@ -618,6 +618,35 @@ def test_path_check_invalid_http_method(self, spec):
         with pytest.raises(APISpecError, match=message):
             spec.path("/pet/{petId}", operations={"dummy": {}})
 
+    def test_path_resolve_response_schema(self, spec):
+        schema = {"schema": "PetSchema"}
```
Bangertm, line 622: This kind of input isn't comparable with using the Marshmallow plugin, right? From what I'm seeing the Marshmallow plugin will get called first look to resolve a  Marshmallow schema for the string "PetSchema" and then raise an exception. I might be wrong about that and I haven't tested it though.

lafrech: You're right, indeed. This is not compatible.

We could leave it as is and decide that the app should be either all-marshmallow or no-marshmallow and support this only if marshmallow plugin is not used.

Or we adapt the marshmallow plugin to pass strings, but this means it won't validate that the string corresponds to an existing schema. The schema instance resolver won't raise on unknown schema but return a string if the schema is unknown. We may hide typos or import order issues, but that's life.

It looks feasible. I didn't think it through but I'm not opposed to the idea.

Should this be considered a breaking change?

lafrech: We could even merge this as a non-breaking fix for users not using the marshmallow plugin and fix the marshmallow plugin compatibility later.

lafrech:
Considering #565, I'm tempted to modify marshmallow plugin as written above. We can make it a major release.

lafrech:
I just added a commit to implement the behaviour described above in MarmallowPlugin: if a string does not resolve to a schema from the registry, consider it a reference.

Breaking changes:

- We don't raise if a schema name is typoed, the error is silently ignored and a false ref is generated.
- Registry errors are not caught and turned into ValueError anymore (e.g. if Component.schema is given a string that is not found in the registry). I don't think it is that bad. The error message was not that helpful and the error can be tracked down with the raw RegistryError.

lafrech:
@Bangertm @sloria I just released a 4.0.0b1 with the latest changes in dev branch and rebased this branch on dev.

This is the last breaking change I'd like to merge into 4.0. It fixes #550 and #565.

I think #500 can be done in a 4.x release (although I'd double-check before releasing 4.0).

#557 can be solved later as well in a 4.x release.

Maybe #579 (OpenAPI 3.1  support) will need some refactoring but I haven't been looking into it yet and I'd rather not wait for it.

sloria:
Sounds good. I defer to you on releasing 4.0.0.

Good work!

Bangertm:
Makes sense to me too.

lafrech:
I shall merge/release this soon.

My only concern was [that comment](https://github.com/marshmallow-code/apispec/issues/565#issuecomment-630204544) from @Bangertm. But the comment doesn't really object to the feature, it merely says that it looks useless.

I don't think it has a significant downside, apart from hiding potential typos in schema names.

Bangertm:
I'm ultimately ok with this. Not convinced it's high value, but ok nonetheless.

lafrech:
> I think #500 can be done in a 4.x release (although I'd double-check before releasing 4.0).

Double-checked. Could be done with a potentially backwards incompatible refactor. Holding 4.0 a little more.

PR #559, opened 2020-04-28
Optionally disable 'nullable' values to follow Typescript convention.

h:
Previously, apispec.ext.marshmallow accepted both undefined and `null`
values for optional fields. This is different from the convention
followed by Typescript, which uses only undefined values in preference
over `null`.

See: https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined

This commit adds an option called `allow_null` that enables the new
functionality. By default, the previous convention is followed for
backward compatibility.


lafrech:
Doing a little cleanup in the repo. This didn't get any reaction. I don't really understand the point and I have the feeling this is not the right place to address this. apispec only does documention. Behaviour about null/empty/etc. happens in marshmallow (and is the subject of a few issues).

Issue #557, opened 2020-04-25
Components are not referenced when combined with oneOf, anyOf, allOf, not


lfiedler:
When combining multiple schemas with `oneOf`, `anyOf`, `allOf` or `not` that where previously added to components, they will not get referenced. I.e. running
```python
from pprint import pprint

from apispec import APISpec

spec = APISpec(
    title="Gisty",
    version="1.0.0",
    openapi_version="3.0.2",
    info=dict(description="A minimal gist API"),
)

spec.components.schema(
    "Gist",
    {
        "properties": {
            "id": {"type": "integer", "format": "int64"},
            "name": {"type": "string"},
        }
    },
)

spec.components.schema(
    "AnotherGist",
    {
        "properties": {
            "id": {"type": "integer", "format": "int64"},
            "name": {"type": "string"},
            "age": {"type": "integer", "format": "int32"},
        }
    },
)

spec.path(
    path="/gist/{gist_id}",
    operations=dict(
        get=dict(
            responses={
                "200": {
                    "content": {
                        "application/json": {
                            "schema": {
                                "oneOf": [
                                    "Gist", "AnotherGist"
                                ]
                            }
                        }
                    }
                }
            }
        )
    ),
)

if __name__ == "__main__": 
    pprint(spec.to_dict())
```
I get
```
{'components': {'schemas': {'AnotherGist': {'properties': {'age': {'format': 'int32',
                                                                   'type': 'integer'},
                                                           'id': {'format': 'int64',
                                                                  'type': 'integer'},
                                                           'name': {'type': 'string'}}},
                            'Gist': {'properties': {'id': {'format': 'int64',
                                                           'type': 'integer'},
                                                    'name': {'type': 'string'}}}}},
 'info': {'description': 'A minimal gist API',
          'title': 'Gisty',
          'version': '1.0.0'},
 'openapi': '3.0.2',
 'paths': OrderedDict([('/gist/{gist_id}',
                        {'get': {'responses': OrderedDict([('200',
                                                            {'content': {'application/json': {'schema': {'oneOf': ['Gist',
                                                                                                                   'AnotherGist']}}}})])}})])}

```
whereas I expected that components would be referenced automatically, i.e.
```
{'components': {'schemas': {'AnotherGist': {'properties': {'age': {'format': 'int32',
                                                                   'type': 'integer'},
                                                           'id': {'format': 'int64',
                                                                  'type': 'integer'},
                                                           'name': {'type': 'string'}}},
[...]

PR #391, opened 2019-02-17
Support oneOf, anyOf, and allOf schema combinators

stefanv:
Adds support for [OpenAPI v3 schema combinators](https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/)

I couldn't get the tests running locally yet, but I presume they should go in `tests/test_core.py`?

PR #560, opened 2020-05-03
Inheritance polymorphism

lfiedler:
This is a draft adressing issue #557. It is based on the pr #554. 

I copied the changes proposed there and adjusted them to the needs in #557. In particular, when cleaning `operations` it will now check for
keywords `allOf`. `anyOf`, `oneOf` and `not` at response or request contents. 
Depending on openapi version not all keywords are checked. The places at which are checked also depend on the version.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -63,3 +63,4 @@ Contributors (chronological)
 - Ashutosh Chaudhary `@codeasashu <https://github.com/codeasashu>`_
 - Fedor Fominykh `@fedorfo <https://github.com/fedorfo>`_
 - Colin Bounouar `@Colin-b <https://github.com/Colin-b>`_
+- Leander Fiedler `@lfiedler <https://github.com/lfiedler>`_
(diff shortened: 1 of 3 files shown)


stefanv:
@lfiedler Will this also address https://github.com/marshmallow-code/apispec/pull/391, or should I revisit that PR again?

lfiedler:
@stefanv I believe it does not address #391 but I do not know for sure. The changes I proposed are independent and outside of the marshmallow extension. They are supposed to work when directly using above mentioned keywords in `operations` in `APISpec.path` (see #557 for an example). 

To be more precise these changes only act on things that are passed to `APISpec.clean_parameters`.
If a schema gets resolved before that method is called and the schema gets replaced by a reference, then `clean_parameters` will not be able to treat this.
On the other hand the changes will not descend too deep into the schema found but instead will only check on the highest level.

Having said that, I don't know what the marhsmallow plugin does. Also I have not tested what happens when using it in conjunction with the marshmallow schema plugin and a marshmallow schema that uses polymorphism.

lafrech:
Thanks for the PR and sorry about the delay.

The recent rework (#655, #671) achieves this.

PR #564, opened 2020-05-18
adding support for marshmallow_enun.EnumFields

treidel:
Adding support to the Marshmallow plugin to populate marshmallow_enum.EnumFields into openapi 'enum' statements. 

Note:  this introduces a dependency in the Marshmallow plugin to the marshmallow_enum library.  I think this is worth having to have automatic encoding of enum values.   If the dependency is a no-no I'll think about wrapping this as a plugin.

### DIFF
--- a/src/apispec/ext/marshmallow/field_converter.py
+++ b/src/apispec/ext/marshmallow/field_converter.py
@@ -13,7 +13,7 @@ import warnings
 
 import marshmallow
 from marshmallow.orderedset import OrderedSet
-
+import marshmallow_enum
 
 RegexType = type(re.compile(""))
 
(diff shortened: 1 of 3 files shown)


treidel:
It looks like the CI is unhappy with my change.  tox ran ok in my dev environment (apart from compaints about python versions I don't have).  I'm guessing I missed adding the marshallow_enum dependency in the right place.   I'll take another stab at this tomorrow..

lafrech:
I believe our stance is

- For extra marshmallow libs from marshmallow-code organization, specific code may be integrated in apispec core, and the lib may be added as optional dependency (and test dependency).

- For third-party libs from outside marshmallow organization, we'd rather keep things in a separate lib, although we're happy is people share their snippets in this bugtracker, and provide feedback about hooks that should be implemented in apispec.

PR #581, opened 2020-07-10
Refactor openapi converter

lafrech:
Based on #526.

Implements refactor discussed in #500 and #526.

I did not address handling webargs field dicts (https://github.com/marshmallow-code/apispec/issues/500#issuecomment-570598012). It can be done later anytime (perhaps easier once MA2 is dropped).

Once #526 is merged, I can rebase this for a clearer diff.

### DIFF
--- a/src/apispec/ext/marshmallow/openapi.py
+++ b/src/apispec/ext/marshmallow/openapi.py
@@ -170,29 +170,11 @@ class OpenAPIConverter(FieldConverterMixin):
 
         https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#parameterObject
         """
-        return self._property2parameter(
-            self.field2property(field),
-            name=name,
-            required=field.required,
-            multiple=isinstance(field, marshmallow.fields.List),
-            location=location,
-        )
-
-    def _property2parameter(self, prop, *, name, required, multiple, location):
-        """Return the Parameter Object definition for a JSON Schema property.
-
-        https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#parameterObject
-
-        :param dict prop: JSON Schema property
-        :param str name: Field name
-        :param bool required: Parameter is required
-        :param bool multiple: Parameter is repeated
-        :param str location: Location to look for ``name``
-        :raise: TranslationError if arg object cannot be translated to a Parameter Object schema.
-        :rtype: dict, a Parameter Object
-        """
         openapi_location = __location_map__.get(location, location)
-        ret = {"in": openapi_location, "name": name, "required": required}
+        ret = {"in": openapi_location, "name": name, "required": field.required}
+
+        prop = self.field2property(field)
+        multiple = isinstance(field, marshmallow.fields.List)
 
         if self.openapi_version.major < 3:
             if multiple:
--- a/src/apispec/ext/marshmallow/openapi.py
+++ b/src/apispec/ext/marshmallow/openapi.py
@@ -161,7 +161,6 @@ class OpenAPIConverter(FieldConverterMixin):
                 location=location,
             )
             for field_name, field_obj in fields.items()
-            if not field_obj.dump_only
         ]
 
     def _field2parameter(self, field, *, name, location):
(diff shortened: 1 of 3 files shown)


src/apispec/ext/marshmallow/openapi.py
```diff
@@ -110,35 +110,29 @@ def resolve_nested_schema(self, schema):
...
+        In OpenAPI 3, only "query", "header", "path" or "cookie" are allowed for the location
+        of parameters. "requestBody" is used when fields are in the body.
+
         https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#parameterObject
         """
-        openapi_default_in = __location_map__.get(default_in, default_in)
-        if self.openapi_version.major < 3 and openapi_default_in == "body":
-            prop = self.resolve_nested_schema(schema)
-
+        openapi_location = __location_map__.get(location, location)
```
Bangertm: Ideally I would like to remove this mapping and rely on the calling code to know the appropriate OpenAPI locations.

src/apispec/ext/marshmallow/openapi.py
```diff
@@ -147,105 +141,38 @@ def schema2parameters(
...
-            multiple=isinstance(field, marshmallow.fields.List),
-            location=location,
-            default_in=default_in,
-        )
-
-    def property2parameter(
-        self, prop, *, name, required, multiple, location, default_in
-    ):
-        """Return the Parameter Object definition for a JSON Schema property.
+        openapi_location = __location_map__.get(location, location)
```
Bangertm: Even if we keep the mapping we shouldn't call it twice. The only reason that this had to be here was if the location was extracted from the field metadata where webargs used different strings.

src/apispec/ext/marshmallow/openapi.py
```diff
@@ -110,35 +110,29 @@ def resolve_nested_schema(self, schema):
...
 
     def schema2parameters(
-        self,
-        schema,
-        *,
-        default_in="body",
-        name="body",
-        required=False,
-        description=None
+        self, schema, *, location, name="body", required=False, description=None
```
Bangertm: I'm torn about what to do with the `name` parameter. It's only used when `location=="body"` and it's default value is tied to that case - I presume to guarantee a valid parameter object if the user doesn't enter anything. Seems like it's adding a little extra confusion to this method but I'm not sure anything better can be done.

src/apispec/ext/marshmallow/openapi.py
```diff
@@ -147,105 +141,38 @@ def schema2parameters(
...
-        parameters = []
-        body_param = None
-        for field_name, field_obj in fields.items():
-            if field_obj.dump_only:
-                continue
-            param = self.field2parameter(
+        return [
+            self._field2parameter(
                 field_obj,
                 name=self._observed_name(field_obj, field_name),
```
Bangertm, line 147: This is probably separate but we could make `self._observed_name` a function. It's not really doing much as a static method. That would clean this up a little in my opinion.

Bangertm (approved):
Overall this looks good. Really cleans up these methods.

src/apispec/ext/marshmallow/openapi.py
lafrech: Right. Fixed.

src/apispec/ext/marshmallow/openapi.py
lafrech: This method is removed in #583 anyway.

src/apispec/ext/marshmallow/openapi.py
lafrech: Yes, we should investigate this. I don't think I need this location, as my calling code could enter the correct value.

Indeed, this could, and probably should, be done in user code.

There are people relying on it, though. See #517.

lafrech:
Thanks @Bangertm for taking the time to review this.

I think I'm gonna merge this and open an issue about the location mapping.

src/apispec/ext/marshmallow/openapi.py
lafrech: I opened #587.

PR #582, opened 2020-07-10
Drop Python 3.5 and configure black

### LINKED ISSUE #556: Drop Python 3.5
Drop Python 3.5 in next major release.

Any objection?

lafrech:
Closes #556.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@ repos:
   rev: v2.4.1
   hooks:
   - id: pyupgrade
-    args: [--py3-plus]
+    args: [--py36-plus]
 - repo: https://github.com/python/black
   rev: 19.10b0
   hooks:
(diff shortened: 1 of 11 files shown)


sloria (approved):
Can also replace `--py3-plus` with  `--py36-plus` in pyupgrade:

https://github.com/marshmallow-code/apispec/blob/f075149bbd1238c1e9e17f5281d18bc0954739cd/.pre-commit-config.yaml#L6

We should also probably use `pyproject.toml` with a black config that has 

```
[tool.black]
line-length = 79
target-version = ['py36', 'py37', 'py38']
```

lafrech:
Changed to py36-plus.

About the pyproject file, whatever you say, I've never looked into black configuration so I'm trusting you. I think we use a longer line length, though. Did you mean 99?

sloria:
Woops, yeah I typod. We should probably match marshmallow's config

```toml
[tool.black]
line-length = 88
target-version = ['py35', 'py36', 'py37', 'py38']
```

sloria (approved):

PR #584, opened 2020-07-11
Allow yaml dump options

CtheSky:
### What I did
I made `spec.to_yaml` accept keyword arguments and pass them to the `yaml.dump`.

### Why
It's good to have more control on the dump behavior. I write many chinese characters in the doc, and `allow_unicode=True` option can make generated yaml more readable.

### DIFF
--- a/src/apispec/core.py
+++ b/src/apispec/core.py
@@ -221,11 +221,14 @@ class APISpec:
         ret = deepupdate(ret, self.options)
         return ret
 
-    def to_yaml(self):
-        """Render the spec to YAML. Requires PyYAML to be installed."""
-        from .yaml_utils import dict_to_yaml
+    def to_yaml(self, **kwargs):
+        """Render the spec to YAML. Requires PyYAML to be installed.
 
-        return dict_to_yaml(self.to_dict())
+        :param dict kwargs: parameters used by :meth:`yaml.dump`
+        """
+        from .yaml_utils import yaml, YAMLDumper
+
+        return yaml.dump(self.to_dict(), Dumper=YAMLDumper, **kwargs)
 
     def tag(self, tag):
         """ Store information about a tag.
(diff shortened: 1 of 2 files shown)


lafrech:
Why not add `**kwargs` to `dict_to_yaml` and leave it in yaml_utils.py?

CtheSky:
That's exactly what i did at the first time, but it requires to pass `**kwargs` twice (`to_dict` -> `dict_to_yaml` -> `yaml.dump`). Since it's only used here so I think it's unnecessary to have an extra delegation.

CtheSky:
Another way to implement it is to provide `YAMLDumper` to the user, and let user call `yaml.dump` just like what `json.dumps` does in readme. How do you think about it? @lafrech

lafrech:
This looks fine enough to me:

```py
    def to_yaml(self, **kwargs):
        """Render the spec to YAML. Requires PyYAML to be installed."""
        from .yaml_utils import dict_to_yaml
        return dict_to_yaml(self.to_dict(), **kwargs)
```
and

```py
def dict_to_yaml(dic, **kwargs):
    return yaml.dump(dic, Dumper=YAMLDumper, **kwargs)
```

We could expose the Dumper and all for more specific needs. Well, in fact we do.

`to_yaml` is just a simpler helper meaning to fit most people's needs but anyone can write his own equivalent custom function.

Let's not overthink this.

CtheSky:
I agree with you that `to_yaml` is just a simple helper. And the change isn't necessary since `Dumper` has been exposed. When I started this pr last week, I was looking for a way to allow unicode characters in generated yaml, after reading the code I found the `Dumper` in `yaml_utils`. I think maybe it's better to mention the `Dumper` is the example doc rather than making `to_yaml` accepts `**kwargs`.

Thanks for the review.

PR #585, opened 2020-07-13
feat: support fields metadata with non-string fields

sanzoghenzo:
This commit solves the issue that I wrongly filed to [flask-apispec repo](https://github.com/jmcarp/flask-apispec/issues/196).

Using desert to convert a dataclass/attrs class to a Schema, the metadata contains a non string key (_DesertSentinel) that causes the metadata2properties method to raise an AttributeError.

This quick fix uses str(key) in the two dict comprehensions to get the string representation of the key and return the dictionary.

### DIFF
--- a/src/apispec/ext/marshmallow/field_converter.py
+++ b/src/apispec/ext/marshmallow/field_converter.py
@@ -395,18 +395,16 @@ class FieldConverterMixin:
         """
         # Dasherize metadata that starts with x_
         metadata = {
-            key.replace("_", "-")
-            if isinstance(key, str) and key.startswith("x_")
-            else key: value
+            key.replace("_", "-") if key.startswith("x_") else key: value
             for key, value in field.metadata.items()
+            if isinstance(key, str)
         }
 
         # Avoid validation error with "Additional properties not allowed"
         ret = {
             key: value
             for key, value in metadata.items()
-            if key in _VALID_PROPERTIES
-            or (isinstance(key, str) and key.startswith(_VALID_PREFIX))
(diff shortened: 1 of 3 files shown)


lafrech:
Thank you for your contribution.

I think I'd check the key is a string rather than stringifying whatever passes by.

Note to self: BTW, I'm realizing it is strange to use `_VALID_PREFIX` after hardcoding `x_`. I shall have a look at this. But that's another issue.

```py
            key.replace("_", "-") if key.startswith("x_") else key: value
            [...]
            if key in _VALID_PROPERTIES or key.startswith(_VALID_PREFIX)
```

sanzoghenzo:
Ok, changed to use `isinstance(key, str)` and fix a linting error.

src/apispec/ext/marshmallow/field_converter.py
```diff
@@ -395,15 +395,18 @@ def metadata2properties(self, field, **kwargs):
         """
         # Dasherize metadata that starts with x_
         metadata = {
-            key.replace("_", "-") if key.startswith("x_") else key: value
+            key.replace("_", "-")
+            if isinstance(key, str) and key.startswith("x_")
+            else key: value
             for key, value in field.metadata.items()
         }
```
lafrech, line 402: I think you could write

```py
        metadata = {
            key.replace("_", "-") if key.startswith("x_") else key: value
            for key, value in field.metadata.items()
            if isinstance(key, str) 
        }
```

to drop non string keys, then remove the test in the second comprehension.

sanzoghenzo: You're absolutely right!
done

lafrech (approved):

Issue #586, opened 2020-07-21
Missing example on using a security_scheme


teancom:
Security schemes as a concept are mentioned [here](https://apispec.readthedocs.io/en/latest/special_topics.html#documenting-security-schemes) but there is no mention in the documentation or the unit tests showing how to actually _apply_ a security scheme to an API in general, or specific end-points (which is my use-case).

It's quite possible that the inability to figure out what's needed is due to my lack of experience with both OpenAPI as well as python programming in general, but contrary-wise I was able to get everything else working exactly how I wanted (for which I'm very grateful! This was me dipping my toe into RESTful programming for the first time, and apispec was *very* helpful in generating great-looking docs).

I can document what didn't work. I had this in my openapi.py:

```
api_key_scheme = {"type": "apiKey", "in": "header", "name": "X-API-KEY"}
spec.components.security_scheme("ApiKeyAuth", api_key_scheme)
```
And this in my code:
```
@apikey_required
def delete_event(id):
    """ DELETE EVENT route
    ---
    delete:
        summary: DELETE EVENT route
        description: Delete a single Event specified by ID
        security:
          - ApiKeyAuth: []
    <snip the rest>
    """
```
This got me *very* close, though. In order for tools like Stoplight.io or Redoc to recognize that the ApiKeyAuth was required for this route, I ended up manually editing the generated YAML file, which looked like this:

```
    delete:
      description: Delete a single Event specified by ID
      parameters:
      - in: path
        name: id
        required: true
        schema:
          format: int32
          type: integer
      responses:
        '200':
          description: JSON object representing an Event
          schema:
            $ref: '#/components/schemas/Event'
        '404':
          description: Event not found.
      security:
      - ApiKeyAuth: []
```

By moving the `- ApiKeyAuth: []` section over to the right two spaces.

I tried a bunch of variations in the docstring in my python code - and it's very, very possible that I'm close - but couldn't figure out the magic incantation to get it to generate correctly. A working example of applying a security_scheme to a route would have been very, very helpful.

If someone can put a bare-minimum working example into this issue, I'd be happy to add it into the docs in a PR.

Thanks again.

teancom:
Well I have no idea what I was doing wrong before, but using the same code it now works. Sigh. I'll still open a PR to add an example of using a security_scheme in a function, just so that nobody else goes down the rabbit-hole like I did.

PR #589, opened 2020-07-23
Add example of using a security_scheme to the documentation

teancom:
This is in reference to issue #586 - I have added an example of using a security scheme to protect a route, in a way that would have helped me personally. Hopefully it helps others as well!

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -63,3 +63,4 @@ Contributors (chronological)
 - Ashutosh Chaudhary `@codeasashu <https://github.com/codeasashu>`_
 - Fedor Fominykh `@fedorfo <https://github.com/fedorfo>`_
 - Colin Bounouar `@Colin-b <https://github.com/Colin-b>`_
+- David Bishop `@teancom <https://github.com/teancom>`_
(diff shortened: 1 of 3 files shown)


sloria:
Thanks! Looks good to me. Can you also copy this to `README.rst`? Also, please add yourself to `AUTHORS.rst`.

teancom:
Done! Thanks again for apispec, btw. It was lovely (and possibly helped get me a job! We'll see next week. 😄)

lafrech:
Thanks for taking the time to document this.

The example in the README lacks the declaration of the security scheme for the resource both in the code and in the json output. It is only present in the yaml output.

Also I'm not sure about documenting this in the quickstart (readme or index). This should be in a dedicated second step section. There are several auth schemes and there is no reason to document api key here and not the others.

teancom:
Oops! Sorry, yeah, updated the missing bits. I have no (real) opinion as to whether this should all be broken out into a separate doc or whatnot. I do think having an example of authorization/security in the first place people look would be a nice thing to have, and I'm using api key because that's what I happened to be using. But if you want to swap it to be JWT or oath2 or whatever, I'm fine with that!

sloria:
This is OK for now. Putting this in a dedicated section would be a good idea, but it's good to have it somewhere for the time being.

PR #595, opened 2020-08-25
Don't document int format as int32 and float format as float

lafrech:
These are dependent on the machine running the code. We could try to play smart and use `sys` to adapt to the machine, but the doc might be generated on another machine than the one serving the code, so I don't think it is worth it.

Changing to int64 / double as it is the most common case.

Sort of "breaking" change. 4.0 would be the right moment for that.


sloria:
What if we just remove `format` full stop for `integer` and `number` rather than assuming the format?

lafrech:
The table in https://swagger.io/specification/#data-types made me think the format was mandatory.

https://swagger.io/docs/specification/data-models/data-types/ says format is optional, so let's remove it.

Users can override the mapping if they want to be more specific.

sloria (approved):

PR #596, opened 2020-08-28
Ignore field metadata with non-string keys

lafrech:
Rebase / squash of #585.

(I couldn't find a way to push in PR branch.)

Closes #585.


lafrech:
I'm not sure marshmallow really means to authorize non-string keys in metadata, but it doesn't hurt to add this check.

PR #597, opened 2020-09-06
do not add non docummented path

CedricCabessa:
As a user I want to be able to have route without docstring and generate a
valid yaml file.

Path without any documentation are returned like this:
```
paths:
  /random_pet: {}
```

However, if the path contains a parameter we have:

```
paths:
  /get_by_id/{id}: {}
```
Which is not ok, according to swagger editor:

`Declared path parameter "id" needs to be defined as a path parameter at either
the path or operation level`

This patch ignore route without any operation defined, so route without documentation are not included.

### DIFF
--- a/src/apispec/core.py
+++ b/src/apispec/core.py
@@ -283,6 +283,9 @@ class APISpec:
 
         self.clean_operations(operations)
 
+        if not operations:
+            return self
+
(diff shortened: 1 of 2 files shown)


CedricCabessa:
any though on this ?
@lafrech may be?

PR #604, opened 2020-10-19
Test against Python 3.9

sloria:


lafrech:
Nice. Didn't notice 3.9 was out already. It has a few nice features I'll be happy to use... in a few years when it is in Debian stable and support for older versions is dropped in my apps and libs.

`"Programming Language :: Python :: 3.9",` is missing in setup.py.

PR #616, opened 2020-12-11
Fix date range

### LINKED ISSUE #614: Minimum/maximum date is not jsonschema compliant
I'm using flask-apispec (but I think that the issue is related to apispec) to generate a swagger 2.0 schema from  webargs/marshmallow models that include datetimes with min/max boundaries like this:

```
mydate = fields.DateTime(
    required=True,
    validate=validate.Range(
        max=datetime.now(pytz.utc).replace(hour=23, minute=59, second=59),
        min=datetime(1900, 1, 1, tzinfo=pytz.utc),
        max_inclusive=True
    )
)
```

The generated json shema is:

```
"mydate": {
    "format": "date-time",
    "maximum": "Tue, 08 Dec 2020 23:59:59 GMT",
    "minimum": "Mon, 01 Jan 1900 00:00:00 GMT",
    "type": "string"
},
```

But when I run schemathesis on the generated schema I obtain a validation error

```
E   jsonschema.exceptions.ValidationError: 'Tue, 08 Dec 2020 23:59:59 GMT' is not of type 'number'
E  
E   Failed validating 'type' in schema['properties']['definitions']['additionalProperties']['properties']['properties']['additionalProperties']['properties']['maximum']:
E       {'type': 'number'}
E  
E   On instance['definitions']['Input']['properties']['mydate']['maximum']:
E       'Tue, 08 Dec 2020 23:59:59 GMT'
```

Schemathesis is right, because minimum and maximum values are expected to be "JSON numbers" (https://tools.ietf.org/html/draft-fge-json-schema-validation-00#section-5.1.3) and not datetimes

On the other side having min and max date validation is a very appreciated feature, so the option to avoid to set these boundaries on the models is not really feasible for my use cases. Furthermore the use of schemathesis is very helpful for my test suite and I would like to continue to use it.

Is there something that I can modify / configure to obtain jsonschema compliant models from webargs/marshmallow schema with datetimes min/max values?

Or maybe is there something that apispec could fix to obtain the result? (for example convert min/max datetimes into timestamps? Having a number should be adherent to the schema specification)

If not possible, I could try to open an issue on schemathesis to ask for some compatibility fixes. They already include [compatibility fixes for fastapi](https://schemathesis.readthedocs.io/en/stable/compatibility.html#using-fastapi) (that by using pydantic produces Draft 7 compatible schemas, while schemathesis validates against Draft 4 and Wright Draft 00) so maybe they would be interested in adding some compatibility fixes also for apispec. But in this case I would like to ask for your help to support the request by providing your motivations for not meeting the specifications or other information that they could ask to me.

I really thank you for your help

Bangertm:
Fixes #614.

Uses an OpenAPI extension to indicate a range if a field is not a number field such as a date. Also refactors the logic for creating range and length attributes.

### DIFF
--- a/src/apispec/ext/marshmallow/field_converter.py
+++ b/src/apispec/ext/marshmallow/field_converter.py
@@ -290,17 +290,16 @@ class FieldConverterMixin:
         ]
 
         attributes = {}
-        for validator in validators:
-            if validator.min is not None:
-                if hasattr(attributes, "minimum"):
-                    attributes["minimum"] = max(attributes["minimum"], validator.min)
-                else:
-                    attributes["minimum"] = validator.min
-            if validator.max is not None:
-                if hasattr(attributes, "maximum"):
-                    attributes["maximum"] = min(attributes["maximum"], validator.max)
-                else:
-                    attributes["maximum"] = validator.max
+        min_list = [
+            validator.min for validator in validators if validator.min is not None
+        ]
+        max_list = [
+            validator.max for validator in validators if validator.max is not None
+        ]
+        if min_list:
+            attributes["minimum"] = max(min_list)
+        if max_list:
+            attributes["maximum"] = min(max_list)
         return attributes
 
     def field2length(self, field, **kwargs):
@@ -328,22 +327,22 @@ class FieldConverterMixin:
         min_attr = "minItems" if is_array else "minLength"
         max_attr = "maxItems" if is_array else "maxLength"
 
-        for validator in validators:
-            if validator.min is not None:
-                if hasattr(attributes, min_attr):
-                    attributes[min_attr] = max(attributes[min_attr], validator.min)
-                else:
-                    attributes[min_attr] = validator.min
-            if validator.max is not None:
-                if hasattr(attributes, max_attr):
-                    attributes[max_attr] = min(attributes[max_attr], validator.max)
-                else:
-                    attributes[max_attr] = validator.max
+        equal_list = [
+            validator.equal for validator in validators if validator.equal is not None
+        ]
+        if equal_list:
+            return {min_attr: equal_list[0], max_attr: equal_list[0]}
 
-        for validator in validators:
-            if validator.equal is not None:
-                attributes[min_attr] = validator.equal
-                attributes[max_attr] = validator.equal
+        min_list = [
+            validator.min for validator in validators if validator.min is not None
+        ]
+        max_list = [
+            validator.max for validator in validators if validator.max is not None
+        ]
+        if min_list:
+            attributes[min_attr] = max(min_list)
+        if max_list:
+            attributes[max_attr] = min(max_list)
         return attributes
 
     def field2pattern(self, field, **kwargs):
--- a/src/apispec/ext/marshmallow/field_converter.py
+++ b/src/apispec/ext/marshmallow/field_converter.py
@@ -290,6 +290,11 @@ class FieldConverterMixin:
         ]
 
         attributes = {}
+        min_attr, max_attr = (
+            ("minimum", "maximum")
+            if marshmallow.fields.Number in type(field).mro()
+            else ("x-minimum", "x-maximum")
+        )
         min_list = [
             validator.min for validator in validators if validator.min is not None
         ]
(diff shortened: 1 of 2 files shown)


src/apispec/ext/marshmallow/field_converter.py
```diff
@@ -328,23 +319,13 @@ def field2length(self, field, **kwargs):
         min_attr = "minItems" if is_array else "minLength"
         max_attr = "maxItems" if is_array else "maxLength"
 
-        for validator in validators:
-            if validator.min is not None:
-                if hasattr(attributes, min_attr):
```
lafrech, line 333: IIUC, the `hasattr` test was wrong and should have been written

```py
    if "min" in validator
```

This code was broken and the function would always return the last validator min since `hasattr` would always return `False`.

This PR fixes that bug.

src/apispec/ext/marshmallow/field_converter.py
```diff
@@ -328,23 +319,13 @@ def field2length(self, field, **kwargs):
...
-            if validator.equal is not None:
-                attributes[min_attr] = validator.equal
-                attributes[max_attr] = validator.equal
-        return attributes
+        equal_list = [
+            validator.equal for validator in validators if validator.equal is not None
+        ]
+        if equal_list:
+            return {min_attr: equal_list[0], max_attr: equal_list[0]}
+
```
lafrech, line 329: Both former and new code don't deal with the case of a user entering two incompatible equal constraints (or an equal constraint incompatible with a range). This is similar to #198 and I'm not sure we should do something about it.

The change introduced here is that if two different equal constraints are provided, we're keeping the last one while we used to keep the first. Just a note, I don't think it matters.

src/apispec/ext/marshmallow/field_converter.py
```diff
@@ -289,19 +289,12 @@ def field2range(self, field, **kwargs):
...
-                    attributes["minimum"] = validator.min
-            if validator.max is not None:
-                if hasattr(attributes, "maximum"):
-                    attributes["maximum"] = min(attributes["maximum"], validator.max)
-                else:
-                    attributes["maximum"] = validator.max
-        return attributes
+        min_attr, max_attr = (
+            ("minimum", "maximum")
+            if marshmallow.fields.Number in type(field).mro()
```
lafrech: This is not perfectly generic as it assumes only subclasses of `Number` field  may represent numbers.

lafrech (approved):
A generic solution would test the type from `field2type_and_format`. It would be a pity to reproduce that function's logic in `field2range` (and tomorrow potentially in another function). This is why I said that we might need to execute `field2type_and_format` first and pass the result as kwargs to other functions.

Yet, I consider the case of a number field that wouldn't inherit from `Number` a corner case we may miss for now, so we can merge this and refactor later. Besides, if we do refactor as I suggest, it will have to be in a major release as it breaks potential custom attribute functions not expecting new kwargs.

Meanwhile, the custom number field not inheriting from `Number` can be dealt with by overriding `field2range`, which is not that bad.

Bangertm:
@lafrech I didn't understand your comment about using the result of `field2type_and_format` before. Thanks for clarifying. Yes that is a better test. In fact `field2range` (and all the other attribute functions) already has access to the result of `field2type_and_format` via the `ret` keyword argument: 
https://github.com/marshmallow-code/apispec/blob/dd31b87a9f77f8665f37127789c09565334dd253/src/apispec/ext/marshmallow/field_converter.py#L160-L161

So I can just change the test to rely on the `type` attribute, which allows users to declare a custom field to be a number and makes this code generic. We could potentially do this without changing the function signature because `kwargs` is available, but I think it's better to make the use of `ret` explicit in the signature. Thoughts?

lafrech:
Oh, right. I read the code too fast and missed the fact that we already have `kwargs` with `ret` in them. Yes, let's use this.

We just need to be sure `field2type_and_format` runs first. Maybe this deserves a comment in `self.attribute_functions` declaration. We already relied on the order since we use `ret` in `nested2properties`.

lafrech:
I rebased (no conflict) and added a new commit to check on `ret["type"]`. It can be `"number"` but also `"integer"`. I changed a test field into a `Float` to test both.

I used `ret` in the signature. I wanted to keep kwargs in case another kwarg would be added some day but I removed it to be consistent with `nested2properties`. We may add **kargs to their signatures anytime and I think we can do it now but I didn't want to add it to this PR.

lafrech:
@Bangertm I'd request a review from you but apparently I can't since it is your PR...

Feel free to merge and even release if you're happy with my changes. Or roll them back if you're not.

lafrech:
Taking it upon myself to merge this. From the discussion we agree on the implementation.

PR #625, opened 2021-01-12
deepupdate change recursion to loop

lucasfcnunes:
Solves RecursionErrors exceptions/crashes caused by bloated memory use or high recursion depth when calling deepupdate().
*The same issues can occur on copy.deepcopy() calls.

### DIFF
--- a/src/apispec/utils.py
+++ b/src/apispec/utils.py
@@ -158,15 +158,19 @@ def dedent(content):
     return content.strip()
 
 
-# http://stackoverflow.com/a/8310229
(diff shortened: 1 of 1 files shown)


lafrech:
Do you have a real life use case where the update process significantly eats up the memory to a point where the app crashes?

Or is this more of a theoretical issue?

lucasfcnunes:
oh, i did.
at the time of this commit i was actually facing this issue

i was recursively building an api spec using a web scraping engine
i tried running it on a new async thread to tweak the python stack parameters, but i ended up having real seg faults/core dumps

PR #634, opened 2021-02-05
Make components dicts public attributes of Components class

lafrech:
This allows the caller to check a component is already registered. For instance to avoid registering it several times.

I could try/except but I think exceptions are more suited for... exceptions. If testing is the normal flow (because the code is expected to run several time and only register once) I think the try/except method does not express the intent.

Another reason is that even in apispec itself, `_schemas` is accessed from another class.

flask-smorest auto registers default responses when a response contains a name from `http.HTTPStatus`. I'd like to allow the user to register his own response and only register a default response if he didn't. There my code needs to register the response only if it is not in components.schemas and it is a pity to keep track of which is registered in another list. More on this in https://github.com/marshmallow-code/flask-smorest/pull/208. This use case is not that important. I just don't see why those should be private.

### DIFF
--- a/src/apispec/core.py
+++ b/src/apispec/core.py
@@ -29,19 +29,19 @@ class Components:
     def __init__(self, plugins, openapi_version):
         self._plugins = plugins
         self.openapi_version = openapi_version
-        self._schemas = {}
-        self._responses = {}
-        self._parameters = {}
-        self._examples = {}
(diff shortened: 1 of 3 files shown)


Bangertm (approved):
Looks good to me. No issue making these public that I can see.

PR #635, opened 2021-02-05
Populate additionalProperties from Meta.unknown

timsilvers:
Derive the value of additionalProperties from Meta.unknown, if defined.
additionalProperties will be true only if Meta.unknown == INCLUDE

Background here: https://marshmallow.readthedocs.io/en/stable/quickstart.html#handling-unknown-fields

If I am missing an existing way of doing this then I'll be happy to hear about it ...


sloria (approved):
Thanks!

PR #636, opened 2021-02-09
Fix kwargs in Components methods

lafrech:
Minor fixes:

- complete docstring
- remove unused kwargs (should this be considered a breaking change ?)


lafrech:
We should probably also harmonize name vs. component_id (let's go with component _id as name is ambiguous for components with a name field). And harmonize the methods content (use of f-strings).

Can be done in next major since signatures are modified.

lafrech:
Superseded by #696.

PR #637, opened 2021-02-09
Allow registration of headers

lafrech:
Contributes to #245.

This is an OAS 3 feature only.


Bangertm (approved):
Assume no need for resolving a Marshmallow Schema in the headers.

lafrech:
Yes, neither did I. Thanks for reviewing.

PR #646, opened 2021-03-15
Allow $ref values in parameter config

svdgraaf:
This is allowed according to the spec:
https://swagger.io/docs/specification/describing-parameters/#common-for-various-paths

This allows to add the common parameters outside of the regular app flow (eg import from elsewhere)

### DIFF
--- a/src/apispec/core.py
+++ b/src/apispec/core.py
@@ -348,7 +348,10 @@ class APISpec:
         """
         seen = set()
         for parameter in [p for p in parameters if isinstance(p, dict)]:
-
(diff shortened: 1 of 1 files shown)


svdgraaf:
I just found out if you just reference the parameter by name, it will auto generate the ref. I still think this PR makes sense though.

lafrech:
Yes, references are auto-resolved.

I'd rather not support two ways of doing the same thing. It is assumed everywhere that a string is a ref and should be automatically documented as such. I'd rather not add an exception here.

Maybe the feature deserves better documentation, though.

PR #648, opened 2021-03-21
Allow to_yaml to pass options to yaml.dump

pbzweihander:
When generating YAML with unicode, it is very annoying to have escaped unicode like `\uAC00`.

This PR allows `APISpec.to_yaml` method to pass additional options for `yaml.dump`.

With this PR, I can use options like `allow_unicode=True` to resolve the problem mention above.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -67,3 +67,4 @@ Contributors (chronological)
 - David Bishop `@teancom <https://github.com/teancom>`_
 - Andrea Ghensi `@sanzoghenzo <https://github.com/sanzoghenzo>`_
 - `@timsilvers <https://github.com/timsilvers>`_
+- Kangwook Lee `@pbzweihander <https://github.com/pbzweihander>`_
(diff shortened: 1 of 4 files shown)


lafrech:
Thanks. The implementation is fine.

I'm just wondering whether passing all args/kwargs to `dict_to_yaml` is the right API. If someday we decided that `to_dict` also takes parameters, we'd be screwed.

Should we namespace that in a `yaml_dump_kwargs` argument?

pbzweihander:
(Sorry for the multiple force-pushes.)

I updated to use `yaml_dump_kwargs` instead of `**kwargs`!

lafrech:
This was more of a question I was asking out loud at this point but thanks for the rework.

Using this argument name is not as nice as passing the arguments directly as kwargs, but it is safer on the long run and it doesn't matter that much. It's not as if this function was called everywhere in user code.

I approve this.

I wouldn't mind a second maintainer approval, just in case.

lafrech (approved):

PR #655, opened 2021-04-01
Refactor references resolution in Components

lafrech:
Attempt to put all component reference resolution work in `Components`.

This would be a first stop towards lazy component registration (https://github.com/marshmallow-code/flask-smorest/issues/236). But even if we don't do it, it is nice to separate concerns.

It might be a not-breaking change but since it is an important rework, I'm marking it as 5.0.

### DIFF
--- a/src/apispec/core.py
+++ b/src/apispec/core.py
@@ -190,6 +190,92 @@ class Components:
         self.security_schemes[component_id] = component
         return self
 
+    def get_ref(self, obj_type, obj):
+        """Return object or reference
+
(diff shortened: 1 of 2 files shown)


lafrech:
This PR modifies or removes public methods in `APISpec`, hence the 5.0 tag.

I don't have all 5.0 work ready and I have no ETA for it, but I'd like to build upon this. May I merge it in master and we'll pull a 4.x branch from 4.4.2 if we want to fix or just add something there?

PR #661, opened 2021-04-12
Update field2pattern doc-string

mjpieters:
The old doc-string is confusing:

- this applies to Regexp, not Range validators (the link is correct but the label is not)
- it only handles one Regexp per field, not a set.

This was probably a copy-paste error, this is the `field2pattern` doc-string with only the class reference updated and nothing else.


lafrech:
Good catch. Thanks.

Care to add yourself to AUTHORS.rst?

lafrech:
Perfect. Thanks.

Issue #659, opened 2021-04-07
Setting Meta.unknown=EXCLUDE should leave `additionalProperties` unset.


kupuguy:
Release 4.4.0 includes a change that sets `additionalProperties: false` when `Meta.unknown` is not `INCLUDE`.

The `Meta.unknown` field has 3 possible values: `RAISE`, `INCLUDE`, and `EXCLUDE`. The new change sets additionalProperties correctly for `RAISE` (additional properties are an error) and `INCLUDE` (additional properties are permitted) but `EXCLUDE` means additional fields may exist but will be removed without generating an error and this change means additional fields now generate an error.

Using INCLUDE and ignoring any extra fields isn't a workaround as (for example) when generating a type library for a typescript client the `additionalProperties: True` adds `[name: string]: any;` to the interface.

Ideally `unknown=RAISE` should set `additionalProperties: false`, `unknown=INCLUDE` should set `additionalProperties: true` and `unknown=EXCLUDE` should not set `additionalProperties` at all. This would mean connexion will still pass through any additional properties rather than erroring but I don't think there's a way to tell connexion to strip unknown fields.

jonadaly:
Note that 4.4.0 is a breaking change as a result of the above, if you use `Meta.unknown` of `EXCLUDE`.

The result of the update is that validation has become stricter (rejecting requests containing unknown fields), which could break people's APIs if they don't realise.

lafrech:
@kupuguy Thanks for the detailed report. Would you have the time to work on a PR?

@jonadaly I don't think this change would break an API. It only applies to the doc, not to the actual validation. I don't see it as a breaking change. It only makes the doc more precise. Well, it was supposed to, but it seems the `EXCLUDE` case was overlooked.

jonadaly:
It seems to applies to the validation as well - because `additionalProperties: false` means that any request with additional unknown fields gets rejected.

We are using apispec with connexion to validate requests, and connexion is raising HTTP 400 errors where previously it was allowing requests with unknown fields.

(It seems to me that setting unknown=EXCLUDE should still allow requests to be made with unknown extra fields - this was the case previously, but isn't any more with 4.4.0)

jonadaly:
(It seems to me that if you set unknown=EXCLUDE, requests with extra unknown fields should be allowed, which was the case previously but isn't with 4.4.0)

lafrech:
[...]

PR #663, opened 2021-04-16
Setting Meta.unknown=EXCLUDE should leave additionalProperties unset. #659

kupuguy:
Meta.unknown has three states:

* INCLUDE permits additional properties so in openapi `additionalProperties: true`
* RAISE rejects additional properties so in openapi `additionalProperties: false`
* EXCLUDE allows additional properties to be present but ignores them so leaving additionalProperties unset is the nearest equivalent.

This PR adds support for the EXCLUDE state.


Bangertm (approved):

lafrech:
Thanks. Please add yourself to AUTHORS.rst.

lafrech:
Sorry. I forgot about this.

PR #669, opened 2021-05-22
Make the build reproducible

lamby:
Whilst working on the [Reproducible Builds](https://reproducible-builds.org/) I noticed that `apispec` could not be built reproducibly. This is due to the copyright message (configured in `docs/conf.py`) including the current build date. (I originally filed this in Debian as bug [#988978](https://bugs.debian.org/988978).)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,6 +1,7 @@
 import datetime as dt
 import os
 import sys
+import time
 
 sys.path.insert(0, os.path.abspath(os.path.join("..", "src")))
 import apispec  # noqa: E402
@@ -23,10 +24,14 @@ intersphinx_mapping = {
 
 issues_github_path = "marshmallow-code/apispec"
 
+build_date = dt.datetime.utcfromtimestamp(
+    int(os.environ.get('SOURCE_DATE_EPOCH', time.time()))
+)
+
 source_suffix = ".rst"
 master_doc = "index"
 project = "apispec"
-copyright = f"Steven Loria {dt.datetime.utcnow():%Y}"
+copyright = f"Steven Loria {build_date:%Y}"
 
 version = release = apispec.__version__
 


lafrech:
This should be harmonized amongst our repos (marshmallow, webargs).

I'd go for the simplest. It could be removing the auto-date feature and bump date manually. After all, we already bump the license year manually.

@sloria ?

lafrech:
@lamby, I suppose marshmallow (and webargs and all...) are affected, right?

@sloria, mind if I remove that dynamic part to hardcode the date just like we do in the license file? Here and in other impacted projects.

lamby:
@lafrech Good qestion. Marshmallow may be fine [as it uses the `mtime` of the file](https://github.com/marshmallow-code/marshmallow/blob/dev/docs/conf.py#L36), but [webargs needs updating too](https://github.com/marshmallow-code/webargs/blob/dev/docs/conf.py#L39). (I would suggest using `SOURCE_DATE_EPOCH` for all of these, rather than mtime stuff.)

lafrech:
Thanks. I didn't know `SOURCE_DATE_EPOCH` was standardised. TBH, I didn't searched and somehow assumed it was a workaround you had to implement specifically for this, hence my proposal to simplify things and spare packagers the trouble.

https://reproducible-builds.org/docs/source-date-epoch/

> Before implementing this, you should scan through our checklist to see if you can avoid implementing it.

I'm still willing to simplify things, but I have no experience in Debian packaging, so if you think `SOURCE_DATE_EPOCH` is better than `mtime` and not worth hardcoding the year, then fine, let's go for it.

lamby:
> I'm still willing to simplify things, but I have no experience in Debian packaging, so if you think SOURCE_DATE_EPOCH is better than mtime and not worth hardcoding the year, then fine, let's go for it.

:+1:

lafrech:
I opened equivalent PRs on marshmallow (https://github.com/marshmallow-code/marshmallow/pull/1862) and webargs (https://github.com/marshmallow-code/webargs/pull/631).

After we merge this, I'll change the copyright line into

```py
    copyright = f"2014-{build_date:%Y}, Steven Loria and contributors"
```

to match webargs.

PR #675, opened 2021-06-03
Don't pass Field metadata as keyword arg

lafrech:
This is deprecated and generates warnings.


lafrech:
The only changes are in the tests, not in the source code (except for docstring examples).

This requires to bump marshmallow version to 3.10 for the tests but it is not a breaking change for the code itself. apispec still works with marshmallow >=3,< 3.10.

I don't think it is worth complicating things to try to also test on marshmallow 3.0. We can always drop marshmallow < 3.10 in next major and until then only test on 3.10+ and *know* it works on marshmallow 3.0.

PR #677, opened 2021-06-08
Add support for marshmallow.fields.Pluck()

### LINKED ISSUE #459: Marshmallow Pluck type not supported correctly?
Hi guys,

I was just trying to use Marshmallow 3 "nesting" via Pluck field (example taken from docs https://marshmallow.readthedocs.io/en/3.0/nesting.html) and swagger documentation rendered via `apispec` shows nested object in an array instead of flat array.

When using this example:

```
class UserSchema(Schema):
    name = fields.String()
    email = fields.Email()
    friends = fields.Pluck('self', 'name', many=True)
# ... create ``user`` ...
serialized_data = UserSchema().dump(user)
pprint(serialized_data)
# {
#     "name": "Steve",
#     "email": "steve@example.com",
#     "friends": ["Mike", "Joe"]
# }
deserialized_data = UserSchema().load(result)
pprint(deserialized_data)
# {
#     "name": "Steve",
#     "email": "steve@example.com",
#     "friends": [{"name": "Mike"}, {"name": "Joe"}]
# }
```

`apispec`-generated documentation says that ` "friends": ["Mike", "Joe"]` is in fact ` "friends": [{"name": "Mike"}, {"name": "Joe"}]`. So documentation is generated as for a normal nested object (without flattening it).

Is it just not supported yet or is there a bug?
Is there a plan to support this?

mjpieters:
Pluck is essentially a transplanted field from another schema.

This PR extends the nested2properties method to special-case Pluck fields, and
should fix #459.

### DIFF
--- a/src/apispec/ext/marshmallow/field_converter.py
+++ b/src/apispec/ext/marshmallow/field_converter.py
@@ -96,6 +96,7 @@ class FieldConverterMixin:
             self.field2pattern,
             self.metadata2properties,
             self.nested2properties,
+            self.pluck2properties,
             self.list2properties,
             self.dict2properties,
         ]
@@ -402,13 +403,32 @@ class FieldConverterMixin:
         :param Field field: A marshmallow field.
         :rtype: dict
         """
-        if isinstance(field, marshmallow.fields.Nested):
+        # Pluck is a subclass of Nested but is in essence a single field; it
+        # is treated separately by pluck2properties.
+        if isinstance(field, marshmallow.fields.Nested) and not isinstance(
+            field, marshmallow.fields.Pluck
+        ):
             schema_dict = self.resolve_nested_schema(field.schema)
             if ret and "$ref" in schema_dict:
                 ret.update({"allOf": [schema_dict]})
             else:
                 ret.update(schema_dict)
         return ret
+
+    def pluck2properties(self, field, **kwargs):
+        """Return a dictionary of properties from :class:`Pluck <marshmallow.fields.Pluck` fields.
+
+        Pluck effectively trans-includes a field from another schema into this,
+        possibly wrapped in an array (`many=True`).
+
+        :param Field field: A marshmallow field.
+        :rtype: dict
+        """
+        if isinstance(field, marshmallow.fields.Pluck):
+            plucked_field = field.schema.fields[field.field_name]
+            ret = self.field2property(plucked_field)
+            return {"type": "array", "items": ret} if field.many else ret
+        return {}
 
     def list2properties(self, field, **kwargs):
         """Return a dictionary of properties from :class:`List <marshmallow.fields.List>` fields.
--- a/tests/test_ext_marshmallow_field.py
+++ b/tests/test_ext_marshmallow_field.py
@@ -5,7 +5,7 @@ import pytest
 from marshmallow import fields, validate
 
 from .schemas import CategorySchema, CustomList, CustomStringField, CustomIntegerField
-from .utils import build_ref
+from .utils import build_ref, get_schemas
 
 
 def test_field2choices_preserving_order(openapi):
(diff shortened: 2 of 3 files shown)


tests/test_ext_marshmallow_field.py
```diff
@@ -318,6 +318,73 @@ def test_nested_field_with_property(spec_fixture):
     }
 
 
+def test_field2property_plock_spec_metadatas(spec_fixture):
```
lafrech: s/plokc/pluck

tests/test_ext_marshmallow_field.py
```diff
@@ -318,6 +318,73 @@ def test_nested_field_with_property(spec_fixture):
...
+        metadata={
+            "description": "Category breed",
+            "invalid_property": "not in the result",
+            "x_extension": "A great extension",
+        },
+    )
+    result = spec_fixture.openapi.field2property(breed)
+
+    spec_fixture.spec.components.schema("Category", schema=CategorySchema)
+    unplucked = spec_fixture.spec.components.schemas["Category"]["properties"]["breed"]
```
lafrech: I think I'd use `tests.utils.get_schema` to avoid relying on the internals of `Components`.

I admit we didn't do it 100% consistently in the tests.

I'm open to discussion and in any case this is not a blocker.

tests/test_ext_marshmallow_field.py
```diff
@@ -318,6 +318,73 @@ def test_nested_field_with_property(spec_fixture):
...
+    assert result == {
+        **unplucked,
+        "description": "Category breed",
+        "x-extension": "A great extension",
+    }
+
+
+def test_field2property_pluck_spec(spec_fixture):
```
lafrech: Not a blocker either. I'd put this test before the `_metadata` test.

lafrech (commented):
Thanks.

The fix is good. I might have put the Pluck case out of the Nested case as Pluck inherits from Nested mainly for practical reasons, but this won't change and if it does we'll adapt apispec. This implementation suits me.

Just a few nitpicks about the tests.

tests/test_ext_marshmallow_field.py
mjpieters: I'll just use `tests.utils.get_schema()` so as not to increase the tech debt.

mjpieters:
> I might have put the Pluck case out of the Nested case as Pluck inherits from Nested mainly for practical reasons

I actually agree, I've pulled out the change into a separate `pluck2properties` method. If only so others can override it separately in a subclass.

mjpieters:
Just in case this was missed: I pushed updates for the review comments, yesterday. Because I amended my git branch and force-pushed, that may not always be apparent.

src/apispec/ext/marshmallow/field_converter.py
```diff
@@ -402,14 +403,33 @@ def nested2properties(self, field, ret):
         :param Field field: A marshmallow field.
         :rtype: dict
         """
-        if isinstance(field, marshmallow.fields.Nested):
+        # Pluck is a subclass of Nested but is in essence a single field; it
+        # is treated separately by pluck2properties.
```
lafrech, line 407: This illustrates a limitation of our design.

Pluck inherits from Nested but has to be dealt with separately.

If someone creates a custom field inheriting a stock field, he may have to not only add a dedicated method but also potentially override the stock field (or its parent) method to add an exception for the custom field.

In practice, only container fields are concerned and we can live with this. I don't think custom fields for these are that common and I don't see a simple alternative to this design.

Still, it is a design smell IMHO. Or perhaps a design smell about Pluck field.

tests/test_ext_marshmallow_field.py
```diff
@@ -318,6 +318,73 @@ def test_nested_field_with_property(spec_fixture):
...
+
+    spec_fixture.spec.components.schema("Category", schema=CategorySchema)
+    unplucked = get_schemas(spec_fixture.spec)["Category"]["properties"]["breed"]
+
+    assert ret["type"] == "array"
+    assert ret["items"] == unplucked
+
+
+def test_pluck_field_with_property(spec_fixture):
+    breed_1 = fields.Pluck(CategorySchema, "breed")
```
lafrech: Perhaps rename to `test_field2property_pluck_field_with_property` for consistency but no big deal. If you don't have time, I can merge and rename later or just let it go. I'm not sure the other tests names are really consistent already.

lafrech (approved):
Thanks !

src/apispec/ext/marshmallow/field_converter.py
mjpieters: You have two options:
- use strict type checks (`type(field) is marshmallow.fields.Nested`). This would break any field inheriting from Nested without changing semantics the way `Pluck` does.
- factor out the common functionality that Pluck needs to reuse from Nested into a common utility class.

The latter would be my choice. It may, however, break 3rd-party code relying on `isinstance(pluck_field, Nested)` to be true.

tests/test_ext_marshmallow_field.py
mjpieters: I used the [`test_field2property_nest_*` tests](https://github.com/marshmallow-code/apispec/blob/f64ff702646375120320a09a85753a3be6e4995c/tests/test_ext_marshmallow_field.py#L249-L318) as a starting point, so the naming here is consistent with those tests. :-)

I'll clean this one up and split it into 2 tests (2 cases here are already covered by `test_field2property_pluck_spec` and `test_field2property_pluck_many`)

mjpieters: In fact, I've refactored all the new tests into a class for a more concise set of tests.

src/apispec/ext/marshmallow/field_converter.py
lafrech: Yes, exactly.

We shall investigate not having `Pluck` inherit `Nested`, but that would probably be a marshmallow 4 change. Meanwhile, we can live with this test in apispec.

lafrech:
Thanks!

PR #684, opened 2021-06-25
Add `writeOnly` and `nullable` to `_VALID_PROPERTIES` list

greyli:
The `nullable` and `writeOnly` are supported but not in the `_VALID_PROPERTIES` list. This PR also adds tests for `readOnly` and `writeOnly`.

P.S. I also found the metadata accepts "doc_default", is it a deprecated value (since we already have `default`)?


lafrech:
Thanks.

Regarding `doc_default`, see #196. But thanks for the heads up because since metadata is namespaced in marshmallow, the `doc_default` thing shouldn't be necessary. We shall remove it in apispec 5 (#687).

PR #685, opened 2021-06-25
Add pip-wheel-metadata/ to gitignore

greyli:
Not sure where it comes from, it appears after executed `pip install -e ".[dev]"`.


lafrech:
This shouldn't be necessary as per https://github.com/github/gitignore/pull/3364.

PR #706, opened 2021-09-03
BUG Fix ext/marshmallow/field_converter.py:219

sripathivenky:
Attribute error in Marshmallow field_converter.py

### DIFF
--- a/src/apispec/ext/marshmallow/field_converter.py
+++ b/src/apispec/ext/marshmallow/field_converter.py
@@ -216,7 +216,11 @@ class FieldConverterMixin:
         if "default" in field.metadata:
             ret["default"] = field.metadata["default"]
         else:
-            default = field.load_default
+            try:
+                default = field.load_default
+            except AttributeError:
+                # Marshmallow 3.11+ doesn't have load_default.
+                default = field.default
             if default is not marshmallow.missing and not callable(default):
                 default = field._serialize(default, None, None)
                 ret["default"] = default


lafrech:
Support for marshmallow < 3.13 was dropped in apispec 5.

https://github.com/marshmallow-code/apispec/blob/dev/CHANGELOG.rst#500b1-2021-07-22

Sinclert:
Hey there @lafrech  👋🏻 

I ended up reaching this issue after a bit of research in order to understand why [the documentation basic example](https://apispec.readthedocs.io/en/latest/#example-application) was not working with my current installation (`apispec` 5.1.0 and `marhsmallow` 3.12.2).

Reading the [referenced CHANGELOG entry](https://github.com/marshmallow-code/apispec/blob/dev/CHANGELOG.rst#500b1-2021-07-22), I see that support for  `marshmallow` <3.13 has been dropped. Please, understand that many projects may have a _loose_ dependency declaration, and the latest version of this package (currently `5.1.0`) will break their compatibility with existing marshmallow schemas.

Wold you consider accepting this PR?

lafrech:
Since apispec requires marshmallow >= 3.13, pip should install marshmallow 3.13+.

In this case, the schemas are not broken, they just generate a deprecation warning, which is fine.

Do you see a case where a user would use a recent apispec version and an older marshmallow version?

Sinclert:
Thanks for the quick reply.

I see that [this snipped in the setup.py](https://github.com/marshmallow-code/apispec/blob/9c6f120a9aa11c13ed45ed3de6a8e4b11ed24981/setup.py#L4-L22) specifies `marshmallow>=3.13.0`, when apispec is installed with some of its many _extra_ variations:

Sinclert:
Thanks for the quick reply.

I see that [this snipped in the setup.py](https://github.com/marshmallow-code/apispec/blob/9c6f120a9aa11c13ed45ed3de6a8e4b11ed24981/setup.py#L4-L22) specifies `marshmallow>=3.13.0`, when apispec is installed with some of its many _extra_ variations:

- apispec[marshmallow].
- apispec[docs].
- apispec[tests].
- apispec[dev].

In my particular case (and probably many others) I was installing the package just by doing `pip install apispec==5.1.0`, and did not know about the _extra_ options. If the intent is for the user to install `apispec[marshmallow]` **when running the basic example**, I suggest you specify so in the [README.md](https://github.com/marshmallow-code/apispec/blob/dev/README.rst) and in [this page of the documentation](https://apispec.readthedocs.io/en/latest/using_plugins.html#example-flask-and-marshmallow-plugins).

sripathivenky:
@lafrech  Totaly understands your reason for closing it. As @Sinclert pointed simple pip install apispec might defiantly cause this confusion. In my current implementation, I am using the below libraries. But those existed in my project for a couple of months before I started apispec to cleverly generate the open-api documents. Given the footprint of code change and as it adds a bit of graceful handle of the exception can we convince you to consider the PR please.  
marshmallow==3.11.1
marshmallow-jsonschema==0.11.1

lafrech:
@Sinclert you're absolutely right. I forgot marshmallow was not a hard dep.

I'm pretty sure 98% of users use marshmallow so we could make it a hard dep for simplicity but that seems wrong. And making it a separate package would be too much trouble.

I think the current situation is a reasonably optimal technical choice.

But I totally agree it deserves to be clearly documented. I opened #707. PR welcome or I'll try to do it soonish when I get the time.

Thanks for the heads up.

PR #727, opened 2021-10-28
Drop Python 3.6

lafrech:
- Drop Python 3.6.
- Don't use OrdereDict anymore since dict is ordered.

Python 3.6 EOLs on December 23rd 2021: https://endoflife.date/python


sloria (approved):

lafrech:
I'll probably wait until Xmas to merge this so that 3.6 is EOL already.

sloria:
yep, sounds good

PR #747, opened 2022-02-04
Add type hints

kasium:
Closes #745

PR #751, opened 2022-03-05
Move from Azure Pipelines to GitHub Actions

lafrech:
This is my flask-smorest config, modified to replicate current Azure config with daily builds against marshmallow-dev and remove test coverage.


lafrech:
I guess "marshmallow-code.apispec Expected — Waiting for status to be reported" is a leftover from he Azure integration and should disappear when we remove it from Azure config itself.

.github/workflows/build-release.yml
```diff
@@ -0,0 +1,46 @@
+name: build
+on:
+  pull_request:
+  # Run builds nightly to catch incompatibilities with new marshmallow releases
+  schedule:
+  - cron: "0 0 * * *"
+jobs:
+  lint:
```
sloria, line 8: should we use pre-commit.ci like in webargs?

.github/workflows/build-release.yml
```diff
@@ -0,0 +1,46 @@
...
+    needs: [lint, tests]
+    name: PyPI release
+    if: startsWith(github.ref, 'refs/tags')
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - run: python -m pip install --upgrade pip wheel
+      - run: pip install twine
+      - run: python setup.py sdist bdist_wheel
```
sloria: the pypa docs recommend using `build` rather than setup.py: https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

sloria (approved):
just a couple non-blocking nits. approving to unblock.

.github/workflows/build-release.yml
lafrech: We could. TBH, I followed the webargs thread from afar. This PR is based on the setup I use in for flask-smorest so it was easy for me to adapt.

We can use pre-commit.ci in a later phase. I'd like to merge this sooner to fix CI tests asap.

.github/workflows/build-release.yml
lafrech: Thanks for the link. I shall look into this. Perhaps it didn't exist back when I set this up for flask-smorest. Or I copied code from an outdated blog post.

.github/workflows/build-release.yml
sloria: sounds good. my comments are not blocking, so go ahead and merge when you feel it's right

.github/workflows/build-release.yml
lafrech: I just changed this, copying this part from webargs. We now use build as instructed in the docs. We're still using twine for consistency with webargs, instead of the GH action proposed in the docs.

lafrech:
@sloria I think this is good to go.

There's still this "marshmallow-code.apispec Expected — Waiting for status to be reported " line. I suppose something must be done in Azure pipeline config but I don't have the credentials. Do you think you could have a look at it?

Thanks.

I'd like to merge this so that CI issues are a thing of the past and we can go on with #747.

sloria (approved):
Great! I disabled the AP check requirement in the repo settings so this can pass. I'll enable the GHA check requirement after merging

lafrech:
I did search for this in the repo settings but couldn't find anything.

sloria:
Hm, I'm not sure why `dev` didn't build when I merged. investigating

sloria:
oh is it because only pull requests are built? https://github.com/marshmallow-code/apispec/blob/7359c1f975b7d14e9e4f94fbd2e6148e6137971f/.github/workflows/build-release.yml#L3

sloria:
bad timing 😬 https://www.githubstatus.com/incidents/sksd097hm0y5 . i think i have the fix locally but can't push

lafrech:
Yep. You beat me to it. Was writing this:

------------------

I'm afraid so. I removed the "push" event due to a misunderstanding.

I just fixed it by putting that "push" back

```
name: build
on:
  push:
  pull_request:
  # Run builds nightly to catch incompatibilities with new marshmallow releases
  schedule:
  - cron: "0 0 * * *"
```

but for some reason, I can't push to this repo anymore (while I could push force in this PR a few minutes ago):

```
remote: Resolving deltas: 100% (3/3), completed with 3 local objects.
remote: Internal Server Error
To ssh://github.com/marshmallow-code/apispec
 ! [remote rejected] fix_ci_gha_push -> fix_ci_gha_push (Internal Server Error)
error: impossible de pousser des références vers 'ssh://github.com/marshmallow-code/apispec'
```

Looks like a transient error: https://www.githubstatus.com/.

PR #765, opened 2022-04-29
Fix typing for APISpec.path path parameter and Plugin methods

### LINKED ISSUE #764: release 5.2.0: The annotations for path in `path()` have changed
https://github.com/marshmallow-code/apispec/blob/eda2887adb8baf414b280ce5bd5528e9614eaa66/src/apispec/core.py#L470-L482

I'm getting annotation errors with this release.

```
error: Argument "path" to "path" of "APISpec" has incompatible type "str"; expected "Optional[Dict[Any, Any]]"
```

lafrech:
Fixes #764.

Also makes mypy check all files, including tests.

I'm a bit surprised the `path` type issue doesn't get caught when checking files in tests/.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,5 +23,4 @@ repos:
   rev: v0.950
   hooks:
   - id: mypy
-    files: ^src/apispec/
     additional_dependencies: ["marshmallow>=3,<4", "types-PyYAML"]
--- a/src/apispec/core.py
+++ b/src/apispec/core.py
@@ -467,7 +467,7 @@ class APISpec:
 
     def path(
         self,
-        path: dict | None = None,
+        path: str | None = None,
         *,
         operations: dict[str, typing.Any] | None = None,
         summary: str | None = None,
--- a/src/apispec/core.py
+++ b/src/apispec/core.py
@@ -1,6 +1,8 @@
 """Core apispec classes and functions."""
 
 from __future__ import annotations
+from collections.abc import Sequence
+
 
 import typing
 from copy import deepcopy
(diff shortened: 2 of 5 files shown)


lafrech:
Ping @kasium who authored #747.

greyli:
Not sure if it's a typing issue in apispec.

Although `MarshmallowPlugin` defines that [its attributes can be `None`](https://github.com/marshmallow-code/apispec/blob/dev/src/apispec/ext/marshmallow/__init__.py#L124
), but after [the `init_spec` was called](https://github.com/marshmallow-code/apispec/blob/eedb86de67e444abb0e783e427d9c42e641e87cf/src/apispec/core.py#L437), its attributes should have the correct value instead of `None`.

It seems MyPy can't understand this, so when I use the `MarshmallowPlugin`:

```py
ma_plugin: MarshmallowPlugin = MarshmallowPlugin(
            schema_name_resolver=self.schema_name_resolver
        )

ma_plugin.converter.field_mapping[fields.URLFor] = ('string', 'url')
```
Then every call to its attributes (e.g. `ma_plugin.converter`) will produce a typing error like this:
```
error: Item "None" of "Optional[OpenAPIConverter]" has no attribute "field_mapping"
```

---

Oops, didn't realize the PR was already merged.

kasium:
@greyli I have a similar issue with marshmallow. As long as you don't have a mypy plugin or some fancy code, mypy can't pick that up, sadly

lafrech:
Yes, I'm afraid this is not specific to apispec. You might have to cast on every call.

If there's anything I can modify here to help with this, I'll gladly do it.

PR #766, opened 2022-05-01
Fix the type of the `path` arg of `APISpec.path()`

greyli:
dict -> str


lafrech:
Thanks. This is already included in #765 (review welcome !).

PR #769, opened 2022-05-11
Set default yaml.dump sort_keys to False to respect dictionary order

### LINKED ISSUE #768: Component schema property ordering regression in ApiSpec.to_yaml() method
In `apispec-5.1.1` schema property ordering in the yaml output matched the order of fields in the marshmallow schema, so long as you added a Meta() class to the schema, with an ordered=True attribute. As of `apispec-5.2.0` the marshmallow schema field ordering is no longer respected - instead the schema property ordering is always alphabetical.

My workaround is to re-implement `YamlDumper` and `_represent_dict()` functionality that was in `yaml_utils` up to `apispec-5.1.1` with an additional `add_representer()` call to ensure **OrderedDict** and **dict** instances are both ordered correctly and use this instead of the `to_yaml()` method:

```
def ordered_yaml_dump(data, stream=None, Dumper=SafeDumper, **kwds):
    class OrderedDumper(Dumper):
        pass

    def _represent_dict(dumper, data):
        return dumper.represent_mapping(yaml.resolver.BaseResolver.DEFAULT_MAPPING_TAG, data.items())

    OrderedDumper.add_representer(OrderedDict, _dict_representer)
    OrderedDumper.add_representer(dict, _dict_representer)
    return yaml.dump(data, stream, OrderedDumper, **kwds)
```
Related commit: https://github.com/marshmallow-code/apispec/commit/09dd98f26f75afefe17f22539dbedc7d64e5cb2b

vorticity:
Implemented fix proposed by @lafrech https://github.com/marshmallow-code/apispec/issues/768#issuecomment-1123292470

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -74,4 +74,4 @@ Contributors (chronological)
 - François Magimel `<https://github.com/Linkid>`_
 - Stefan van der Walt `<https://github.com/stefanv>`_
 - `<https://github.com/kasium>`_
-- Edwin Erdmanis `@vorticity <https://github.com/vorticity>`
+- Edwin Erdmanis `@vorticity <https://github.com/vorticity>`_
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -74,3 +74,4 @@ Contributors (chronological)
 - François Magimel `<https://github.com/Linkid>`_
 - Stefan van der Walt `<https://github.com/stefanv>`_
 - `<https://github.com/kasium>`_
+- Edwin Erdmanis `@vorticity <https://github.com/vorticity>`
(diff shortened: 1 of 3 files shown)


AUTHORS.rst
```diff
@@ -74,3 +74,4 @@ Contributors (chronological)
 - François Magimel `<https://github.com/Linkid>`_
 - Stefan van der Walt `<https://github.com/stefanv>`_
 - `<https://github.com/kasium>`_
+- Edwin Erdmanis `@vorticity <https://github.com/vorticity>`
```
lafrech: Missing trailing underscore.

lafrech (commented):
Thanks.

Do you think you could add a unit test for this?

AUTHORS.rst
vorticity: Good catch, thanks

lafrech (approved):
Great, thanks!

PR #778, opened 2022-07-13
Add field parameter functions

### LINKED ISSUE #500: explode and style are hardcoded in OpenAPIConverter.property2parameter

lafrech:
Add a layer above `field2property`to allow for more customization of the result.

Generalizes the `multiple`mechanism to deal with not only `List`.

There might be room for factorization with attribute functions by reworking `FieldConverterMixin` but `field2property` is also called in `fields2jsonschema` and I don't want to break something there.

Closes #500.

Still needs tests, doc, and I'd like to add an example of documenting webargs'`DelimitedList`.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -18,8 +18,9 @@ primary_domain = "py"
 default_role = "py:obj"
 
 intersphinx_mapping = {
-    "python": ("http://python.readthedocs.io/en/latest/", None),
-    "marshmallow": ("http://marshmallow.readthedocs.io/en/latest/", None),
+    "python": ("https://python.readthedocs.io/en/latest/", None),
+    "marshmallow": ("https://marshmallow.readthedocs.io/en/latest/", None),
+    "webargs": ("https://webargs.readthedocs.io/en/latest/", None),
 }
 
 issues_github_path = "marshmallow-code/apispec"
--- a/docs/using_plugins.rst
+++ b/docs/using_plugins.rst
@@ -295,6 +295,32 @@ method. Continuing from the example above:
 The function passed to `add_attribute_function` will be bound to the converter.
 It must accept the converter instance as first positional argument.
 
+In some rare cases, typically with container fields such as fields derived from
+:class:`List <marshmallow.fields.List>`, documenting the parameters using this
+field require some more customization.
+This can be achieved using the `add_parameter_attribute_function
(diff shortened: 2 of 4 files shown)


lafrech:
@Bangertm @sloria, I intend to merge this soon. The other v6 PRs should be easy so we're close to v6.

No pressure for a review. If you want to check this, I can wait. Otherwise, no problem.

sloria (commented):
Looks good!

Bangertm (approved):
This looks good to me.

PR #793, opened 2022-09-15
pre-commit autoupdate + update setup.cfg

lafrech:
pre-commit autoupdate + a bit of cleanup in setup.cfg

More context in https://github.com/marshmallow-code/marshmallow/pull/2051.

Removes a few ignores that don't trigger anyway.


lafrech:
CI failure about unused type:ignore might be due to the typing fix in latest marshmallow. Removing type:ignore.

PR #805, opened 2022-10-19
Add support for marshmallow.fields.Number.as_string

rockTA:
Some applications serialize decimal numbers as strings, as binary representation of floats can't be precise. However, when the OpenAPI documentation is created, this is currently not being taken into account if the `Decimal` type is used with the `as_string` option. With this change the documentation accurately represents what is being returned.


rockTA:
Push :)

lafrech (changes requested):
Thanks for contributing and sorry about the delay.

I think it would be nicer if we added a `number2properties` method overriding the output of `field2type_and_format`.

matejsp:
This PR is also missing also adding format as decimal when using number as_string.

```
      amount:
          description: Some amount
          example: '90.00'
          readOnly: true
          type: string
          format: decimal
```

PR #808, opened 2022-11-06
Remove unneeded `mock` dependency

mgorny:
Remove the `mock` dependency.  It seems that its last use was removed in 5293ab11f88f3e3662d18084327eeb697a173184, and all tests pass without it being installed.


mgorny:
(I haven't added myself to `AUTHORS` since this isn't really a code contribution ;-))

lafrech:
Good catch. Thanks.

mgorny:
Thanks!

PR #815, opened 2022-12-05
Datetime support for multiple format

TheBigRoomXXL:
Adding a function `datetime2properties` to support the differents formats of `marshamllow.field.DateTime`. 
An exemple for custom DateTime format has been added to the doc in `Using Plugins / Marshmallow Plugin`

The support for `timestamp` and `timestamp_ms` mean that DateTime type can be either "string" or "integer". Format can also have multiple values.  Because of that 2 test where modified to account for that variability: 
- `test_field2property_type`
- `test_field2property_formats`

Then 6 tests were added to test the different output of `datetime2properties`.

### DIFF
--- a/src/apispec/ext/marshmallow/field_converter.py
+++ b/src/apispec/ext/marshmallow/field_converter.py
@@ -549,7 +549,7 @@ class FieldConverterMixin:
                     "example": "1676451245.596",
                     "min": "0",
                 }
-            elif field.format ==  "timestamp_ms":
+            elif field.format == "timestamp_ms":
                 ret = {
                     "type": "number",
                     "format": "float",
--- a/src/apispec/ext/marshmallow/field_converter.py
+++ b/src/apispec/ext/marshmallow/field_converter.py
@@ -110,6 +110,7 @@ class FieldConverterMixin:
             self.list2properties,
             self.dict2properties,
             self.timedelta2properties,
+            self.datetime2properties,
         ]
 
     def map_to_openapi_type(self, field_cls, *args):
@@ -516,6 +517,53 @@ class FieldConverterMixin:
             else:
                 choices = (m.value for m in field.enum)
             ret["enum"] = [field.field._serialize(v, None, None) for v in choices]
+        return ret
+
+    def datetime2properties(self, field, **kwargs: typing.Any) -> dict:
+        """Return a dictionary of properties from :class:`DateTime <marshmallow.fields.DateTime` fields.
+
+        :param Field field: A marshmallow field.
+        :rtype: dict
+        """
+        ret = {}
+        if isinstance(field, marshmallow.fields.DateTime) and not isinstance(
+            field, marshmallow.fields.Date
+        ):
+            if field.format == "iso" or field.format is None:
+                # Will return { "type": "string", "format": "date-time" }
+                # as specified inside DEFAULT_FIELD_MAPPING
+                pass
+            elif field.format == "rfc":
+                ret = {
+                    "type": "string",
+                    "format": None,
+                    "example": "Wed, 02 Oct 2002 13:00:00 GMT",
+                    "pattern": r"((Mon|Tue|Wed|Thu|Fri|Sat|Sun), ){0,1}\d{2} "
+                    + r"(Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec) \d{4} \d{2}:\d{2}:\d{2} "
+                    + r"(UT|GMT|EST|EDT|CST|CDT|MST|MDT|PST|PDT|(Z|A|M|N)|(\+|-)\d{4})",
+                }
+            elif field.format == "timestamp":
+                ret = {
+                    "type": "number",
+                    "format": "float",
+                    "example": "1676451245.596",
+                    "min": "0",
+                }
+            elif field.format ==  "timestamp_ms":
+                ret = {
+                    "type": "number",
+                    "format": "float",
+                    "example": "1676451277514.654",
+                    "min": "0",
+                }
+            else:
+                ret = {
+                    "type": "string",
+                    "format": None,
+                    "pattern": field.metadata["pattern"]
+                    if field.metadata.get("pattern")
+                    else None,
+                }
         return ret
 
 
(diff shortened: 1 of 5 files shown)


TheBigRoomXXL:
I am sorry but do you need something more before merging? I don't really know why this pull request stay open if there is no problem with it.

lafrech:
Simply lack of benevolent maintainer time. Please keep this open. I just can't promise when it will be reviewed. To anyone, feel free to chime in and help with the review.

PRs that are obviously wrong are closed, PRs that are trivially good are merged. Other PRs take time to review. Sorry about that. Thanks for contributing.

Bangertm:
The one thing I'm thinking about with this code is whether it should be more closely integrated with `field2type_and_format`. The first condition is leaving the results of of `field2type_and_format` in place, which was a bit confusing when I first looked at it, because you have to scan up a few hundred lines to figure out what's going on. 

It might be better to remove `DateTime` from the field mapping and call this function from `field2type_and_format` for that special case.

TheBigRoomXXL:
@lafrech Thank you for your fast feeback, I just needed to know if there was something expected of me. 


@Bangertm  You are right, finding the result of the first condition might be a little confusing. I see two options to fix that:

1 - just add a comment explaning the expected result:
```python
if field.format == "iso" or field.format is None:
  # Will return { "type": "string", "format": "date-time" } as specified inside DEFAULT_FIELD_MAPPING
  pass
```

2 - Force the result here (as you proposed if I understood you correctly?):

```python
if field.format == "iso" or field.format is None:
ret = {
  "type": "string",
  "format": "date-time",
}
```

The funny thing is that if you do that you suddenly break the tests for date format. It looks like `isinstance(field, marshmallow.fields.DateTime)` return true if `field`  of type marshmallow.fields.Date. I guess it kind of make sense but it is a little weird and **I don't know if it's expected behavior**.

To avoid any confusion in the future, it would probably be wise to `and not isinstance(field, marshmallow.fields.Date)` to the first if statement (line 527). 

Regarding the original problem I think the option 1 is best as it doesn't break the usual flow and is still quite clear.

TheBigRoomXXL:
So I made 3 changes to the original merge request: 

- I added `and not isinstance(field, marshmallow.fields.Date )` to avoid any false positive
- I added a comment describing the output of the iso format so that it doesn't override any existing logic. 
- I changed `{"type": "integer", "format": "timestamp",}` to `{"type": "number", "format": "float",}` because marshamllow return float, not integer.

src/apispec/ext/marshmallow/field_converter.py
```diff
@@ -516,6 +517,48 @@ def enum2properties(self, field, **kwargs: typing.Any) -> dict:
...
+            elif field.format == "rfc":
+                ret = {
+                    "type": "string",
+                    "format": None,
+                    "example": "Wed, 02 Oct 2002 13:00:00 GMT",
+                    "pattern": r"((Mon|Tue|Wed|Thu|Fri|Sat|Sun), ){0,1}\d{2} "
+                    + r"(Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec) \d{4} \d{2}:\d{2}:\d{2} "
+                    + r"(UT|GMT|EST|EDT|CST|CDT|MST|MDT|PST|PDT|(Z|A|M|N)|(\+|-)\d{4})",
+                }
+            elif "timestamp" in field.format:
```
lafrech: I'd use strict equality, specifying both `timestamp` and `timestamp_ms`. It is clearer to read, especially the `if "_ms" in field.format` part. And more future-proof (if some other format gets added that contains `"timestamp"`.

tests/test_ext_marshmallow_field.py
```diff
@@ -26,7 +26,7 @@ def test_field2choices_preserving_order(openapi):
         (fields.Boolean, "boolean"),
         (fields.Bool, "boolean"),
         (fields.UUID, "string"),
-        (fields.DateTime, "string"),
+        (fields.DateTime, ["string", "integer"]),
```
lafrech: Why? This is overridden in `datetime2properties` for the cases other than ISO, which is string. Right? (You may want to add a comment here as well to explain that.)

src/apispec/ext/marshmallow/field_converter.py
```diff
@@ -516,6 +517,48 @@ def enum2properties(self, field, **kwargs: typing.Any) -> dict:
...
+                # as specified inside DEFAULT_FIELD_MAPPING
+                pass
+            elif field.format == "rfc":
+                ret = {
+                    "type": "string",
+                    "format": None,
+                    "example": "Wed, 02 Oct 2002 13:00:00 GMT",
+                    "pattern": r"((Mon|Tue|Wed|Thu|Fri|Sat|Sun), ){0,1}\d{2} "
+                    + r"(Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec) \d{4} \d{2}:\d{2}:\d{2} "
+                    + r"(UT|GMT|EST|EDT|CST|CDT|MST|MDT|PST|PDT|(Z|A|M|N)|(\+|-)\d{4})",
```
lafrech, line 543: Out of curiosity. Did you make this regex or did you get it somewhere? If so, specifying the source for further reference would be nice.

lafrech (changes requested):
Thanks! Sorry about the delay. Just a few minor questions/comments.

src/apispec/ext/marshmallow/field_converter.py
TheBigRoomXXL: Its my own. It's not perfect, there is some possible false positive. Given that this regex is only for documentation and not validation I tried to make it descriptive and relatively short (as much as possible for a regex...) 

Some examples are available [on regex101](https://regex101.com/r/hOctM6/1)

src/apispec/ext/marshmallow/field_converter.py
TheBigRoomXXL: It 's done.

TheBigRoomXXL:
Thank you for your review.  I will be more reactive this time if you need any other change!

sloria:
Made some minor fixups so this is mergeable. Thanks so much for the contribution @TheBigRoomXXL!

PR #828, opened 2023-03-06
fix: resolve refs for callbacks

codectl:
Schemas present in ```callbacks``` property in OAS3 should also be resolved to refs


lafrech (changes requested):
Looks good to me.

Although the example in https://swagger.io/docs/specification/callbacks/ don't show it explicitly, I imagine callbacks could have response schemas, perhaps also parameters (not only request body).

I believe the code would work but do you think you could complete the test to check it?

Thanks.

Also please add yourself to contributors file.

codectl:
@lafrech yep no problem

lafrech (approved):

Issue #833, opened 2023-03-30
Incorrect handling of marshmallow nested schema with allow_none=True in OpenAPI 3.1


jc-harrison:
If I define a marshmallow schema like this:
```python
from marshmallow import fields, Schema

class FooSchema(Schema):
    bar = fields.Integer(required=True)

class MySchema(Schema):
    foo = fields.Nested(FooSchema, allow_none=True)
```
it behaves as I intend - `MySchema` can load objects with a `foo` field that either matches `FooSchema` or is null, but `FooSchema` itself doesn't allow null values:
```python
>>> MySchema().loads('{"foo": {"bar": 1}}')
{'foo': {'bar': 1}}
>>> MySchema().loads('{"foo": null}')
{'foo': None}
>>> FooSchema().loads('null')
marshmallow.exceptions.ValidationError: {'_schema': ['Invalid input type.']}
```

But if I generate an API spec with this schema, using openapi version 3.1:
```python
from apispec import APISpec
from apispec.ext.marshmallow import MarshmallowPlugin

spec = APISpec(
    title="My API Spec",
    version="1.0.0",
    openapi_version="3.1.0",
    plugins=[MarshmallowPlugin()],
)
spec.components.schema("MySchema", schema=MySchema)
```
the resulting API spec is not equivalent to my marshmallow schema:
```python
>>> print(json.dumps(spec.to_dict()["components"]["schemas"], indent=2))
{
  "Foo": {
    "type": "object",
    "properties": {
      "bar": {
        "type": "integer"
      }
    },
    "required": [
      "bar"
    ]
  },
  "MySchema": {
    "type": "object",
    "properties": {
      "foo": {
        "type": [
          "null"
        ],
        "allOf": [
          {
            "$ref": "#/components/schemas/Foo"
          }
        ]
      }
    }
  }
}
```
This spec says that the value of `foo` _must_ be null, AND must be a `Foo` object, so in fact there is no valid input value here.

I think a correct spec in this instance should contain:
```
{
  "Foo": {
    "type": "object",
    "properties": {
      "bar": {
        "type": "integer"
      }
    },
    "required": [
      "bar"
    ]
  },
  "MySchema": {
    "type": "object",
    "properties": {
      "foo": {
        "anyOf": [
          {
            "$ref": "#/components/schemas/Foo"
          },
          {
            "type": "null"
          }
        ]
      }
    }
  }
}
```

I'm using apispec 6.3.0 and marshmallow 3.19.0

[...]

PR #841, opened 2023-05-25
fix: nested optionals now combine the ref with type null in anyOf

jmgoncalves:
Proposed fix for https://github.com/marshmallow-code/apispec/issues/833

 I tried to do what was described by a contributor https://github.com/marshmallow-code/apispec/issues/833#issuecomment-1494552232 but when the nullable handling is moved to after nested and pluck handling, the context seems to be different - `ret` looks like `{ '$ref': '#/components/type' }` rather than `{ 'allOf': { '$ref': '#/components/type' }, ...}`

Addressing this at the nested handling seemed more self-contained.

### DIFF
--- a/src/apispec/ext/marshmallow/field_converter.py
+++ b/src/apispec/ext/marshmallow/field_converter.py
@@ -439,7 +439,16 @@ class FieldConverterMixin:
         ):
             schema_dict = self.resolve_nested_schema(field.schema)  # type:ignore
             if ret and "$ref" in schema_dict:
-                ret.update({"allOf": [schema_dict]})
+                if "type" in ret:
+                    type = ret.pop("type")
+                    if isinstance(type, list):
+                        types = [{"type": t} for t in type]
+                    else:
+                        types = [{"type": type}]
+                    types.append(schema_dict)
+                    ret.update({"anyOf": types})
+                else:
+                    ret.update({"allOf": [schema_dict]})
             else:
                 ret.update(schema_dict)
         return ret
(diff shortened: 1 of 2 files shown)


lafrech:
@bangterm, WDYT? Are there cases in [your comment](https://github.com/marshmallow-code/apispec/issues/833#issuecomment-1494552232) that are not addressed here? Pluck field?

Bangertm:
@lafrech I'll take a closer look tonight. The cases that jumped out to me were Pluck and the when a schema is defined inline (i.e. the resolver returns `none`.

Mike

Bangertm:
[This](https://github.com/marshmallow-code/apispec/commit/da7e825983b450bcf9c68ff510cc6158300c5be3) is more what I had in mind. I'm not sure if the generated properties are accurate for each case so those need review. Looking to handle: 
1. Pluck
2. Nested defined inline
3. Nested as a reference
4. Nested with other properties

Moving `field2nullable` execution to the end takes care of 1 & 2 and then the two conditions added to `field2nullable` take care of 3 and 4. 

The thing that's not done here is making it so that user defined attribute functions execute prior to `field2nullable` so users don't have to duplicate nullability logic. Whether we do that or not requires some thought and would arguably be a breaking change.

lafrech:
Thanks for your work on this.

I don't see any other way. We can ensure `field2nullable` runs last by taking it out of the list and calling it explicitly.

I'm not sure this should be considered a breaking change, I mean any change in existing functions may already break the user-defined functions that run after them. In any case, major releases are cheap so this shouldn't be a blocker.

jmgoncalves:
Alternative preferred solution #845 works well for me so closing this one

PR #851, opened 2023-07-17
Fix conversion of deprecated flag on parameters

### LINKED ISSUE #850: Deprecated flag is not configured correctly on query parameter objects
When using the `metadata` parameter on a Marshmallow field, the `deprecated` field is not properly configured on parameter objects with a location of `"query"` in the generated spec (the `description` field is set as expected, but not `deprecated`):

```python
from apispec import APISpec


expected_result = """paths:
  /endpoint:
    get:
      parameters:
      - in: query
        name: bar
        description: A field called bar
        deprecated: true
        schema:
          type: string
        required: true
info:
  title: My API Spec
  version: 1.0.0
openapi: 3.1.0
"""

actual_result = """paths:
  /endpoint:
    get:
      parameters:
      - in: query
        name: bar
        description: A field called bar
        schema:
          type: string
          deprecated: true
        required: true
info:
  title: My API Spec
  version: 1.0.0
openapi: 3.1.0
"""


class TestSchema(Schema):
    bar = fields.Str(
        required=True,
        metadata={"description": "A field called bar", "deprecated": True},
    )


spec = APISpec(
    title="My API Spec",
    version="1.0.0",
    openapi_version="3.1.0",
)
spec.path(
    "/endpoint",
    operations={"get": {"parameters": [{"in": "query", "schema": TestSchema}]}},
)

assert spec.to_yaml() == actual_result  # passes when it should fail
assert spec.to_yaml() == expected_result  # fails when it should pass
```

The `deprecated` field works correctly for request and response bodies, but not for query parameters. It seems to me that setting `deprecated = True` in the metadata on a field instance should mark it as deprecated on the parameter object itself and not the nested schema object, since the schema itself is not what is deprecated.

It looks like the `OpenAPIConverter._field2parameter` method might need updating to promote the `deprecated` flag from the nested `schema` object into the parent object, similar to how `description` is handled.

tsokalski:
Fixes #850. This promotes the `deprecated` value on to the main parameter object rather than setting it on the contained `schema` object.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -77,3 +77,4 @@ Contributors (chronological)
 - Edwin Erdmanis `@vorticity <https://github.com/vorticity>`_
 - Mounier Florian `@paradoxxxzero <https://github.com/paradoxxxzero>`_
 - Renato Damas `@codectl <https://github.com/codectl>`_
+- Tayler Sokalski `@tsokalski <https://github.com/tsokalski>`_
--- a/src/apispec/ext/marshmallow/openapi.py
+++ b/src/apispec/ext/marshmallow/openapi.py
@@ -197,6 +197,8 @@ class OpenAPIConverter(FieldConverterMixin):
         else:
             if "description" in prop:
                 ret["description"] = prop.pop("description")
+            if "deprecated" in prop:
+                ret["deprecated"] = prop.pop("deprecated")
             ret["schema"] = prop
 
         for param_attr_func in self.parameter_attribute_functions:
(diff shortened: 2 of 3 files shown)


lafrech:
Looks good, thanks. Sorry about the delay.

CI issue should be fixed in dev branch.

tsokalski:
@lafrech any thoughts on releasing this sometime soon? Maybe in a 6.3.1 release? I know it's not big, but I'd love to remove the workaround in my codebase 😄.

lafrech:
Oh, right. Done.

PR #856, opened 2023-08-29
Fix schema name that has end with space for redoc ui and elements ui

duchuyvp:
I strip `component_id` before register a new schema when to prevent its name has some bad space, generated spec has schama name that end with space like 'Schema name ' doesn't work with `redoc` ui (by ReDoc) and `elements` ui (by Stoplight)

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -78,3 +78,4 @@ Contributors (chronological)
 - Mounier Florian `@paradoxxxzero <https://github.com/paradoxxxzero>`_
 - Renato Damas `@codectl <https://github.com/codectl>`_
 - Tayler Sokalski `@tsokalski <https://github.com/tsokalski>`_
+- Luna Lovegood `@duchuyvp <https://github.com/duchuyvp>`_
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,8 @@ Features:
 - ``MarshmallowPlugin``: Support different datetime formats
   for ``marshmallow.fields.DateTime`` fields (:issue:`814`).
   Thanks :user:`TheBigRoomXXL` for the suggestion and PR.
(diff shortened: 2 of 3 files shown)


lafrech:
AFAIU, the space in the schema name a user error that may be a bit tricky to spot but that should be easily fixed.

I may be convinced otherwise but I don't think apispec should engage in this. We can't workaround all typos.

duchuyvp:
You have point out that maybe it's a user error, I did try named a schema like "Pet In Schema" and the error occur, cause after remove postfix "Schema" in that name, it's stay "Pet In " and this does not look like user error or typo. Even I can simply fix this by replace the name with "Pet In" or "Pet InSchema" but the second way may looks like typo error.

lafrech:
Good point. We could improve the schema name generator to strip spaces there.

How do you create the schema?

Schemas are generally created as classes so you can't have spaces in their names.

```py
class PetSchema(Schema):
    ...
```

duchuyvp:
Sorry about delay. 

I used `from_dict` to generate schema.

```python
from marshmallow.fields import String

pet_schema = Schema.from_dict({"name": String()}, name="Pet In Schema")
```

lafrech:
OK. 

Users probably shouldn't give names with spaces in the first place but since it is not forbidden, I have no objection to stripping the generated name in the default name resolver: https://github.com/marshmallow-code/apispec/blob/cb90ed2931a5450a652fb32c34f082730d2f2dd1/src/apispec/ext/marshmallow/__init__.py#L87-L93.

duchuyvp:
Thanks for your consider. Would the branch be merge?

duchuyvp:
pre-commit.ci run

sloria:
Made some minor fixups so it's ready to merge. Thanks for the PR!

PR #862, opened 2023-10-01
Set the `format` of `marshmallow.fields.Float` fields to `float`

ddorian:
See `float` format for number data types https://swagger.io/docs/specification/data-models/data-types/

### DIFF
--- a/src/apispec/ext/marshmallow/field_converter.py
+++ b/src/apispec/ext/marshmallow/field_converter.py
@@ -22,7 +22,7 @@ from marshmallow.orderedset import OrderedSet
 DEFAULT_FIELD_MAPPING: dict[type, tuple[str | None, str | None]] = {
     marshmallow.fields.Integer: ("integer", None),
     marshmallow.fields.Number: ("number", None),
-    marshmallow.fields.Float: ("number", None),
+    marshmallow.fields.Float: ("number", "float"),
     marshmallow.fields.Decimal: ("number", None),
     marshmallow.fields.String: ("string", None),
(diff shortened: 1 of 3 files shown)


lafrech:
Shouldn't it be `double`?

https://swagger.io/docs/specification/data-models/data-types/

Or system-dependent?

https://docs.python.org/3/library/stdtypes.html#numeric-types-int-float-complex

> Floating point numbers are usually implemented using double in C; information about the precision and internal representation of floating point numbers for the machine on which your program is running is available in [sys.float_info](https://docs.python.org/3/library/sys.html#sys.float_info).

We can't rely (solely) on `sys.float.info` as the doc may be generated on another machine than the one running the API.

ddorian:
> Shouldn't it be double?

Yes.

> We can't rely (solely) on sys.float.info as the doc may be generated on another machine than the one running the API.

That is the problem with the dev IMO. What do you propose?

lafrech:
Since we can't tell, I'd say leave things as is and let the user override if needed.

We may want to document this, though.

ddorian:
Maybe we can add a validator to make the "type" (example on Function/Method fields) or format (Float,Number,etc) required ?

lafrech:
I'm sorry, I don't understand what you mean.

ddorian:
> I'm sorry, I don't understand what you mean.

I've added a validator that checks that all Float fields should have "type" defined, just to be "consistent" on the openapi output.
I've done that for "FunctionField" & "MethodField", but made "type" required there.

Issue #869, opened 2023-11-29
Incorrectly serialized format when using a range validator on a Decimal Marshmallow field


intZero:
Hello! One of our Marshmallow serializers includes this field:
`conversion_factor = fields.Decimal(required=True, validate=[Range(min=0)])`

In 6.1.0+ this serializes to:
```
conversion_factor:
  type: number
  minimum: !!python/object/apply:decimal.Decimal
  - '0'
```
I've traced the issue to this section of code in field2range() in field_converter.py:
```
# Serialize min/max values with the field to which the validator is applied
return {
    k: field._serialize(v, None, None)
    for k, v in make_min_max_attributes(validators, min_attr, max_attr).items()
}
```
The main problem is that in _serialize for a Number:
```
"""Return a string if `self.as_string=True`, otherwise return this field's `num_type`."""
```
And since for a Decimal field the default self.as_string=False (standard usage) we get "!!python/object/apply:decimal.Decimal". I suggest that for field types where _serialize does this you instead run _to_string which will format the number accordingly:
```
conversion_factor:
  type: number
  minimum: '0.000000'
```

lafrech:
Thanks for reporting. Would you like to work on this (change + test)?

intZero:
Yeah I've made the change and written a couple of tests for it, should I open a PR against dev?

lafrech:
Yes, please !

intZero:
Cool, could I get access to create branch intzero/fix-decimal-range please?

lafrech:
The procedure is to push the branch to your GH fork, then open the PR.

intZero:
My bad! Have opened a PR here https://github.com/marshmallow-code/apispec/pull/870

lafrech:
Sorry for not spotting this any sooner, but I think the way to go is to set `as_string=True` in `Decimal` field, rather than adding field-specific code in a generic function.

See 

- https://github.com/marshmallow-code/flask-smorest/issues/517
- https://marshmallow.readthedocs.io/en/stable/marshmallow.fields.html#marshmallow.fields.Decimal

intZero:
Yeah I saw your reply to the PR. I really don't think that asking people to change their business logic by changing the serializer to use a different variable type is a solution imho! Secondly, it isn't a solution anyway as you will end up with:
```  
type: number
  minimum: '0'
  maximum: '1000000'
```
[...]

PR #870, opened 2023-12-01
Fix decimal range validators

intZero:
"Native" numerical types such as Int and Float serialize correctly but Decimal needs to specifically be serialized to a float for the renderer.

From this issue: https://github.com/marshmallow-code/apispec/issues/869

Previously, this field
```
conversion_factor = fields.Decimal(required=True, validate=[Range(min=0, max=1000000)])
```

would render to:
```
conversion_factor:
  type: number
  minimum: !!python/object/apply:decimal.Decimal
  - '0'
  maximum: !!python/object/apply:decimal.Decimal
  - '1000000'
```

Now it will render to:
```
conversion_factor:
  type: number
  minimum: 0.0
  maximum: 1000000.0
```


lafrech:
I think it is the same issue as flask-smorest's https://github.com/marshmallow-code/flask-smorest/issues/517, except in flask-smorest we want to solve it by using flask json serializer.

Did you set `as_string=True`? See docstring in `Decimal` field.

PR #888, opened 2024-02-26
Fix nullable enums

### LINKED ISSUE #812: Incorrect handling of nullable enums in marshmallow converter
According to the [OpenAPI specification](https://swagger.io/docs/specification/data-models/enums/#nullable), nullable enums must list `null` in the allowed options. apispec doesn't do this by default, so in my projects I've had to patch this behaviour in with the following attribute function:

```py
def field_to_properties(self, field, ret):
    enum = ret.get('enum', [])
    nullable = ret.get('nullable', False)
    if enum and nullable and None not in enum:
        return {'enum': [*enum, None]}
    return {}
```

It would be nice if apispec implemented this.

kolditz-senec:
This PR fixes issue #812. `null` is added to the list of enum values if the enum is nullable.

Let me know if something can be improved or feel free to make changes on this branch as you see fit. Thanks for maintaining this project!


sloria:
Thank you! updated and merged in #889

PR #889, opened 2024-02-26
Fix nullable enums

sloria:


lafrech:
I was about to merge #888. Thanks @sloria for taking care of this.

PR #845, opened 2023-06-23
Fix nested nullable

Bangertm:
This fixes #833. `field2nullable` executes after other attribute functions including user added attribute functions and now has logic for handling cases with a reference object and an `allOf` object.

PR #898, opened 2024-04-08
Fix nested nullable for 3.1.0

kolditz-senec:
Fixes issue #833. Similar to PR #845, but 

* prevents "nullable" as a sibling of "$ref",
* does not change behavior for versions <3.1.0 (because the [semantics of nullable was ambiguous before 3.0.3](https://github.com/OAI/OpenAPI-Specification/blob/main/proposals/2019-10-31-Clarify-Nullable.md)), and
* does not change the evaluation order of custom attribute functions (custom functions get the results of the built-in functions passed in as an argument and are responsible for generating valid results).

### DIFF
--- a/tests/test_ext_marshmallow_field.py
+++ b/tests/test_ext_marshmallow_field.py
@@ -206,9 +206,10 @@ def test_nested_nullable(spec_fixture):
 
     field = fields.Nested(Child, allow_none=True)
     res = spec_fixture.openapi.field2property(field)
-    if spec_fixture.openapi.openapi_version.major < 3:
+    version = spec_fixture.openapi.openapi_version
+    if version.major < 3:
         assert res == {"$ref": "#/definitions/Child", "x-nullable": True}
-    elif spec_fixture.openapi.openapi_version.minor < 1:
+    elif version.major == 3 and version.minor < 1:
         assert res == {
             "nullable": True,
             "allOf": [{"$ref": "#/components/schemas/Child"}],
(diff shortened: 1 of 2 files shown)


tests/test_ext_marshmallow_field.py
```diff
@@ -199,6 +199,41 @@ def test_field_with_allow_none(spec_fixture):
...
+@pytest.mark.parametrize("spec_fixture", ("2.0", "3.0.0", "3.1.0"), indirect=True)
+def test_nested_nullable(spec_fixture):
+    class Child(Schema):
+        name = fields.Str()
+
+    field = fields.Nested(Child, allow_none=True)
+    res = spec_fixture.openapi.field2property(field)
+    if spec_fixture.openapi.openapi_version.major < 3:
+        assert res == {"$ref": "#/definitions/Child", "x-nullable": True}
+    elif spec_fixture.openapi.openapi_version.minor < 1:
```
sloria: this won't properly handle version 4.1, 5.1, etc. 

can you do a more precise check here for OAI version < 3.1?

note: i realize this is how it's done in `field2nullable`, so that should also be changed, but that's not a blocker for merging this PR.

tests/test_ext_marshmallow_field.py
```diff
@@ -199,6 +199,41 @@ def test_field_with_allow_none(spec_fixture):
...
+@pytest.mark.parametrize("spec_fixture", ("2.0", "3.0.0", "3.1.0"), indirect=True)
+def test_nullable_pluck(spec_fixture):
+    class Example(Schema):
+        name = fields.Str()
+
+    field = fields.Pluck(Example, "name", allow_none=True)
+    res = spec_fixture.openapi.field2property(field)
+    if spec_fixture.openapi.openapi_version.major < 3:
+        assert res == {"type": "string", "x-nullable": True}
+    elif spec_fixture.openapi.openapi_version.minor < 1:
```
sloria: same as above: handle < 3.1 more precisely

sloria (changes requested):
thank you! just minor update to make

sloria:
thanks! this is merged in https://github.com/marshmallow-code/apispec/pull/904

sloria:
released in 6.6.1

PR #937, opened 2024-09-02
Support of DateTime and Date formats

expobrain:
Adds full support for rendering [Date](https://marshmallow.readthedocs.io/en/stable/marshmallow.fields.html#marshmallow.fields.Date) and [DateTime](https://marshmallow.readthedocs.io/en/stable/marshmallow.fields.html#marshmallow.fields.DateTime) Marshmallow fields with each of their `format` values.

This solve the issue of `Date` and `DateTime` with a non-default format being rendered with `format: null` instead of `format: date` and `format: date-time` respectively.


src/apispec/ext/marshmallow/field_converter.py
```diff
@@ -574,16 +573,62 @@ def datetime2properties(self, field, **kwargs: typing.Any) -> dict:
                     "example": "1676451277514.654",
                     "min": "0",
                 }
+            elif field.format is not None:
```
lafrech: At this point, `field.format` can't be `None` due to line 549, right?

lafrech (commented):
Hi. Thanks for digging into this.

I didn't take to time to review thoroughly but here's a question already.

Any idea why it stopped working as expected in version 6.4?

bhperry:
Also, format and pattern should not ever by None, according to openapi spec they must be strings. Probably should be removed from the ret dict when they are None.

PR #953, opened 2024-11-02
fix: Nullable nested fields comply with 3.0 spec

### LINKED ISSUE #952: Incorrect handling of marshmallow nested schema with allow_none=True in OpenAPI 3.0
This is a spinoff of #833, but for OpenAPI 3.0.

The following schemas:

```py
from apispec import APISpec

spec = APISpec(
    title="My API Spec",
    version="1.0.0",
    openapi_version="3.0.3",
)


class FooSchema(Schema):
    bar = fields.Integer(required=True)


class MySchema(Schema):
    foo = fields.Nested(FooSchema, allow_none=True)
```

Generates the following OpenAPI spec:

```py
{
  "Foo": {
    "type": "object",
    "properties": {
      "bar": {
        "type": "integer"
      }
    },
    "required": [
      "bar"
    ]
  },
  "MySchema": {
    "type": "object",
    "properties": {
      "foo": {
        "nullable": true,
        "allOf": [
          {
            "$ref": "#/components/schemas/Foo"
          }
        ]
      }
    }
  }
}
```

This looks correct at first glance, but isn't compliant with the `3.0.3` revision of OpenAPI, which clarified how `nullable` should work. See [Clarify Semantics of nullable in OpenAPI 3.0](https://github.com/OAI/OpenAPI-Specification/blob/main/proposals/2019-10-31-Clarify-Nullable.md#detailed-design) for details.

> nullable: true is only meaningful in combination with a type assertion specified in the same Schema Object. nullable acts as a type modifier, allowing null in addition to the specified type.
>
> nullable: true operates within a single Schema Object. It does not "override" or otherwise compete with supertype or subtype schemas defined with allOf or other applicators. It cannot be directly "inherited" through those applicators, and it cannot be applied to an inherited type constraint.

The above spec causes issues when used with data validators such as [openapi-core](https://github.com/python-openapi/openapi-core).

I think that the correct version of the above spec would look like:

```py
{
  "Foo": {
    "type": "object",
    "properties": {
      "bar": {
        "type": "integer"
      }
    },
    "required": [
      "bar"
    ]
  },
  "MySchema": {
    "type": "object",
    "properties": {
      "foo": {
        "anyOf": [
          {
            "type": "object",
            "nullable": true
          },
          {
            "$ref": "#/components/schemas/Foo"
          }
        ]
      }
    }
  }
}
```

I've tested this with apispec 6.7.0 and marshmallow 3.23.1. I also have investigated a fix for this and am happy to open a PR if the above result is acceptable.

ShepleySound:
Fixes #952

### DIFF
--- a/src/apispec/ext/marshmallow/field_converter.py
+++ b/src/apispec/ext/marshmallow/field_converter.py
@@ -301,9 +301,13 @@ class FieldConverterMixin:
             if self.openapi_version.major < 3:
                 attributes["x-nullable"] = True
             elif self.openapi_version.minor < 1:
-                attributes["nullable"] = True
                 if "$ref" in ret:
-                    attributes["allOf"] = [{"$ref": ret.pop("$ref")}]
+                    attributes["anyOf"] = [
+                        {"type": "object", "nullable": True},
+                        {"$ref": ret.pop("$ref")},
+                    ]
+                else:
+                    attributes["nullable"] = True
             else:
                 if "$ref" in ret:
                     attributes["anyOf"] = [{"$ref": ret.pop("$ref")}, {"type": "null"}]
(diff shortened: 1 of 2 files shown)


lafrech (approved):
After reading [the blog post you linked to](https://danott.dev/thoughts/openapi-nullable-object-anyof-oneof-allof) and going through your comments again, I think you're right.

PR #961, opened 2024-12-06
Improve handling of nullable Raw fields for OAS 3.1.0

### LINKED ISSUE #960: Nullable `Raw` fields are not handled well for OAS 3.1.0
When converting a nullable `Raw` Marshmallow field into its corresponding OpenAPI representation for OAS 3.1.0, the generated representation does not match the behaviour of the actual schema/field:

```python
import jsonschema

from apispec import APISpec
from apispec.ext.marshmallow import OpenAPIConverter


class TestSchema(Schema):
    raw_field = fields.Raw(allow_none=True)


spec = APISpec(title="Pets", version="0.1", openapi_version="3.1.0")
converter = OpenAPIConverter(
    openapi_version="3.1.0", schema_name_resolver=None, spec=spec
)
json_schema = converter.schema2jsonschema(TestSchema)
data = {"raw_field": "test"}
TestSchema().load(data)  # Works, as expected
jsonschema.validate(data, json_schema)  # Fails unexpectedly
```

The output of printing the `json_schema` variable above is the following:
```python
{'type': 'object', 'properties': {'raw_field': {'type': ['null']}}}
```

It appears the `FieldConverterMixin`'s `field2nullable` is appending `null` to the list of supported types for OAS 3.1.0, which makes sense in many cases, but when there's no pre-existing `type` value defined, the result is just `['null']`, which means that the field can only be `null`. This doesn't match the behaviour of the defined schema.

I've implemented a workaround for this where I've overridden the `DEFAULT_FIELD_MAPPING` for `marshmallow.fields.Raw` to map to all supported JSON types (`["boolean", "object", "array", "number", "string"]`), but it's a bit of a hack, and would be nice if the library could support this a bit better.

tsokalski:
Modified the implementation of `field2nullable` for OAS 3.1.0 to only append `"null"` to the list of types if there is already a type for the field. This prevents issues in which a nullable field with no specific associated OAS type can get bound to `"type": ["null"]`, which is overly restrictive.

Fixes #960.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,6 +106,3 @@ venv.bak/
 
 # ruff
 .ruff_cache/
-
-# PyCharm
-.idea
--- a/.gitignore
+++ b/.gitignore
@@ -106,3 +106,6 @@ venv.bak/
 
 # ruff
 .ruff_cache/
+
+# PyCharm
+.idea
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,8 +1,13 @@
 Changelog
 ---------
 
-6.9.0 (unreleased)
+6.8.1 (unreleased)
 ******************
+
+Bug fixes:
+
+- Fix handling of nullable Raw fields for OAS 3.1.0 (:issue:`960`).
+  Thanks :user:`tsokalski` for reporting and fixing.
 
 Support:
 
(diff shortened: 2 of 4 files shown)


lafrech:
Thanks for contriburing.

Please put editor-specfic stuff in your own global .gitignore.

https://stackoverflow.com/questions/7335420/global-git-ignore

tsokalski:
Sorry about that, I didn't realize a global .gitignore was a thing - thanks for the tip!

I've removed the offending section from the `.gitignore` file. Please let me know if you have any other feedback.

tsokalski:
Bump @lafrech - any additional feedback on this?

PR #972, opened 2025-04-10
Use set instead of list for `exclude` in `filter_excluded_fields` fucntion

allrob23:
This PR updates the `filter_excluded_fields` function to use a `set` instead of a `list` for the `exclude` variable. While the number of elements in `Meta.exclude` and `Meta.dump_only` may not be large, this change still provides a small optimization by improving membership checks from O(N) to O(1). 

The overall behavior of the function remains unchanged. Fields are still filtered based on `Meta.exclude` and `dump_only` as before. The `exclude` variable, now a `set`, is used in the `if` condition to check field membership, benefiting from the faster lookups without altering the logic.


sloria:
thanks!

PR #976, opened 2025-05-10
Avoid overwriting `additionalProperties` with the default `{}`

0x0400:
A few versions ago, the output of `Dict()` omitted `additionalProperties` entirely and I used `Dict(metadata={"additionalProperties": True})` as a workaround. Now the output of `additionalProperties: {}` will breakdown our internal tools.
Although `additionalProperties: {}` is equivalent to `additionalProperties: true`, I think it's better not to overwrite the manually set value.


sloria:
Good catch! thanks for the pr

PR #977, opened 2025-05-12
Update core.py

ShekharDipankar:
Add description field in APISPEC class to define description of swagger


sloria:
This is a breaking change because it adds a new required argument. Passing `description` is already suppported by passing `info`. See the first example in https://apispec.readthedocs.io/en/latest/quickstart.html#basic-usage

PR #982, opened 2025-06-17
feat: add `list_as_array` flag to MarshmallowPlugin for array-root requestBody

nomi3:
OpenAPI 3.x fully supports request bodies whose root schema is an array, but apispec + marshmallow currently cannot generate such schemas automatically. Projects must hand-write YAML/dict snippets for every endpoint that needs it.

This PR introduces an opt-in flag—list_as_array—to MarshmallowPlugin that lets the plugin emit type: array schemas automatically when:

- a plain fields.List(...) is supplied as the schema, or
- a Schema (class or instance) contains exactly one fields.List field with data_key=None.
 
The default behaviour remains unchanged (object-root), so the change is non-breaking.

### DIFF
--- a/src/apispec/ext/marshmallow/openapi.py
+++ b/src/apispec/ext/marshmallow/openapi.py
@@ -56,6 +56,7 @@ class OpenAPIConverter(FieldConverterMixin):
         openapi_version: Version | str,
         schema_name_resolver,
         spec: APISpec,
+        list_as_array: bool = False,
     ) -> None:
         self.openapi_version = (
             Version(openapi_version)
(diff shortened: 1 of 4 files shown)


src/apispec/ext/marshmallow/openapi.py
```diff
@@ -108,6 +110,30 @@ def resolve_nested_schema(self, schema):
 
         :param schema: schema to add to the spec
         """
+        if isinstance(schema, marshmallow.fields.List) and getattr(
```
lafrech: Why use `getattr`? Isn't `list_as_array` always defined?

nomi3: You're right! I've removed the `getattr` calls and now use `self.list_as_array` directly.

lafrech:
Does it have to be optional?

I mean is there any reason to pass an array if this flag is not set?

If so, wouldn't that use case be broken if this flag is set to accommodate for a given resource?

PR #988, opened 2025-09-03
Set additionnalProperties to False if unknown is not set / support setting unknown at schema instantiation

### LINKED ISSUE #821: additionalProperties=false is not set when using default Marshmallow schema options
The default value for the Marshmallow schema option `unknown` is `RAISE`. I expected this to result in `additionalProperties: false` being set on the generated schema. Unfortunately, this doesn't currently happen. Unless the `unknown` option is set explicitly, no `additionalProperties` keyword is added to the generated schema.

Is adding support for this feasible? I don't really want to have to manually add the necessary metadata to all my schemas.

lafrech:
- Fixes #821: Set `additionnalProperties` to `False` if `unknown` is not set
- Set `additionnalProperties` according to instance `unknown` rather than class `Meta`, to support setting it at instantiation

This could be wrong if `unknown` is set as `load` parameter, but I don't think we can address this.

### DIFF
--- a/src/apispec/ext/marshmallow/openapi.py
+++ b/src/apispec/ext/marshmallow/openapi.py
@@ -258,7 +258,9 @@ class OpenAPIConverter(FieldConverterMixin):
             jsonschema["title"] = Meta.title
         if hasattr(Meta, "description"):
             jsonschema["description"] = Meta.description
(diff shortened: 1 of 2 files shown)


sloria (approved):
nice!

PR #990, opened 2025-09-19
Update TimeDelta field type to float for Marshmallow v4

LewisHaley:
In Marshmallow v3.17.0, up until v4, the `TimeDelta` field could be serialized into either an integer or a float, depending on user specification of the field. [1] In the case the user set `serialization_type=float`, apispec would generate the wrong JSON type in the API docs.

With Marshmallow v4, this was removed and `TimeDelta` fields *always* serialize to floats, in which case apispec always generates the wrong docs.

This commit updates the default field mapping anticipating Marshmallow v4's float serialization, with some additional support added for <v4 by looking for the deprecated and removed `serialization_type` attribute on the field object.

[1] https://marshmallow.readthedocs.io/en/stable/marshmallow.fields.html#marshmallow.fields.TimeDelta

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -87,3 +87,4 @@ Contributors (chronological)
 - Robert Shepley `@ShepleySound <https://github.com/ShepleySound>`_
 - Robin `@allrob23 <https://github.com/allrob23>`_
 - Xingang Zhang `@0x0400 <https://github.com/0x0400>`_
+- Lewis Haley `@LewisHaley <https://github.com/LewisHaley>`_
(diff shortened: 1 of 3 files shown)


lafrech:
Thanks.

Tests fail on MA3.

I'm not sure we need to manage the field misuse case but I don't mind. Let's keep this.

LewisHaley:
> I'm not sure we need to manage the field misuse case but I don't mind. Let's keep this.

I ended up stripping it out because when using marshmallow.fields.TimeDelta constructor properly, it rejects bad input.

lafrech:
Looks good.

Please add yourself to `AUTHORS.rst`.

lafrech:
Thanks!

Released in 6.8.4.

Issue #579, opened 2020-07-07
OpenAPI 3.1 support


lafrech:
There's been a 3.1.0 RC release and they decided to make breaking changes.

https://github.com/OAI/OpenAPI-Specification/releases

It didn't go through all the changes yet to see if and how we are impacted.

At least, since they don't follow semver anymore and things can change in minor versions, we may have to modify some functions to pass the version object or string rather than just the major version as int.

Also I did a quick search to find an EOL for OAS v2 and I didn't find anything. Maybe there is no need for an EOL on a spec, as opposed to a code library.

Bangertm:
I took a look at the breaking changes section: 
> The specification versioning no longer follows SemVer.
The nullable keyword has been removed from the Schema Object (null can be used as a type value).
exclusiveMaximum and exclusiveMinimum cannot accept boolean values (following JSON Schema).
Due to the compliance with JSON Schema, there is no longer interaction between required and readOnly/writeOnly in relation to requests and responses.
format (whether byte, binary, or base64) is no longer used to describe file payloads. As part of JSON Schema compliance, now contentEncoding and contentMediaType can be used for such specification.

Of these I think the only one that poses an issue for apispec is the nullable keyword being removed. Going forward it looks like JSON Schema indicates that a parameter is nullable via the type field like: 
```json
{
    "type": ["string", "null"]
}
```
So we will need to rework the way `field2nullable` works. Possibly combining it with `field2type_and_format`.

apispec doesn't use exclusiveMaximum or exclusiveMinimum so no issue there.

I don't think apispec does anything to tie required to readOnly/writeOnly.

Dealing with files is more of a web framework thing so that's probably left as a client responsibility.

niraveen:
May I know when version OpenAPI 3.1.0 will be supported? I see webhook schema is added in 3.1.0

lafrech:
No one is working on it at the moment to my knowledge.

Contributions welcome. If there's any change you need, feel free to discuss it here or in a dedicated issue.

gwerbin:
> No one is working on it at the moment to my knowledge.
> 
> Contributions welcome. If there's any change you need, feel free to discuss it here or in a dedicated issue.

What do you think the difficulty level is for this task, for a new contributor? It seems like the list of changes is small, but there are always "gotchas" and an adjustment period for people who aren't already familiar with the code.

Bangertm:
[...]

PR #999, opened 2025-11-18
Feat/support `examples` property

Flix6x:
I realize you did some work yourself on supporting the `examples` property, but I'm not sure if that is still ongoing. The [`examples` branch](https://github.com/marshmallow-code/apispec/tree/examples) hasn't been updated in 3 years.

Related discussion on supporting OpenAPI 3.1: https://github.com/marshmallow-code/apispec/issues/579

The [OpenAPI upgrading docs](https://learn.openapis.org/upgrading/v3.0-to-v3.1.html#change-schema-example-to-examples) suggest moving from `example` to `examples`, but as of now, both seem to be supported, so I haven't removed the singular `example` from the list of valid properties.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -88,3 +88,4 @@ Contributors (chronological)
 - Robin `@allrob23 <https://github.com/allrob23>`_
 - Xingang Zhang `@0x0400 <https://github.com/0x0400>`_
 - Lewis Haley `@LewisHaley <https://github.com/LewisHaley>`_
+- Felix Claessen `@Flix6x <https://github.com/Flix6x>`_
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,7 @@ Other changes:
 - Officially support Python 3.14 (:pr:`998`).
 - Drop support for Python 3.8 (:pr:`994`).
 - Support ``examples`` property from field metadata (:pr:`999`).
+  Thanks :user:`Flix6x` for the PR.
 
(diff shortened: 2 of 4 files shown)


lafrech:
Hi. Thanks for contributing.
In flask-smorest, I use examples extensively (see for instance https://github.com/marshmallow-code/flask-smorest/pull/68) without the need for this change here. But there I do it at schema level while this change acts at field level, which I guess is fine, too, so I guess you're (I was) right about it.
Perhaps the unit tests would clarify the use case. Do you think you could add one?

Flix6x:
Thanks, indeed my use case is **field** examples rather than **schema** examples. Something like [this](https://github.com/marshmallow-code/apispec/pull/999/commits/eab75c737746e46c9002f1d59b1fcd3c46f5855c)? I noticed two things:

1. The test would fail on `dev`, as expected.
2. The test succeeds on `feat/support-examples-property`, which I feel is somewhat unexpected, given that the fixture checks version 2.0 and 3.0.0. At first I thought the field examples were introduced in v3.1, but I think 3.0 already had them. 2.0 not, though.

lafrech:
Thanks!

> The test succeeds on `feat/support-examples-property`, which I feel is somewhat unexpected, given that the fixture checks version 2.0 and 3.0.0. At first I thought the field examples were introduced in v3.1, but I think 3.0 already had them. 2.0 not, though.

apispec doesn't check that the user uses properties that are valid wrt OpenAPI version.

PR #1013, opened 2026-02-13
fix: recognize iso8601 and rfc822 as valid DateTime format aliases

### LINKED ISSUE #970: iso8601 format not detected correctly
Date and DateTime with format argument as iso8601 it is not rendered correctly.

Actually marshallow supports also the following:
```
        "iso": utils.isoformat,
        "iso8601": utils.isoformat,
        "rfc": utils.rfcformat,
        "rfc822": utils.rfcformat,
```

```
works:
datetime = fields.DateTime(dump_only=True, format='iso')
does not work:
datetime2 = fields.DateTime(dump_only=True, format='iso8601')
```

https://github.com/marshmallow-code/marshmallow/blob/dev/src/marshmallow/fields.py#L1275-L1291

karthiksai109:
Fixes #970
 
Marshmallow's DateTime field accepts both short and long format names
(iso/iso8601, rfc/rfc822) in its SERIALIZATION_FUNCS and
DESERIALIZATION_FUNCS. However, datetime2properties only checked for
the short forms, causing iso8601 and rfc822 to fall through to the
custom format branch — producing incorrect OpenAPI output.
 
- iso8601 → was missing "format": "date-time", got "format": None
- rfc822 → was missing the RFC pattern/example, got "format": None
 
Added the long-form aliases to the existing checks and tests for both.
Full test suite: 598 passed, 6 skipped.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -89,3 +89,4 @@ Contributors (chronological)
 - Xingang Zhang `@0x0400 <https://github.com/0x0400>`_
 - Lewis Haley `@LewisHaley <https://github.com/LewisHaley>`_
 - Felix Claessen `@Flix6x <https://github.com/Flix6x>`_
+- Karthik Ramadugu `@karthiksai109 <https://github.com/karthiksai109>`_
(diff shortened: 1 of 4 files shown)


sloria:
Thank you!

PR #1014, opened 2026-02-17
Add OpenAPI Links support (#615)

### LINKED ISSUE #615: Support for OpenAPI Links?
Is there support for [OpenAPI Links](https://swagger.io/docs/specification/links/)? I couldn't find it anywhere.

Manually adding the following to the docstring works
```yaml
links:
    Get User By Id:
        operationId: UserGet
        parameters:
            id: $response.body#/id
```
but I don't think the validation part works, as something like this
```yaml
links:
    Get User By Id:
        operationId: UserGet
        parameters:
            id: RandomString
```
also validates (using `apispec.utils.validate_spec`).

kingdomOfIT:
Closes #615

Adds support for defining OpenAPI Links on path operations via the `links` 
kwarg on `APISpec.path()`, as discussed in the issue with @zedrdave.

Links accept a dictionary of tuples in the format agreed with @lafrech:

links={
    "GetUser": ("/users/{user_id}", "GET"),
    "UpdateUser": ("/users/{user_id}", "PUT"),
    "GetUserCustomParam": ("/users/{user_id}", "GET", {"user_id": "$response.body#/id"})
}

Implementation automatically uses `operationId` when the target operation 
has one registered, falling back to `operationRef` with JSON Pointer escaping 
when not. Links are only processed for OpenAPI 3+.

### DIFF
--- a/src/apispec/core.py
+++ b/src/apispec/core.py
@@ -51,12 +51,14 @@ class Components:
         self.parameters: dict[str, dict] = {}
         self.headers: dict[str, dict] = {}
         self.examples: dict[str, dict] = {}
+        self.links: dict[str, dict] = {}
         self.security_schemes: dict[str, dict] = {}
         self.schemas_lazy: dict[str, dict] = {}
         self.responses_lazy: dict[str, dict] = {}
         self.parameters_lazy: dict[str, dict] = {}
         self.headers_lazy: dict[str, dict] = {}
         self.examples_lazy: dict[str, dict] = {}
+        self.links_lazy: dict[str, dict] = {}
 
         self._subsections = {
             "schema": self.schemas,
@@ -64,6 +66,7 @@ class Components:
             "parameter": self.parameters,
             "header": self.headers,
             "example": self.examples,
+            "link": self.links,
             "security_scheme": self.security_schemes,
         }
         self._subsections_lazy = {
@@ -72,6 +75,7 @@ class Components:
             "parameter": self.parameters_lazy,
             "header": self.headers_lazy,
             "example": self.examples_lazy,
+            "link": self.links_lazy,
         }
 
     def to_dict(self) -> dict[str, dict]:
@@ -294,6 +298,24 @@ class Components:
         self._register_component("example", component_id, component, lazy=lazy)
         return self
 
+    def link(
+        self, component_id: str, component: dict, *, lazy: bool = False
+    ) -> Components:
+        """Add a link which can be referenced
+
+        :param str component_id: identifier by which link may be referenced
+        :param dict component: link fields
+        :param bool lazy: register component only when referenced in the spec
+
+        https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#linkObject
+        """
+        if component_id in self.links:
+            raise DuplicateComponentNameError(
+                f'Another link with name "{component_id}" is already registered.'
+            )
+        self._register_component("link", component_id, component, lazy=lazy)
+        return self
+
     def security_scheme(self, component_id: str, component: dict) -> Components:
         """Add a security scheme which can be referenced.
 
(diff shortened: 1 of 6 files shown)


src/apispec/core.py
```diff
@@ -629,3 +656,69 @@ def _clean_operations(
...
+            route = link_tuple[0]
+            method = link_tuple[1]
+            parameters = link_tuple[2] if len(link_tuple) == 3 else {}
+
+            if not isinstance(route, str):
+                raise APISpecError(f"Link '{link_name}' route must be a string")
+            if not isinstance(method, str):
+                raise APISpecError(f"Link '{link_name}' method must be a string")
+            if not isinstance(parameters, dict):
+                raise APISpecError(f"Link '{link_name}' parameters must be a dict")
```
lafrech: I wouldn't bother checking those types. We generally don't do that in the codebase.
Besides, a mapping that is not a `dict` would be fine.

src/apispec/core.py
```diff
@@ -629,3 +656,69 @@ def _clean_operations(
...
+
+            # Validate method
+            method = method.lower()
+            valid_methods = set(VALID_METHODS[self.openapi_version.major])
+            if method not in valid_methods:
+                raise APISpecError(
+                    f"Link '{link_name}' has invalid HTTP method: {method}"
+                )
+
+            # Prefer operationId if target operation already registered
```
lafrech: Is it about registration?

Should this read "Prefer operationId if defined"?

tests/test_links.py
```diff
@@ -0,0 +1,385 @@
+import pytest
```
lafrech, line 1: Other components tests are in test_core, in `TestComponents` with component type prefixing the name (e.g. test_links_...). I'd rather stick to that.

If this file grows too large (in fact, it already does), we may split it to put components aside, for instance.

tests/test_links.py
```diff
@@ -0,0 +1,385 @@
...
+        links = get_links(spec)
+        assert "GetUser" in links
+        assert links["GetUser"] == link
+
+    def test_link_component_lazy_registration(self, spec):
+        link = {"operationRef": "#/paths/~1users~1{user_id}/get"}
+        spec.components.link("GetUser", link, lazy=True)
+        # Should not appear until referenced
+        links = get_links(spec)
+        assert "GetUser" not in links
```
lafrech: Other "lazy" tests check that the item does appear once referenced.

tests/test_links.py
```diff
@@ -0,0 +1,385 @@
...
+
+@pytest.mark.parametrize("spec", ("3.0.0",), indirect=True)
+class TestLinksErrorHandling:
+    def test_invalid_tuple_length(self, spec):
+        with pytest.raises(APISpecError, match="must be a tuple of"):
+            spec.path(
+                path="/posts",
+                operations={"post": {"responses": {"201": {"description": "Created"}}}},
+                links={
+                    "GetPost": ("/posts/{id}",),  # Only one element
```
lafrech: I'd test both 1 and 4. Can be achieved with parametrization.

tests/test_links.py
```diff
@@ -0,0 +1,385 @@
...
+        with pytest.raises(APISpecError, match="invalid HTTP method"):
+            spec.path(
+                path="/posts",
+                operations={"post": {"responses": {"201": {"description": "Created"}}}},
+                links={
+                    "GetPost": ("/posts/{id}", "INVALID"),
+                },
+            )
+
+    def test_invalid_parameters_type(self, spec):
```
lafrech: As written above, I wouldn't bother with dict and str types.

lafrech (commented):
Thanks for contributing!

Just a few minor comments.

lafrech (approved):
Thanks. This looks good.

There is indeed a lot of new tests, so OK for the new file, no big deal.

lafrech:
Leaving this open for a few days to let other users/maintainers give it a look.

Feel free to ping me if this gets out of my head and doesn't ship next week.

kingdomOfIT:
> Leaving this open for a few days to let other users/maintainers give it a look.
> 
> Feel free to ping me if this gets out of my head and doesn't ship next week.

Just a quick reminder here, @lafrech 😃

lafrech:
Thanks. (I did forget.)

This is released in 6.10.0.

Issue #308, opened 2018-10-08
RFC: Remove validate_spec feature


lafrech:
See discussions in https://github.com/marshmallow-code/apispec/pull/299.

No strong feeling either way. Indeed, it probably works for most users, and anyone unhappy about the default values may install and call prance independently, so this implementation at least serves the purpose of showing how to do it.

sloria:
We can leave this issue open for discussion, but I'm going to remove this from the 1.0 milestone in the interest of cutting scope.

lafrech:
Recent CI issues are due to changes in openapi_spec_validator 0.5.x, which is a dependency of prance. Until prance releases a fix, we should pin openapi_spec_validator<0.5 for the "validation" extra requires.

I wouldn't mind removing the feature from the public API and only keeping it internally for the tests. I admit this wouldn't solve the whole issue, just make it less of a problem to modify dependency versions without it being a breaking change. Also, if there hasn't been any issue in 4 years, maybe I'm overreacting.

My point is that if it is only used for the test, we are free to change versions, even the lib used internally, even remove the whole thing without notice, so it makes maintenance a bit easier.

Nobody has :+1: or :-1: since I opened this RFC so perhaps nobody cares.

SpeeddY:
Just run into this issue as I went to update `openapi-spec-validator` to 0.5.1, was about to raise an issue here but stumbled upon this thread. `ImportError: cannot import name 'validate_v2_spec' from 'openapi_spec_validator'`. The prance issue: https://github.com/RonnyPfannschmidt/prance/issues/131

lafrech:
Yes, I saw that issue. I should have linked to it. Thanks.

Does this mean you're using the feature and you're :-1: about removing it?

In practice the feature doesn't provide much more than an abstraction for the validation frameworks. Anyone can install any validator on their own.

SpeeddY:
We are using the feature but it would be an easy refactor to remove it (the refactor would use the new "detecting version validate_spec" method from openapi-spec-validator)

If you go ahead with this RFC I presume you can remove the `openapi-spec-validator` dependency entirely? 
(I am not a downvote)
If you decided to keep it, it would be great if it could work with the latest openapi-spec-validator

PR #1021, opened 2026-03-03
docs: remove reference to removed validate_spec utility

### LINKED ISSUE #969: Remove mentions of validate_spec from docs
Hi! New user here.

I just spent an hour figuring out where `validate_spec` has got to, before I came across #308 

It's mentioned in a code snippet in https://github.com/marshmallow-code/apispec/blob/dev/docs/special_topics.rst and maybe other places as well. It seems that should be removed.

I'm happy to PR, but do we want to just drop all mention from the docs? I would need some guidance.

I do want to validate my spec, and in #308 there was mention of `openapi-spec-validator`. What would be appropriate to document?

toroleapinc:
## Summary

Remove the import and call to `validate_spec` from the special_topics documentation example. This function was removed from the public API (see #308) but the docs still referenced it, causing confusion for new users.

Closes #969

For spec validation, users can use the [openapi-spec-validator](https://pypi.org/project/openapi-spec-validator/) package directly.

### DIFF
--- a/docs/special_topics.rst
+++ b/docs/special_topics.rst
@@ -93,7 +93,6 @@ Here is an example that includes a `Server Object <https://github.com/OAI/OpenAP
     import yaml
     from apispec import APISpec
     from apispec.ext.marshmallow import MarshmallowPlugin
-    from apispec.utils import validate_spec
 
     OPENAPI_SPEC = """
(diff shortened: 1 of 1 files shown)


sloria:
thanks!

PR #1067, opened 2026-08-01
fix: use OpenAPI 'minimum' for timestamp DateTime formats

### LINKED ISSUE #1064: DateTime(format="timestamp") generates invalid OpenAPI schema using min instead of minimum
**Description**

`apispec` generates an invalid OpenAPI schema for `marshmallow.fields.DateTime(format="timestamp")`.

The generated schema contains the `min` property:

```yaml
changedOn:
  type: number
  format: float
  example: "1676451245.596"
  min: "0"
````

However, `min` is not a valid OpenAPI Schema Object property. As a result, the generated specification fails validation with `openapi-spec-validator`:

```
OpenAPIValidationError:
Property 'min' is not allowed
```

This was reproduced with:

* apispec 6.10.0
* marshmallow 3.26.2
* openapi-spec-validator 0.9.0
* OpenAPI 3.0.2

**Expected behavior**

The generated schema should use the standard OpenAPI keyword `minimum` instead of `min`, for example:

```yaml
changedOn:
  type: number
  format: float
  example: "1676451245.596"
  minimum: 0
```

or omit the constraint entirely if no minimum value is intended.

**Minimal reproduction**

````python
from apispec import APISpec
from pprint import pprint

class TestSchema(Schema):
    ts = fields.DateTime(format="timestamp")

spec = APISpec(
    title="test",
    version="1.0",
    openapi_version="3.0.2",
)

spec.components.schema("Test", schema=TestSchema)

pprint(spec.to_dict()["components"]["schemas"]["Test"])
`````

Output:

```python
{'additionalProperties': False,
 'properties': {'ts': {'example': '1676451245.596',
                       'format': 'float',
                       'min': '0',
                       'type': 'number'}},
 'type': 'object'}
```

The file `field_converter.py` contains the code that causes this issue:
```python
            elif field.format == "timestamp":
                ret = {
                    "type": "number",
                    "format": "float",
                    "example": "1676451245.596",
                    "min": "0",
                }
            elif field.format == "timestamp_ms":
                ret = {
                    "type": "number",
                    "format": "float",
                    "example": "1676451277514.654",
                    "min": "0",
                }

```

**Specification reference**

The OpenAPI Schema Object defines the numeric constraint keyword as `minimum`; `min` is not a valid keyword.

OpenAPI 3.0.2 Schema Object:
[https://spec.openapis.org/oas/v3.0.2#schema-object](https://spec.openapis.org/oas/v3.0.2#schema-object)

Sanjays2402:
Closes #1064

`DateTime(format="timestamp")` / `format="timestamp_ms"` produced `"min": "0"`, which is not an OpenAPI Schema Object keyword, so the generated spec failed `openapi-spec-validator` with `Property 'min' is not allowed`. Emits the standard numeric `minimum: 0` instead; the two existing timestamp tests are updated to match.

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

### DIFF
--- a/src/apispec/ext/marshmallow/field_converter.py
+++ b/src/apispec/ext/marshmallow/field_converter.py
@@ -592,14 +592,14 @@ class FieldConverterMixin:
                     "type": "number",
                     "format": "float",
                     "example": "1676451245.596",
-                    "min": "0",
+                    "minimum": 0,
                 }
             elif field.format == "timestamp_ms":
                 ret = {
                     "type": "number",
                     "format": "float",
                     "example": "1676451277514.654",
-                    "min": "0",
+                    "minimum": 0,
                 }
             else:
                 ret = {
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,12 @@ Changelog
 unreleased
 **********
 
+Bug fixes:
+
+- ``MarshmallowPlugin``: Emit ``minimum: 0`` rather than the non-standard
+  ``min: "0"`` for ``DateTime`` fields with the ``"timestamp"`` and
+  ``"timestamp_ms"`` formats (:issue:`1064`).
+
 Other changes:
 
 - Drop support for marshmallow 3, which is EOL.
--- a/tests/test_ext_marshmallow_field.py
+++ b/tests/test_ext_marshmallow_field.py
@@ -591,7 +591,7 @@ def test_datetime2property_timestamp(spec_fixture):
     assert res == {
         "type": "number",
         "format": "float",
-        "min": "0",
+        "minimum": 0,
         "example": "1676451245.596",
     }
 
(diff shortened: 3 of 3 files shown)


lafrech (approved):
LGTM. Thanks!

src/apispec/ext/marshmallow/field_converter.py
```diff
@@ -592,14 +592,14 @@ def datetime2properties(self, field, **kwargs: typing.Any) -> dict:
                     "type": "number",
                     "format": "float",
                     "example": "1676451245.596",
-                    "min": "0",
+                    "minimum": 0,
```
lafrech, line 595: Should be 

    "minimum": "0",

lafrech (changes requested):

src/apispec/ext/marshmallow/field_converter.py
ULisichkin: In this case, the value "0" is invalid and causes the validator to fail, as I already mentioned in the opened Issue.
See example https://spec.openapis.org/oas/v3.0.2#simple-model

PR #1069, opened 2026-08-14
fix: intersect OneOf and Equal choices in field2choices

### LINKED ISSUE #198: Minor issues with field2choices

alexchen-sys:
Fixes #198

When multiple choice validators are defined on a field (such as multiple `validate.OneOf`, multiple `validate.Equal`, or a mix of `OneOf` and `Equal`), the allowed choices should be the intersection of the values allowed by each validator.

Previously:
- `validate.Equal` overrode any preceding `validate.OneOf` choices.
- Multiple `validate.Equal` validators concatenated their values rather than intersecting them.

### Changes
- Treat `validate.Equal` comparable values as single-element choice sets and intersect all choice sets via `OrderedSet`.
- Added unit tests for multiple `OneOf`, `OneOf` + `Equal`, and multiple `Equal` validators (both intersecting and non-intersecting).

### DIFF
--- a/src/apispec/ext/marshmallow/field_converter.py
+++ b/src/apispec/ext/marshmallow/field_converter.py
@@ -242,21 +242,14 @@ class FieldConverterMixin:
         """
         attributes = {}
 
-        comparable = [
-            validator.comparable
-            for validator in field.validators
-            if hasattr(validator, "comparable")
-        ]
-        if comparable:
-            attributes["enum"] = comparable
-        else:
-            choices = [
-                OrderedSet(validator.choices)
-                for validator in field.validators
-                if hasattr(validator, "choices")
-            ]
-            if choices:
-                attributes["enum"] = list(functools.reduce(operator.and_, choices))
+        choices = []
+        for validator in field.validators:
+            if hasattr(validator, "choices"):
+                choices.append(OrderedSet(validator.choices))
+            elif hasattr(validator, "comparable"):
+                choices.append(OrderedSet([validator.comparable]))
+        if choices:
+            attributes["enum"] = list(functools.reduce(operator.and_, choices))
 
         if field.allow_none:
             enum = attributes.get("enum")
--- a/tests/test_ext_marshmallow_field.py
+++ b/tests/test_ext_marshmallow_field.py
@@ -198,7 +198,62 @@ def test_field_with_choices_multiple(spec_fixture):
         ]
     )
     res = spec_fixture.openapi.field2property(field)
-    assert set(res["enum"]) == {"brian", "john"}
+    assert res["enum"] == ["brian", "john"]
+
+
+def test_field_with_choices_multiple_non_intersecting(spec_fixture):
+    field = fields.Str(
+        validate=[
+            validate.OneOf(["freddie", "brian"]),
+            validate.OneOf(["john", "roger"]),
+        ]
+    )
+    res = spec_fixture.openapi.field2property(field)
+    assert res["enum"] == []
+
+
+def test_field_with_choices_and_equal(spec_fixture):
+    field = fields.Str(
+        validate=[
+            validate.OneOf(["freddie", "brian", "john"]),
+            validate.Equal("brian"),
+        ]
+    )
+    res = spec_fixture.openapi.field2property(field)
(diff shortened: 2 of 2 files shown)


tests/test_ext_marshmallow_field.py
```diff
@@ -198,7 +198,62 @@ def test_field_with_choices_multiple(spec_fixture):
         ]
     )
     res = spec_fixture.openapi.field2property(field)
-    assert set(res["enum"]) == {"brian", "john"}
+    assert res["enum"] == ["brian", "john"]
```
lafrech: I guess the use of set here was intentional as we don't want to ensure the order.

tests/test_ext_marshmallow_field.py
```diff
@@ -198,7 +198,62 @@ def test_field_with_choices_multiple(spec_fixture):
...
+
+
+def test_field_with_choices_multiple_non_intersecting(spec_fixture):
```
lafrech: I'd write oneof instead of choices in the test name.

tests/test_ext_marshmallow_field.py
```diff
@@ -198,7 +198,62 @@ def test_field_with_choices_multiple(spec_fixture):
...
+        validate=[
+            validate.OneOf(["freddie", "brian"]),
+            validate.Equal("john"),
+        ]
+    )
+    res = spec_fixture.openapi.field2property(field)
+    assert res["enum"] == []
+
+
+def test_field_with_multiple_equal_matching(spec_fixture):
```
lafrech: I think the whole thing could be factorized in three tests:

- multiple oneof
- multiple equal
- oneof and equal

using parametrization to pass inputs and expected output.

src/apispec/ext/marshmallow/field_converter.py
```diff
@@ -242,21 +242,14 @@ def field2choices(
...
-            attributes["enum"] = comparable
-        else:
-            choices = [
-                OrderedSet(validator.choices)
-                for validator in field.validators
-                if hasattr(validator, "choices")
-            ]
-            if choices:
-                attributes["enum"] = list(functools.reduce(operator.and_, choices))
+        choices = []
```
lafrech, line 245: I've been wondering why we don't use a set here and update it in the loop.

This is probably due to a slight performance advantage doing it this way.

lafrech (approved):
100+ PRs a month is a bit of a reg flag, but this PR makes sense, so...

lafrech (changes requested):
See review comments.

tests/test_ext_marshmallow_field.py
alexchen-sys: Done — reverted to set() comparison.

tests/test_ext_marshmallow_field.py
alexchen-sys: Renamed.

tests/test_ext_marshmallow_field.py
alexchen-sys: Factorized into three parametrized tests: multiple_oneof, multiple_equal, oneof_and_equal.

alexchen-sys:
Thanks for the review. Comments addressed in e485237 — set comparison restored, tests renamed to oneof, factorized into three parametrized tests.

